PR #2, opened 2017-03-21
Import code from georg's repo

maurodoglio:


probe_scraper/parsers/third_party/parse_events.py
```diff
@@ -0,0 +1,304 @@
...
+    return string.capwords(s, sep).replace(sep, "")
+
+class OneOf:
+    """This is a placeholder type for the TypeChecker below.
+    It signals that the checked value should match one of the following arguments
+    passed to the TypeChecker constructor.
+    """
+    pass
+
+class TypeChecker:
```
maurodoglio: The implementation of this class seems to be a bit odd. An alternative could be to have a set of TypeChecker classes which know how to run their own validation. Something like
```python
class AtomicTypeChecker:
    """Validate a simple value against a given type"""
    def __init__(self, instance_type):
        self.instance_type = instance_type
   
    def check(self, identifier, key, value):
        pass


class ListTypeChecker:
    """Validate a list of values against a given type"""
    def __init__(self, instance_type):
        self.instance_type = instance_type
   
    def check(self, identifier, key, value):
        pass


class DictTypeChecker:
    """Validate the keys and the values of a dict against a given type"""
    def __init__(self, keys_instance_type, values_instance_type):
        self.keys_instance_type  = keys_instance_type
        self.values_instance_type = values_instance_type
   
    def check(self, identifier, key, value):
        pass


class MultiTypeChecker:
    """Validate a simple value against a list of possible types"""
    def __init__(self, *instance_types):
        if not instance_type:
            raise ValueError("At least one instance type is required")
        self.instance_types = instance_types
        self.values_instance_type = values_instance_type
   
    def check(self, identifier, key, value):
        pass
```
This way you could still call .check on the instances of these classes, ignoring what's the actual TypeChecker class that you are using. Also, this design lets you define even more specific checkers (or validators) e.g. 
```python
class StringTypeChecker(AtomicTypeChecker):
    def __init__(min_length, max_length):
        super(StringTypeChecker, self).__init__(basestring)
        self.min_length = min_length
        self.max_length = max_length

    def check(self, identifier, key, value):
        assert isinstance(value, basestring) and self.min_length <= len(value) <= self.max_length
```

georgf: This is good feedback. However, the files in `third_party/` are imports from mozilla-central, so i'd like to keep changes to them here to a minimum - i should have called that out.
I filed [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1349452) on this.

probe_scraper/parsers/events.py
```diff
@@ -0,0 +1,84 @@
...
+
+    defaults = {
+        "expiry_version": "never",
+        "expiry_day": "never",
+        "name": e.methods[0],
+        "description": "<TODO>",
+        "cpp_guard": None,
+    }
+
+    data = {
```
georgf: Is this helpful in style?
I find it more readable to keep data-oriented dicts in this same style, to quickly see the contents.

georgf:
I fixed the review comments except where i replied otherwise.

probe_scraper/scraper.py
```diff
@@ -0,0 +1,192 @@
...
+ERROR_CACHE_FILENAME = 'probe_scraper_errors_cache.json'
+
+
+def load_tags(channel):
+    uri = CHANNELS[channel]['base_uri'] + "json-tags"
+    r = requests.get(uri)
+    if r.status_code != requests.codes.ok:
+        raise Exception("Request returned status " + str(r.status_code) + " for " + uri)
+
+    ctype = r.headers['content-type']
```
maurodoglio: ctype is the name of a module in the python standard library. Please change it to `content_type` which is also more meaningful.

georgf:
Thanks for the feedback Mauro!
I think i addressed everything, so merging this in.

PR #4, opened 2017-03-28
Enable CI using TravisCI

Dexterp37:
This commit will enable linting through flake8 and pytest testing on every PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: python
+python:
+  - "2.7"
+# command to install dependencies
+install:
+  - pip install -r requirements.txt -r test_requirements.txt
+  - pip install .
+before_script:
+  - flake8 .
+# command to run tests
+script: pytest tests
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
 from setuptools import setup
-from distutils.core import setup
 
 setup(name='probe-scraper',
       version='0.1',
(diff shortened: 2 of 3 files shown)


Dexterp37:
@maurodoglio , any idea why it cannot find `pytest`? The same steps work correctly locally

maurodoglio:
@Dexterp37 It looks like travis is using a cached version of pytest. It could be that an older version is installed on that machine by default.
I would try to specify a minimum version pytest>=3 in the requirements file, see https://github.com/pytest-dev/pytest/issues/1629

.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+  - "2.7"
+# command to install dependencies
+install:
+  - pip install -r requirements.txt
+  - pip install .
+  - pip install -r test_requirements.txt
+before_script:
+  - flake8 .
+# command to run tests
+script: pytest tests
```
maurodoglio: Missing blank line at the end of file

.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: python
+python:
+  - "2.7"
+# command to install dependencies
+install:
+  - pip install -r requirements.txt
```
maurodoglio: you can combine the first and the second call to pip:
```bash
pip install -r requirements.txt -r test_requirements.txt
```

Dexterp37:
Thanks @maurodoglio who helped fixing the test errors!

PR #5, opened 2017-03-30
Add support for Nightly and "tip" revisions

georgf:
This adds:
- pulling from the Nightly channel, using the "aurora base" tags it has
- pulling tip revisions on each channel to always show the latest changes


georgf:
Yes, this is tested locally with the web frontend.

PR #10, opened 2017-04-03
Update notebook to properly run

georgf:
Update notebook to:
- not error out on current EMR git state
- just remove existing code, cache & output dirs to avoid stale states

### DIFF
--- a/notebooks/load_and_run.ipynb
+++ b/notebooks/load_and_run.ipynb
@@ -47,6 +47,51 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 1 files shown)


notebooks/load_and_run.ipynb
```diff
@@ -47,6 +47,51 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "!git config --global user.email \"gfritzsche@mozilla.com\" && \\\n",
```
Dexterp37: Is this command intentional or a leftover? If the former, why is this needed?

georgf: This is intentional.
git was erroring out on our current machines without it.

PR #21, opened 2017-09-06
Modify lastUpdate time information

DegCode:
Add time of the day and UTC infomation to lastUpdate using dateutil module.
Change it to be a full ISO format date string.


georgf:
[This issue](https://github.com/georgf/fx-data-explorer/issues/9) could be a good follow-up if you are interested.
It is the front-end change for using the date info you added here.

PR #22, opened 2017-09-06
Make histogram parser to skip unknown properties

georgf:
The probe-scraper job is currently broken as we introduced a new
property for histograms that the tools here are not aware of.
Until we merge/unify the parser tools, let’s skip unknown
keys/properties here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,5 @@ temp/
 _tmp/
 _out/
 _temp/
+
+*.json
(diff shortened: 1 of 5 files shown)


Dexterp37:
The version in m-c already has this change. Can't we use that one and update this copy? Or is that what you are doing?

georgf:
We would need to two-way merge for that.
We forked the parsers, did some changes here and didn't merge back yet.
So... taking that copy would potentially be more involved AFAICT.

Issue #23, opened 2018-01-08
Break down the output by release channel


Dexterp37:
In order to make the scraper future proof and to support other internal products we need to break down the output by channel (Nightly, Aurora, Beta, Release, ESR).

We will still need to support the monolithic output mode with all the probes from all the channels to make it easy for the data-explorer to consume it.

Dexterp37:
The probe scraper currently produces 3 JSON files:

- `revisions.json`
- `general.json`
- `probes.json`

The `general.json` only contains the `lastUpdate` property, which is the date the files were generated.

The `revisions.json` file contains a list of revision hash -> Firefox version mapping for each channel.

Finally, the `probes.json` file contains the data for all the probes. It has the following format:

```
{
  probe_id: {
    type: 'histogram',
    name
    history: {
      "nightly": [
        {
          optout: True,
          ... other probe details ...
          revisions: {first: ..., last: ...}
        },
        ...
      ],
      ... other channels ...
    }
  }
}
```

My proposal is to keep the current format, writing a separate `probes.json` file for each channel, and only list the relevant channel data in the `history` section of the probe.

@sunahsuh and @fbertsch, since you are the main consumers of the probe data I can think of right now, what do you think of this proposal?

fbertsch:
Quick recap fro discussion in IRC:
- This will include historic probes. Yay!
- We can remove revisions.json and just sub in the FF version
- Long-term we may need to rethink how we deal with historic revisions

Dexterp37:
This happened, the output is already being written to S3 and the documentation changes landed. Closing this off!

PR #25, opened 2018-01-10
Enable writing probe data broken down by channel

Dexterp37:
This adds a new output mode, alongside the existing one, which allows to write one file for each release channel. Each individual file contains only the probes for that channel.

This adds the output generator for #23 .

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -5,6 +5,7 @@
 import os
 import json
 import datetime
+import errno
 from dateutil.tz import tzlocal
 import tempfile
 import argparse
@@ -38,6 +39,37 @@ def general_data():
     }
 
 
+def write_probe_data(probe_data, revisions, out_dir):
+    # Serialize extracted data.
+    def dump_json(data, out_dir, file_name):
+        # Make sure that the output directory exists. This also creates
+        # intermediate directories if needed.
+        try:
+            os.makedirs(out_dir)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise
+
+        path = os.path.join(out_dir, file_name)
+        with open(path, 'w') as f:
+            print "  " + path
+            json.dump(data, f, sort_keys=True, indent=2)
+
+    # Save all our files to "outdir/firefox/..." to mimic a REST API.
+    base_dir = os.path.join(out_dir, "firefox")
+
+    print "\nwriting output:"
+    dump_json(general_data(), base_dir, 'general.json')
+    dump_json(revisions, base_dir, 'revisions.json')
+
+    # Break down the output by channel. We don't need to write a revisions
+    # file in this case, the probe data will contain human readable version
+    # numbers along with revision numbers.
+    for channel, channel_probes in probe_data.iteritems():
+        data_dir = os.path.join(base_dir, channel, "main")
+        dump_json(channel_probes, data_dir, 'all_probes')
+
+
 def main(temp_dir, out_dir):
     # Scrape probe data from repositories.
     node_data = scraper.scrape(temp_dir)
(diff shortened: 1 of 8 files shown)


Dexterp37:
This is intentionally missing the file upload part: is it enough to upload all the files to S3, in a separate directory?

georgf:
Totally, let's just upload the files into a separate directory for now.
Then we can work on the API/directory structure in a separate issue.

tests/test_transform_probes.py
```diff
@@ -133,6 +133,54 @@
...
+                    'cpp_guard': None,
+                    'description': 'A description.',
+                    'details': {
+                        'high': 10,
+                        'keyed': False,
+                        'kind': 'exponential',
+                        'low': 1,
+                        'n_buckets': 5
+                    },
+                    'expiry_version': '53.0',
```
georgf: While we're here and changing version handling... Let's just strip the output to '53' (i.e. major version only).
We're making that change client-side too and never used minor versions.

Dexterp37:
This PR changes the format of the `probes.json` file when generating it for the break-down-by-channel mode.

The monolithic mode file has the following format:

```
{
  probe_id: {
    type: 'histogram',
    name: 'some name',
    history: {
      "nightly": [
        {
          optout: True,
          ... other probe details ...
          revisions: {first: ..., last: ...}
        },
        ...
      ],
      ... other channels ...
    }
  }
}
```

The broken down by channel probes file has the following format:

```
{
  probe_id: {
    type: 'histogram',
    name: 'some name',
    history: [
        {
          optout: True,
          ... other probe details ...
          revisions: {
              first_revision: ...,
              first_version: ...,
              last_revision: ...,
              last_version: ...,
          }
        },
        ...
      ],
  }
}
```

Please note the changes:

- the `history` is a vector for the broken down by channel version; it only contains the historical entries for a single channel;
- the `revisions` section contains the human readable version along with the revision for the broken down by channel version.

tests/test_transform_probes.py
```diff
@@ -133,6 +133,54 @@
...
+                    'optout': True,
+                    'revisions': {
+                        'first_revision': 'node_id_2',
+                        'first_version': '51',
```
georgf: Can we add them as `versions: { first, last }` and leave `revisions` as is?
Then we can also add these to the "all channels output" and have the generally same format (except for the history layout).
probe-dictionary can soft-migrate then, we wouldn't need any special handling.

fbertsch: @Dexterp37 no, I think that's a good idea, it will make it easier to parse and use; but it does then pose the question: why have the monolithic at all, since it is just equal to the sum of the channel-breakdown partrs?

tests/test_transform_probes.py
Dexterp37: Done!

tests/test_transform_probes.py
Dexterp37: The monolithic will only be used during the transition period, so that we're not forced to change the probe dictionary

Dexterp37:
Here's some sample [output](https://drive.google.com/file/d/1Z0JxmMGRFZJOp7VfoHbRNur2E3kmQ992/view?usp=sharing).

tests/test_transform_probes.py
georgf: We can revisit this in the future, but for e.g. the probe-dictionary i usually need to pull all the channels.
Having this all in one file does:
- limit the number of http requests (perf impact)
- probably improves compression ratio (perf impact)
- generally reduces duplicate data (some probe data is shared across channels)

probe_scraper/runner.py
```diff
@@ -89,6 +99,8 @@ def dump_json(data, file_name):
                         help='Directory to store output files in.',
                         action='store',
                         default='.')
+    parser.add_argument('--break-by-channel', action='store_true',
```
georgf: One larger question here:
Why are we running the job two times?
Can we just output to per-channel and a monolithic file from one job run?

Dexterp37:
Ok, both commits are now ready for another look @georgf ! This is the sample [output](https://drive.google.com/file/d/1CfDo4mQELR9OTt_EbnZOPQfSTFPf2XGv/view?usp=sharing).

probe_scraper/runner.py
```diff
@@ -62,21 +94,13 @@ def main(temp_dir, out_dir):
                 results = PARSERS[probe_type].parse(paths, details["version"])
                 probes[channel][node_id][probe_type] = results
 
-    # Transform extracted data.
+    # Transform extracted data: get both the monolithic and by channel probe data.
     revisions = transform_revisions.transform(node_data)
-    probe_data = transform_probes.transform(probes, node_data)
+    probes_by_channel = transform_probes.transform(probes, node_data, True)
+    probes_by_channel["all"] = transform_probes.transform(probes, node_data, False)
```
georgf: Can we use something more descriptive than `True` & `False`?
Using a named argument or other ways.

PR #27, opened 2018-01-16
Add the file formats to the README.md file

Dexterp37:
This additionally removes the empty README file. This should land, ideally, after #25 gets merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,15 +20,6 @@ Run tests:
 pytest
 ```
 
-## Output location
-The output from the scheduled job can be found at:
-
-The output from the scheduled job can be found at:
-
-* https://analysis-output.telemetry.mozilla.org/probe-scraper/data/general.json
-* https://analysis-output.telemetry.mozilla.org/probe-scraper/data/revisions.json
-* https://analysis-output.telemetry.mozilla.org/probe-scraper/data/probes.json
-
 ## Module overview
 
 The module is built around the following data flow:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,3 +39,118 @@ The code layout consists mainly of:
...
+### Probe data file
+This file contains the data for the probes. The data might be spread across multiple files. It has the following format:
+
+```
+{
+  "<probe type>/<probe name>": {
+    "history": {
+      "<channel>": [
+        {
+          "cpp_guard": <string or null>,
```
sunahsuh: Could we get a description of this field?

README.md
```diff
@@ -48,3 +39,118 @@ The code layout consists mainly of:
...
+            "keyed": false,
+            "kind": "enumerated",
+            "low": 1,
+            "n_buckets": 12
+          },
+          "expiry_version": "never",
+          "optout": true,
+          "revisions": {
+            "first": "320642944e42a889db13c6c55b404e32319d4de6",
+            "last": "6f5fac320fcb6625603fa8a744ffa8523f8b3d71"
```
sunahsuh: Now that I see an example entry, I'm a bit concerned about ease of getting "probes active in the current latest release in each channel", which is a pretty common pattern

Dexterp37: The [probe dictionary](https://georgf.github.io/fx-data-explorer/index.html?search=&searchtype=in_any&optout=true&channel=release&constraint=is_in&version=57) does this by getting all the probes for all the channels and then filtering for `revision.first == <current revision>`. I think we can probably make it better though: do you want to file another issue on this repo?

README.md
Dexterp37: Good catch! This probe data comes directly from the Telemetry probe definition. I've added a note at the bottom of this section with a link to the Telemetry docs with the full documentation on all the fields that can be found there.

Dexterp37:
Let me go ahead and merge this, since this is just a doc update. Feel free to open another issue about changing the API if needed, @sunahsuh !

PR #28, opened 2018-01-16
Update the scalar parser and disable strict type checks

Dexterp37:
This fixes the parser complaining about the scalar definitions containing the "all_children" entry.

This was tested on ATMO. Sample output is available [here](https://drive.google.com/file/d/18c6TtKvl_5GdIR4QPEYbh8epi5UvJ6OF/view?usp=sharing).


Dexterp37:
Yes, it's just updating the scalar parser to the latest m-c. There were no errors going through the historical files (the output looks good to me as well).

Issue #32, opened 2018-01-20
Consider dropping Aurora channel


georgf:
We should check if anyone intends to use historical Aurora channel data.
If not, we could just drop support for that channel.

georgf:
We don't have concerns about dropping this.
The Aurora channel is gone for a while, so let's just proceed.

PR #37, opened 2018-02-15
Drop the "aurora" channel from the scraper

Dexterp37:
I tested this locally: aurora data is not being collected and the scraper still runs correctly. The latest aurora data will still be available on the REST-like API under `firefox/aurora` unless we decide to delete it.


Dexterp37:
This fixes #32

PR #45, opened 2018-05-09
Allow for unknown fields in Events.yaml

georgf:
This fixes the current failure that is caused by adding the new
“products” field.


georgf:
We already have divergence on the measurement parsers.
I'd leave upstream alone and rather figure out a general solution in the future (having a common code base).

PR #49, opened 2018-06-01
Fixes #44 - Scrape bug_numbers from histograms, scalars, and events

### LINKED ISSUE #44: Scrape bug numbers from probes that have them
Some probes have bug_numbers fields (in fact, all new ones are required to have them, and owners are encouraged to add them when they are updated). This seems like useful information we can surface in the probe scraper.

Marking this "help wanted" for mentorship. @Dexterp37 and @georgf have offered to mentor this bug.

After this is complete, we can file a bug in https://github.com/mozilla/telemetry-dashboard for augmenting Probe Dictionary to display this information and linkify the bug numbers to make it easy to flow from the dictionary to bugzilla to learn more.

chutten:
ref: #44, mozilla/probe-dictionary#43

### DIFF
--- a/probe_scraper/parsers/events.py
+++ b/probe_scraper/parsers/events.py
@@ -15,6 +15,7 @@ def extract_events_data(e):
         "expiry_version": "expiry_version",
         "expiry_day": "expiry_day",
         "cpp_guard": "cpp_guard",
+        "bug_numbers": "bug_numbers",
 
         "methods": "details/methods",
         "objects": "details/objects",
@@ -29,6 +30,7 @@ def extract_events_data(e):
         "name": e.methods[0],
         "description": e.description,
         "cpp_guard": None,
+        "bug_numbers": [],
     }
 
     data = {
--- a/probe_scraper/parsers/events.py
+++ b/probe_scraper/parsers/events.py
@@ -38,7 +38,7 @@ def extract_events_data(e):
     }
 
     for source_field, target_field in props.iteritems():
-        value = getattr(e, source_field, None)
+        value = getattr(e, source_field, e._definition.get(source_field, None))
         if value is None and source_field in defaults:
             value = defaults[source_field]
         set_in_nested_dict(data, target_field, value)
--- a/probe_scraper/parsers/histograms.py
+++ b/probe_scraper/parsers/histograms.py
@@ -36,6 +36,8 @@ def extract_histogram_data(histogram):
         value = None
         if hasattr(histogram, source_field):
             value = getattr(histogram, source_field)()
+        elif source_field in histogram._definition:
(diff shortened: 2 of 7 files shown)


chutten:
It only now occurs to me that the README doesn't include instructions on how to dry-run locally :(

Let me see if I can figure it out.

fbertsch:
@chutten: just head to the probe-scraper repo and run: `python probe-scraper/runner.py`. If it's taking too long (it will), change `MIN_FIREFOX_VERSION` in `scrapers/moz_central_scraper.py` to e.g. `59` so it doesn't pull down as many files.

It will print out the location of the output json files, which are just the ones uploaded to s3 in the Airflow job.

chutten:
Good thing you asked me to run this as it results in empty bug_numbers lists. I assume this is because `hasattr` is returning false for some reason... I dunno. I'll investigate this tomorrow, and file an Issue to update the README for how to perform a local dry-run.

chutten:
@fbertsch @Dexterp37 RFAL.

Turns out that the types EventData and Histogram didn't have `bug_numbers` properties and, thus, we were getting the default values for them every time. (Scalars were fine). So for events I dug into EventData's raw `_definition` and for Histograms I now store the raw definition on `_definition` so I can pull the same trick.

Dexterp37:
@chutten this looks good to me. Can you verify that adding the `definition` trick doesn't have any side effect on the data generation? (e.g. we now look up stuff in there instead of `defaults`)

chutten:
What sort of side effects did you have in mind? I did a diff -r between two runs (one with and one without my changes). The only differences are the addition of a lot of bug_numbers fields (most of them empty arrays, despite our best efforts) and the updated `lastUpdate` timestamp in `general`

As for updating the architecture of the probe parsing utils, a bug might not be the most appropriate. We have grander plans for them eventually (pulling them into their own project), so a design doc might be better?

fbertsch:
@chutten grand plans for parsing? You have piqued my interest :) design doc sounds great!

Dexterp37:
> What sort of side effects did you have in mind? I did a diff -r between two runs (one with and one without my changes). The only differences are the addition of a lot of bug_numbers fields (most of them empty arrays, despite our best efforts) and the updated lastUpdate timestamp in general

That's ok then! I was looking for differences between the value reported from the getter and the value from the definition. That doesn't seem to be the case, so feel free to merge this once ready :)

chutten:
The "grand plan" so far is "Put the canonical source for probe parsing in its own project and vendor it in to m-c and wherever else it's used. Add capabilities as necessary." ... and I don't think it'll make Q3 planning, though I guess I should add that to the list.

PR #50, opened 2018-06-05
Parse and deploy repositories

fbertsch:
There are a few related changes:
- New classes for parsing `repositories.yaml` and representing
  a repository
- A jsonschema for the `repositories.yaml` file
- A test (so the build fails) for the master `repositories.yaml` file
- Deploy of the `repositories.json` data

That last one will serve as the centralized location of
(appName, os) pairs for all of the ingestion pipeline.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,4 @@ _out/
 _temp/
 
 *.json
+!schemas/*.json
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -1,3 +1,7 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 import json
 import jsonschema
 import yaml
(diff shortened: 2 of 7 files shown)


fbertsch:
@Dexterp37 hmm, that error is not coming up on my machine. I wonder if my machine is using the `pip` installed version of probe-scraper instead of the repo itself. Will investigate.

Dexterp37:
@fbertsch  this looks good to me now, but I can't still run this locally. It might be a problem with my machine, so feel free to merge if you can't replicate the problem locally/on a cluster.

fbertsch:
@Dexterp37 It is a bug, it should work now for you. Give it a shot and let me know.

Dexterp37:
Works for me now, thanks for fixing this. I have one last doubt:

```
$ python probe_scraper/runner.py --only-git-probes

writing output:
  ./mobile_metrics_example/general
  ./mobile_metrics_example/mobile-metrics/all_probes
  ./mobile-metrics/repositories
```

The `repositories` file is on a different directory than the other ones. @fbertsch , is this intended?

fbertsch:
@Dexterp37 and I resolved this in person, `repositories` is in `mobile-metrics` because that's the ping type; `all_probes` goes in `<repo>/<ping-type>/`.

We'll eventually also have a `repositories` file for `main` ping addon scalars, if we'd like.

PR #53, opened 2018-06-08
bug 1467735 - Update parse_events' regex to mozilla-central's

chutten:
This updates it to match the current version in mozilla-central. See https://bugzilla.mozilla.org/show_bug.cgi?id=1467735


Dexterp37:
I'll respin the probe-scraper for today.

Issue #13, opened 2017-04-11
Print some stats before saving probe data to disk


georgf:
To make it more easy to judge whether things worked correctly etc., it would be great to print some basic stats before saving the outputs to disk in runner.py.
E.g.:

- how many probes of each type we have (histograms, scalars, events, ...)
- how many versions/revisions we extracted data for

sannanansari:
I want to Work on this. You can guide me. This is my first bug.

Dexterp37:
Hey @sannanansari , hi and welcome!

One possible approach for doing this is the following:

- Extract the number of revisions/version for each channel by counting the entries in each key of `revisions`, [here](https://github.com/mozilla/probe-scraper/blob/ba84fe98a29dab3c8d0fb94e9dbbaacb7b043575/probe_scraper/runner.py#L123).
- Change the `transform` [method](https://github.com/mozilla/probe-scraper/blob/ba84fe98a29dab3c8d0fb94e9dbbaacb7b043575/probe_scraper/transform_probes.py#L166) so that it counts the number of probes per type, per channel and returns a structure with the data.
- Print all the gathered statistics.

Does this make sense?

sannanansari:
I am getting some error.
ERROR:--
Unable to parse whitelist (C:\Users\Sannan Ansari\Documents\GitHub\probe-scraper\probe_scraper\parsers\third_party\histogram-whitelists.json). Assuming all histograms are acceptable.
Traceback (most recent call last):
  File "runner.py", line 19, in <module>
    from scrapers import git_scraper, moz_central_scraper
  File "C:\Users\Sannan Ansari\Documents\GitHub\probe-scraper\probe_scraper\scrapers\git_scraper.py", line 6, in <module>
    from git import Repo
  File "C:\Python27\lib\site-packages\git\__init__.py", line 85, in <module>
    raise ImportError('Failed to initialize: {0}'.format(exc))
ImportError: Failed to initialize: Bad git executable.
The git executable must be specified in one of the following ways:
    - be included in your $PATH
    - be set via $GIT_PYTHON_GIT_EXECUTABLE
    - explicitly set via git.refresh()

All git commands will error until this is rectified.

This initial warning can be silenced or aggravated in the future by setting the
$GIT_PYTHON_REFRESH environment variable. Use one of the following values:
    - quiet|q|silence|s|none|n|0: for no warning or exception
    - warn|w|warning|1: for a printed warning
    - error|e|raise|r|2: for a raised exception

[...]

PR #55, opened 2018-07-31
Bugs found

sannanansari:
Changes in  bugs finding

### DIFF
--- a/probe_scraper/transform_probes.py
+++ b/probe_scraper/transform_probes.py
@@ -156,7 +156,6 @@ def sorted_node_lists_by_channel(node_data):
                 'node_id': node_id,
                 'version': data['version'],
             })
-
     for channel, data in channels.iteritems():
         channels[channel] = sorted(data, key=lambda n: int(n["version"]), reverse=True)
     return channels
@@ -180,10 +179,9 @@ def transform(probe_data, node_data, break_by_channel):
             print "  from: " + str({"node": node_id, "version": readable_version})
             for probe_type, probes in probe_data[channel][node_id].iteritems():
                 # Group the probes by the release channel, if requested
-                 
+                result_data[probe_type][channel] = {"node": node_id, "version": readable_version} 
                 extract_node_data(node_id, channel, probe_type, probes, result_data,
                                   readable_version, break_by_channel)
-            result_data[channel] = {"node": node_id, "version": readable_version}
     return result_data
 
 


sannanansari:
"All checks have failed" please you can explain about that. Please Review it.

sannanansari:
As I want to save data. So I save whatever is print on screen.
result_data[channel] = {"node": node_id, "version": readable_version}
Channel = key
But actually you about remaining.
Basically I have save whatever is printed on screen

Dexterp37:
Hi @sannanansari ! Just to get a better idea, as Georg mentioned, is this related to #13 ? If not, what is the problem you're trying to fix here? Thanks!

sannanansari:
#13  @Dexterp37  you have comment to:-
# Change transform method so it can count number of probes per type, per channel and returns a structure with the data.
I was working in this way just to save data and return data.
So I just save whatever having being printing on Screen.

Dexterp37:
@sannanansari I see! I can't find where you are counting in the PR you just sent. Was it left out? I'm dropping a few comments on it anyway!

probe_scraper/transform_probes.py
```diff
@@ -156,10 +156,8 @@ def sorted_node_lists_by_channel(node_data):
                 'node_id': node_id,
                 'version': data['version'],
             })
-
```
Dexterp37: Let's not remove blank lines :)

probe_scraper/transform_probes.py
```diff
@@ -156,10 +156,8 @@ def sorted_node_lists_by_channel(node_data):
...
     for channel, data in channels.iteritems():
         channels[channel] = sorted(data, key=lambda n: int(n["version"]), reverse=True)
-
```
Dexterp37: Let's not remove blank lines :)

probe_scraper/transform_probes.py
```diff
@@ -172,19 +170,18 @@ def transform(probe_data, node_data, break_by_channel):
            release channel.
     """
     channels = sorted_node_lists_by_channel(node_data)
-
-    result_data = {}
+    result_data =defaultdict(dict)
```
Dexterp37: Why did you change this part?

probe_scraper/transform_probes.py
```diff
@@ -172,19 +170,18 @@ def transform(probe_data, node_data, break_by_channel):
...
     for channel, channel_data in channels.iteritems():
-        print "\n" + channel + " - transforming probe data:"
+        print "\n" + str(channel) + " - transforming probe data:"
```
Dexterp37: Why did you change this part?

probe_scraper/transform_probes.py
```diff
@@ -172,19 +170,18 @@ def transform(probe_data, node_data, break_by_channel):
...
         for entry in channel_data:
             node_id = entry['node_id']
             readable_version = entry["version"]
             print "  from: " + str({"node": node_id, "version": readable_version})
             for probe_type, probes in probe_data[channel][node_id].iteritems():
                 # Group the probes by the release channel, if requested
+                result_data[probe_type][channel] = {"node": node_id, "version": readable_version} 
                 extract_node_data(node_id, channel, probe_type, probes, result_data,
                                   readable_version, break_by_channel)
-
```
Dexterp37: Let's not remove blank lines :)

probe_scraper/transform_probes.py
```diff
@@ -172,19 +170,18 @@ def transform(probe_data, node_data, break_by_channel):
...
     for channel, channel_data in channels.iteritems():
-        print "\n" + channel + " - transforming probe data:"
+        print "\n" + str(channel) + " - transforming probe data:"
         for entry in channel_data:
             node_id = entry['node_id']
             readable_version = entry["version"]
             print "  from: " + str({"node": node_id, "version": readable_version})
             for probe_type, probes in probe_data[channel][node_id].iteritems():
                 # Group the probes by the release channel, if requested
+                result_data[probe_type][channel] = {"node": node_id, "version": readable_version}
```
Dexterp37: You can create a new dictionary instead of reusing `result_data` and just use the new one for keeping track/counting the probes.

probe_scraper/transform_probes.py
sannanansari: I think that default dict is perfect because it won't give key error.

probe_scraper/transform_probes.py
sannanansari: I have just change it. Not any Reason behind it.

probe_scraper/transform_probes.py
sannanansari: OK. I will work on it.

sannanansari: OK. Means I have Done Wrong this line.
**result_data[probe_type][channel] = {"node": node_id, "version": readable_version}**
You can Review it.

probe_scraper/transform_probes.py
Dexterp37: In general, if there isn't any specific reason to change a line, then it's better to leave it alone :) That makes it less likely to add new bugs to the code!

probe_scraper/transform_probes.py
Dexterp37: @sannanansari , I would expect this line to become something like:

```
probe_counts[channel][probe_type] += 1
```

Dexterp37:
Let's close this, it appears you pushed another PR (#57 ). Let's take the conversation there.

PR #57, opened 2018-08-16
Code

sannanansari:
screenshot 2

### DIFF
--- a/probe_scraper/transform_probes.py
+++ b/probe_scraper/transform_probes.py
@@ -172,7 +172,7 @@ def transform(probe_data, node_data, break_by_channel):
            release channel.
     """
     channels = sorted_node_lists_by_channel(node_data)
-
+    new_data = {}
     result_data = {}
     for channel, channel_data in channels.iteritems():
         print "\n" + channel + " - transforming probe data:"
@@ -182,10 +182,14 @@ def transform(probe_data, node_data, break_by_channel):
             print "  from: " + str({"node": node_id, "version": readable_version})
             for probe_type, probes in probe_data[channel][node_id].iteritems():
                 # Group the probes by the release channel, if requested
+                if probe_type not in result_data:
+                    result_data[probe_type] = 1
+                else:
+                    result_data[probe_type] +=1
                 extract_node_data(node_id, channel, probe_type, probes, result_data,
-                                  readable_version, break_by_channel)
-
-    return result_data
+                                 readable_version, break_by_channel)
+        new_data[channel] = result_data
+    return new_data
 
 
 def make_commit_hash_probe_definition(definition, commit):


sannanansari:
You can review it.

sannanansari:
**The command "flake8 --max-line-length 100 ." failed and exited with 1 during .
What this Error. Explain?

Dexterp37:
> `**The command "flake8 --max-line-length 100 ." failed and exited with 1 during .`
> What this Error. Explain?

This means that some changes you made make some line longer than 100 characters, which is the maximum allowed.

sannanansari:
Sorry I don't know about limit. By this I can assume I have Done my work. Not completely but I can assume it.

probe_scraper/transform_probes.py
```diff
@@ -182,10 +182,14 @@ def transform(probe_data, node_data, break_by_channel):
             print "  from: " + str({"node": node_id, "version": readable_version})
             for probe_type, probes in probe_data[channel][node_id].iteritems():
                 # Group the probes by the release channel, if requested
+                if probe_type not in result_data:
```
Dexterp37: The requirements from the [issue](https://github.com/mozilla/probe-scraper/issues/13#issuecomment-405236410) are a bit more involved than this: we should count, in this function, the number of probe types per channel.

Also, all the counts must be stored in `probes_counts`, not in `result_data`. The latter must not be modified.

probe_scraper/transform_probes.py
```diff
@@ -182,10 +182,14 @@ def transform(probe_data, node_data, break_by_channel):
...
+                    result_data[probe_type] = 1
+                else:
+                    result_data[probe_type] +=1
                 extract_node_data(node_id, channel, probe_type, probes, result_data,
-                                  readable_version, break_by_channel)
-
-    return result_data
+                                 readable_version, break_by_channel)
+        new_data[channel] = result_data
+    return new_data
```
Dexterp37: We should return both `result_data` and `probes_counts` here.

probe_scraper/transform_probes.py
```diff
@@ -182,10 +182,14 @@ def transform(probe_data, node_data, break_by_channel):
...
+                if probe_type not in result_data:
+                    result_data[probe_type] = 1
+                else:
+                    result_data[probe_type] +=1
                 extract_node_data(node_id, channel, probe_type, probes, result_data,
-                                  readable_version, break_by_channel)
-
-    return result_data
+                                 readable_version, break_by_channel)
+        new_data[channel] = result_data
```
Dexterp37: This is not needed if we return both the data structures.

probe_scraper/transform_probes.py
sannanansari: No. I'm getting this output only.

sannanansari: The requirements from the issue are a bit more involved than this: we should count, in this function, the number of probe types per channel.
Also, all the counts must be stored in probes_counts, not in result_data. The latter must not be modified.

**OK. I will for that.

probe_scraper/transform_probes.py
sannanansari: ok.

probe_scraper/transform_probes.py
sannanansari: But it's giving Error at runner.py "'tuple' object does not support item assignment"
I thing is better idea to make one before returning it.

probe_scraper/transform_probes.py
sannanansari: Output was diffrent because I was checking that I have done correct or not. Sobefore returning I just print it(for checking).

Dexterp37: Issue #13 requires us to print the statistics as well, so let's print them at the end of the transformation, around the top level caller of this function.

probe_scraper/transform_probes.py
Dexterp37: You probably need to modify the call site in order to adjust for the tuple being returned. Let's use two objects, please.

sannanansari: screenshot 6
screenshot 7 _li

I know you want but first you see it.

Difference is that if will return  whole probe_counts after calculating as you can see 2 picture(I have mark 1 and 2 note: do not consider the dictionary just see values in last). The whole probe_counts will calculate all scalar,histogram,event sum and return it. 
If we will store in on dictionary is better(You can see 2 picture I have mark 2). It seems better than first.
What I understand I have said now you can guide?
And you can change in value in both histograms.

Dexterp37: I'm not entirely sure I understand what you mean, but what we want is to return data from the function and then find a way to present it: that shouldn't be a problem, regardless of how many return values we have.

fbertsch:
Closing this just to clean up the open PRs. @sannanansari, if you'd like to work on this more, please let us know! We're happy to have your help, and can re-open this PR.

Issue #42, opened 2018-04-09
Probe Scraper should include a list of buckets for each histogram


fbertsch:
related: https://github.com/mozilla/telemetry-dashboard/issues/531

ab0092:
@fbertsch can i get some more details so that i can start working on this ?

fbertsch:
@ab0092 definitely! Basically, histograms are a collection of buckets, and the count of the number of values we saw within that bucket. You can see how the definitions [0] map to a number of buckets in [1], which simulates histograms (for example, an exponential histogram with 4 buckets ending at 100 has buckets `[0, 1, 51, 100]`).

It would be helpful to have that information directly available in the probe-scraper, so that we don't have to generate it from the definitions. The exact code to use is here [2], which includes linear and exponential histograms. You can use the `python_moztelemetry` library directly to retrieve the buckets and have them included in the probe-scraper's definitions. You'll need to update the code in [3] and [4] to account for the new field.

[0] Definitions: https://dxr.mozilla.org/mozilla-central/source/toolkit/components/telemetry/Histograms.json
[1] https://telemetry.mozilla.org/histogram-simulator/#low=1&high=100&n_buckets=4&kind=linear&generate=normal
[2] Code to use, given the histogram definition: https://github.com/mozilla/python_moztelemetry/blob/master/moztelemetry/parse_histograms.py
[3] https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/parsers/histograms.py#L10
[4] https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/transform_probes.py

Issue #26, opened 2018-01-16
Extract `record_in_processes` from Histograms.json


Dexterp37:
Information about in which process a particular histogram is recorded is currently not extracted by the probe scraper.

PR #60, opened 2019-01-11
Scrape processes, alert_emails, and bucket enumeration

fbertsch:
Also fixes a bug in the histogram_tools that used `definition`
instead of `_definition`.

Tested this locally, output seems to be good.

### DIFF
--- a/probe_scraper/parsers/histograms.py
+++ b/probe_scraper/parsers/histograms.py
@@ -21,6 +21,7 @@ def extract_histogram_data(histogram):
         "keyed": "details/keyed",
         "kind": "details/kind",
         "record_in_processes": "details/record_in_processes",
+        "ranges": "details/buckets",
     }
 
     defaults = {
@@ -29,6 +30,7 @@ def extract_histogram_data(histogram):
         "expiration": "never",
         "bug_numbers": [],
         "alert_emails": [],
+        "ranges": [],
     }
 
     data = {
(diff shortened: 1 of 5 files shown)


chutten:
Do... do we want emails?

fbertsch:
@chutten definitely, it would help a lot of people to get name or email in the probe-dictionary, so they know who to ask questions to. Should make our lives a bit easier too!

fbertsch:
hmm, tests are failing, and it looks like the output does not actually have `notification_emails`. Investigating.

chutten:
It's called `alert_emails` in Histograms, I think.

I wonder if publishing emails on the probe info service and probe dictionary is any different than having them published on hg.mo

fbertsch:
Okay, squashed the commits. The output has the expected fields now, the tests pass.

fbertsch:
@chutten do you mean a privacy concern? Or..?

fbertsch:
Fixes #42 and #26

chutten:
Privacy not so much as ease of scraping. Which might not be something we worry about in 2019 as much as I remember worrying about it in 2009

fbertsch:
I'd wager that the `histograms.json` and `scalars.yaml` files are more likely to be scraped than the probe-info-service data, since they are in moz-central.

chutten:
Fair enough.

Dexterp37:
Since the privacy discussion has settled, I think we can merge this :) LGTM!

PR #63, opened 2019-01-24
Revert "Add buckets enumeration to output"

badboy:
This reverts commit 665b1970a49f3459e60b54c11d80171d96283409.

Old `Histograms.json` contains things like "10000" or "16 * 1024 * 1024"
for fields like `low`, `high` or `n_values` (e.g. strings & expressions,
not numbers). See #9.

Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1521742


Dexterp37:
Makes sense.

PR #64, opened 2019-01-24
[WIP] Glean support

Dexterp37:
This is the super basic thing that allows calling a glean parser from the scraper. The idea was to use the real  [glean parser](https://github.com/mozilla/glean_parser/) to get the metric info out. Ideally, one could write a new template for the probe-scraper format (like [this](https://github.com/mozilla/glean_parser/tree/master/glean_parser/templates) for the Kotlin code) and let the parser.. generate. @mdboom might know more about this or suggest better ways! :)


Dexterp37:
FYI @fbertsch !

fbertsch:
Now being done in https://github.com/mozilla/probe-scraper/pull/72

PR #66, opened 2019-02-12
Add `first_added` date to probe-info-service

### LINKED ISSUE #65: Include "first_added" dates in probe-info-service
## Description of Problem

For the [Main Ping Processing](https://docs.google.com/document/d/1oPp_J4AoB3llgAe5Gp1SVPtkA2xUjZR5TKL0CVOZKx8/edit#) in GCP, we plan on including every historical probe, so that they can continue to be queried even after they expire.

To do this, we need a total ordering or probes, with new probes appended. For example, if the current set of probes was `{SEARCH_COUNTS, GC_MS}`, but then tomorrow's nightly we added `{PAGELOAD_IS_SSL}`, I would want the probe ordering to be:

Today's: [GC_MS, SEARCH_COUNTS]
Tomorrow's: [GC_MS, SEARCH_COUNTS, PAGELOAD_IS_SSL]

## `first_added` field

The easiest way to accomplish this is to include a "first_added" timestamp. If any probes share a "first_added" timestamp, they will be sorted alphanumerically. Some probes (e.g. `STYLO_PARALLEL_RESTYLE_FRACTION_WEIGHTED_STYLED`) only showed up on release, and others (e.g. `GC_MS`) only show up on pre-release. As such we can include a "first_added" date for each channel:
```
"GC_MS": {
   "history": {...},
    "name": "GC_MS",
    "type": "histogram",
    "first_added": {
        "nightly": 1547013850,
        "beta": 1547014850
     }
}
```

We can achieve this by:
1. Scraping all revisions from each probe file, e.g. [Histograms.json](https://hg.mozilla.org/mozilla-central/log/tip/toolkit/components/telemetry/Histograms.json)
2. Iterate over each revision, getting list of new probes added in that revision
3. Getting the `pushdate` from [json-rev](https://hg.mozilla.org/mozilla-central/json-rev/tip/).
4. Joining that data with the existing probe-info-service data

The only piece I don't know how to do is: given the revision, get the associated json-rev file.

fbertsch:
Fixes #65

There are quite a few related changes here:
* Buildhub scraper gets data from the BH API
    - pagination of results included
* moz-central-scraper scrapes the revisions
    - made versions an int until needs to be a str
    - pulled out some behavior into new functions
    - added tests for some behavior
* runner retrieves buildhub-revision data separately
    - merges the first-added dates with the probe-info-service data

New tests:
The new tests include network connection, and are marked as such.
New pytest files disable running these tests automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ The data is pulled from two different sources:
 
 A web tool to explore the data is available [here](https://telemetry.mozilla.org/probe-dictionary/).
 
+
 ## Adding a New Git Repository
 
 To scrape a git repository for probe definitions, an entry needs to be added in `repositories.yaml`.
@@ -36,15 +37,26 @@ pip install -r test_requirements.txt
 python setup.py develop
 ```
 
-Run tests:
+Run tests. This by default does not run tests that require a web connection:
 ```
-pytest
+pytest tests/
+```
+
+To run all tests, including those that require a web connection:
+```
+pytest tests/ --run-web-tests
 ```
 
 To test whether the code conforms to the style rules, you can run:
 ```
 flake8
 ```
+
+### Tests with Web Dependencies
+
+Any tests that require a web connection to run should be marked with `@pytest.mark.web_dependency`.
+
+These will not run by default, but will run on CI.
 
 ### Performing a Dry-Run
 
@@ -69,6 +81,7 @@ The code layout consists mainly of:
 - `probe_scraper`
   - `runner.py` - the central script, ties the other pieces together
   - `scrapers`
+     - `buildhub.py` - pull build info from the [BuildHub](https://buildhub.moz.tools) service
      - `moz_central_scraper.py` - loads probe registry files for multiple versions from mozilla-central
      - `git_scraper.py` - loads probe registry files from a git repository (no version or channel support yet, just per-commit)
   - `parsers/` - extract probe data from the registry files
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@ The data is pulled from two different sources:
 
 A web tool to explore the data is available [here](https://telemetry.mozilla.org/probe-dictionary/).
 
-
 ## Adding a New Git Repository
 
 To scrape a git repository for probe definitions, an entry needs to be added in `repositories.yaml`.
--- a/probe_scraper/parsers/events.py
+++ b/probe_scraper/parsers/events.py
@@ -60,7 +60,7 @@ class EventsParser:
         # Events.yaml had a format change in 53, see bug 1329620.
         # We don't have important event usage yet, so lets skip
         # backwards compatibility for now.
-        if version and int(version) < 53:
+        if version and version < 53:
             return {}
 
         if len(filenames) > 1:
--- a/probe_scraper/parsers/events.py
+++ b/probe_scraper/parsers/events.py
@@ -56,11 +56,13 @@ def extract_events_data(e):
 
 
 class EventsParser:
-    def parse(self, filenames, version=None):
+    def parse(self, filenames, version=None, channel=None):
         # Events.yaml had a format change in 53, see bug 1329620.
         # We don't have important event usage yet, so lets skip
         # backwards compatibility for now.
-        if version and version < 53:
+        if (version and channel) and (
+          ((channel != "nightly" and version < 53)
+           or (channel == "nightly" and version < 54))):
             return {}
 
         if len(filenames) > 1:
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -124,7 +124,7 @@ def parse_moz_central_probes(scraped_data):
 
 
 def add_first_appeared_dates(probes_by_channel, first_appeared_dates):
-    for channel, probes in list(probes_by_channel.items()):
+    for channel, probes in probes_by_channel.items():
         for probe_id, info in probes.items():
             if channel == "all":
                 dates = first_appeared_dates[probe_id]
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -117,7 +117,7 @@ def parse_moz_central_probes(scraped_data):
     for channel, revisions in scraped_data.iteritems():
         for revision, details in revisions.iteritems():
             for probe_type, paths in details["registries"].iteritems():
-                results = PARSERS[probe_type].parse(paths, details["version"])
+                results = PARSERS[probe_type].parse(paths, details["version"], channel)
                 probes[channel][revision][probe_type] = results
 
     return probes
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -114,9 +114,9 @@ def parse_moz_central_probes(scraped_data):
     """
 
     probes = defaultdict(lambda: defaultdict(lambda: defaultdict(int)))
-    for channel, revisions in scraped_data.iteritems():
-        for revision, details in revisions.iteritems():
-            for probe_type, paths in details["registries"].iteritems():
+    for channel, revisions in scraped_data.items():
+        for revision, details in revisions.items():
+            for probe_type, paths in details["registries"].items():
                 results = PARSERS[probe_type].parse(paths, details["version"], channel)
                 probes[channel][revision][probe_type] = results
 
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -99,8 +99,8 @@ def write_repositories_data(repos, out_dir):
 def parse_moz_central_probes(scraped_data):
     """
     Parse probe data from files into the form:
-    channel_name -> {
-      node_id -> {
+    channel_name: {
+      node_id: {
         histogram: {
           name: ...,
           ...
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -30,7 +30,8 @@ class DummyParser:
 
 FROM_EMAIL = "telemetry-alerts@mozilla.com"
 DEFAULT_TO_EMAIL = "dev-telemetry-alerts@mozilla.com"
-
+FIRST_APPEARED_DATE_KEY = "first_added"
+DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
 
 PARSERS = {
     # This lists the available probe registry parsers:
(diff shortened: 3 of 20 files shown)


fbertsch:
@georgf there are some issues here. I included new CLI options for running on a single version, but when doing so the versions don't match correctly [because of this logic](https://github.com/mozilla/probe-scraper/blob/5e574da6207894fec50b73dfdd973a4fe02f6305/probe_scraper/scrapers/moz_central_scraper.py#L91). I would like to get your input on this edge case. In the larger sense, when running over all revisions, this won't be an issue. The reason is because there won't be any "missing" revisions, that the `json-tags` data includes but the `Buildhub` data doesn't.

Another is the length of time this takes to run. We're looking at downloading thousands of files per-run. My first thought at optimization is to update the job in `telemetry-airflow` to zip the files-cache, upload it to s3, and download it before running. That way we only download all of the files once, and incrementally update the cache from then-on.

Let me know if you run into any issues running the tests:
```
pytest tests/ --run-web-tests
```

or the probe-scraper itself, e.g. (for a single version, with the problem I described above):
```
python probe_scraper/runner.py --firefox-version 64 --only-moz-central-probes
```

probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -62,39 +63,67 @@ def load_tags(channel):
...
             })
 
-    results = sorted(results, key=lambda r: int(r["version"]))
+    results = sorted(results, key=lambda r: r["version"])
 
-    # Add tip revision.
-    if tip_node_id != results[-1]["node"]:
-        latest_version = str(int(results[-1]["version"]) + 1)
+    # Add tip revision, if we're including the most recent version
+    if (tip_node_id != results[-1]["node"] and
```
fbertsch: Would also like your input on this version logic, now that I've included the `max_fx_version` parameter.

probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -41,7 +42,7 @@
     },
 }
 
-MIN_FIREFOX_VERSION = 30
+MIN_FIREFOX_VERSION = 64
```
fbertsch: Testing artifact, will remove.

fbertsch:
The failing build seems unrelated. The tests passed but the build failed to terminate, resulting after 10 minutes in a failed build because CI thought it stalled. I've restarted it, let's see if it passes.

fbertsch:
@Dexterp37 I also think that eventually we can remove the `json-tags` scraping altogether, and get nodes (revisions) only from Buildhub.

README.md
```diff
@@ -11,6 +11,7 @@ The data is pulled from two different sources:
 
 A web tool to explore the data is available [here](https://telemetry.mozilla.org/probe-dictionary/).
 
+
```
Dexterp37: nit: was this intentional?

probe_scraper/parsers/events.py
```diff
@@ -60,7 +60,7 @@ def parse(self, filenames, version=None):
         # Events.yaml had a format change in 53, see bug 1329620.
         # We don't have important event usage yet, so lets skip
         # backwards compatibility for now.
-        if version and int(version) < 53:
+        if version and version < 53:
```
Dexterp37: Why are we removing `int(version)` ?

tests/test_buildhub.py
```diff
@@ -0,0 +1,137 @@
+
+import pytest
+from probe_scraper.scrapers.buildhub import Buildhub, NoDataFoundException
+from datetime import datetime
+
+FX_RELEASE_62_0_3 = {
+    "revision": "c9ed11ae5c79df3dcb69075e1c9da0317d1ecb1b",
+    "date": datetime(2018, 10, 01, 18, 40, 35),
```
Dexterp37: nit: I think the leading 0 in 01 can be removed

tests/test_moz_central_scraper.py
```diff
@@ -0,0 +1,62 @@
...
+    channel = "release"
+    revision = "c9ed11ae5c79df3dcb69075e1c9da0317d1ecb1b"
+
+    registries = {
+        probe_type: [os.path.join(tmp_dir, "hg", revision, path) for path in paths]
+        for probe_type, paths in moz_central_scraper.REGISTRY_FILES.items()
+    }
+
+    record = {
+        "date": datetime(2018, 10, 01, 15, 55, 45),
```
Dexterp37: nit: leading 0 in 01

probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -62,39 +63,67 @@ def load_tags(channel):
     return data
 
 
-def extract_tag_data(tag_data, channel):
+def extract_major_version(version_str):
```
Dexterp37: Can we add test coverage for this bunch of functions?

probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -62,39 +63,67 @@ def load_tags(channel):
...
-        # need this.
-        if channel != "nightly":
-            version = str(int(version) - 1)
-
-        if int(version) >= MIN_FIREFOX_VERSION:
+        version = extract_tag_version(channel, tag["tag"])
+        version = adjust_version(channel, version)
+        latest_version = max(version, latest_version)
+
+        if (version >= min_fx_version and
```
Dexterp37: Are we sure `version `is an integer?

probe_scraper/scrapers/buildhub.py
```diff
@@ -0,0 +1,159 @@
...
+                d = d[:-6]
+
+            date = None
+            try:
+                date = datetime.strptime(d, Buildhub.date_formats[0])
+            except ValueError:
+                pass
+
+            if date is None:
+                date = datetime.strptime(d, Buildhub.date_formats[1])
```
mdboom: Maybe put this parsing in its own function so you could just return from a for loop on the first format that works?

Also, you should probably use `self.date_formats` here -- no need to name the class by name.

probe_scraper/parsers/events.py
fbertsch: One of the changes I made was to make version an integer until it needs to be cast to a string later in the job. That simplified a few of the instances of version comparison.

fbertsch: Same as below, version is now an int.

probe_scraper/runner.py
```diff
@@ -95,29 +96,55 @@ def write_repositories_data(repos, out_dir):
     dump_json(json_data, os.path.join(out_dir, "mobile-metrics"), "repositories")
 
 
-def load_moz_central_probes(cache_dir, out_dir):
-    # Scrape probe data from repositories.
-    node_data = moz_central_scraper.scrape(cache_dir)
+def parse_moz_central_probes(scraped_data):
+    """
+    Parse probe data from files into the form:
+    channel_name -> {
```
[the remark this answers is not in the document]
fbertsch: Nope, I'll update this

probe_scraper/scrapers/moz_central_scraper.py
fbertsch: I did! Check `test_extract_major_version`

probe_scraper/parsers/events.py
Dexterp37: I see, thank you :)

Dexterp37:
> 
> 
> @Dexterp37 I also think that eventually we can remove the `json-tags` scraping altogether, and get nodes (revisions) only from Buildhub.

Any chance you could create an issue here for this?

Also, CI failed again :(

probe_scraper/scrapers/moz_central_scraper.py
Dexterp37: The logic looks sound to me. We're basically keeping track of it in this loop instead of referring to it as `results[-1]["version"] + 1`, right?

fbertsch:
@Dexterp37 quick update: This is failing because I'm retrieving [some nightly files from 53 that don't have the update events format.](https://hg.mozilla.org/mozilla-central/file/13736e2db6eb94b02dd28cc88f2943b8109aa374/toolkit/components/telemetry/Events.yaml)

I'm assuming the `json-tags` endpoint doesn't contain these revisions, and that's why we're only now running into this.

Dexterp37:
> 
> 
> @Dexterp37 quick update: This is failing because I'm retrieving [some nightly files from 53 that don't have the update events format.](https://hg.mozilla.org/mozilla-central/file/13736e2db6eb94b02dd28cc88f2943b8109aa374/toolkit/components/telemetry/Events.yaml)
> 
> I'm assuming the `json-tags` endpoint doesn't contain these revisions, and that's why we're only now running into this.

I see. Thanks for the update.

fbertsch:
Now [this revision isn't parsable because the field exceeds the 20 character limit.](https://hg.mozilla.org/mozilla-central/file/3cfc350101967376909ad3c729f9779ae0ab7a94/toolkit/components/telemetry/Events.yaml#l527).

fbertsch:
Rebasing on Python 3, and will incorporate changes requested in the reviews. Still running on existing data and finding incompatible files. Generally I'm just eliminating the checks when `strict_type_checks == False`.

probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -62,39 +63,67 @@ def load_tags(channel):
...
+    extract the major version as an int.
+    """
+    search = re.search(r"^(\d+)\.", version_str)
+    if search is not None:
+        return int(search.group(1))
+    else:
+        raise Exception("Invalid version string " + version_str)
+
+
+def extract_tag_version(channel, version_str):
```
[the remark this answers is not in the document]
fbertsch: On it!

probe_scraper/scrapers/buildhub.py
fbertsch: My only hesitancy with using a loop was I want the last element to fail if it doesn't parse. That will indicate to future us that they changed the format of the date string, and with only two elements it seems simple enough to unroll?

fbertsch:
Okay, this ran successfully and the output was the same as the current `master` branch (with the addition of the `first_added` dates).

One _minor_ difference: Because we are now no longer scraping the nightly 53 events files, `event/navigation.search#search` now has a first revision of `6583496f169cd8a13c531ed16e98e8bf313eda8e` and first version of `54` for nightly. This seems acceptable to me.

Dexterp37:
> 
> 
> Okay, this ran successfully and the output was the same as the current `master` branch (with the addition of the `first_added` dates).

Great! Glad to hear this!

> One _minor_ difference: Because we are now no longer scraping the nightly 53 events files, `event/navigation.search#search` now has a first revision of `6583496f169cd8a13c531ed16e98e8bf313eda8e` and first version of `54` for nightly. This seems acceptable to me.

It is to me as well. Feel free to merge when you see fit, @fbertsch

Issue #67, opened 2019-02-13
Change existing git repo parsing to Glean parsing


fbertsch:
We currently have a setup to include the types of files we expect for Desktop: `Histograms.json`, `Scalars.yaml`, etc., but for Github repos. The Glean work is replacing that effort, so we need to update the code to parse `metrics.yaml` files instead.

This work will encompass:
- Parsing and writing out information on Glean probes to the probe-dictionary
- Parsing and writing out information on Glean pings to the probe-dictionary

The former will be used for schema creation. The latter will be used for validation and table names in the pipeline.

We can use the [Glean Parser](https://mozilla.github.io/glean_parser) library to parse the `metrics.yaml` files and [write out results](https://github.com/mozilla/glean_parser/blob/master/glean_parser/metrics.py#L52). The rationale for letting the Glean Parser serialize is then the scraper doesn't need to have intimate knowledge of which fields are required and which may be updated; instead we can just add and remove fields from the parser.

PR #72, opened 2019-02-25
Glean Support

fbertsch:


fbertsch:
This does not yet include parsing of `Glean`-provided data, that we need to scrape from https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/metrics.yaml.

We need to work out a solid plan of attack for that `metrics.yaml` and others that may be slurped in by the application.

fbertsch:
Addresses #67

probe_scraper/runner.py
```diff
@@ -78,21 +79,20 @@ def write_moz_central_probe_data(probe_data, revisions, out_dir):
...
-        base_dir = os.path.join(out_dir, repo)
+        base_dir = os.path.join(out_dir, "glean", repo)
 
         print("\nwriting output:")
-        dump_json(general_data(), base_dir, "general")
 
-        data_dir = os.path.join(base_dir, "mobile-metrics")
-        dump_json(probe_data, data_dir, "all_probes")
+        dump_json(general_data(), base_dir, "general")
+        dump_json(probe_data, base_dir, "metrics")
```
fbertsch: Removes a bit of nesting, now we'll have:
```
/glean/fenix/metrics
/glean/fenix/general
/glean/reference-browser/metrics
/glean/reference-browser/general
```

probe_scraper/runner.py
```diff
@@ -130,58 +130,54 @@ def load_moz_central_probes(cache_dir, out_dir):
...
-            }
+            And(str, lambda x: len(x) == 40): [And(str, lambda x: os.path.exists(x))]
         }
     })
 
     schema.validate(data)
 
 
-def load_git_probes(cache_dir, out_dir, repositories_file, dry_run):
+def load_glean_metrics(cache_dir, out_dir, repositories_file, dry_run):
```
fbertsch: I removed all interim objects that had to worry about `probe-type`, because we only have 1: `metric`. While this may seem confusing at first, I think in general it clears up what this code is trying to do.

probe_scraper/runner.py
```diff
@@ -130,58 +130,54 @@ def load_moz_central_probes(cache_dir, out_dir):
...
-                except Exception:
-                    msg = "Improper file in {}\n{}".format(', '.join(paths), traceback.format_exc())
+    metrics = defaultdict(lambda: defaultdict(lambda: defaultdict(int)))
+    for repo_name, commits in repos_metrics_data.items():
+        for commit_hash, paths in commits.items():
+            try:
+                results, errs = GLEAN_PARSER.parse(paths)
+                metrics[repo_name][commit_hash] = results
+            except Exception:
+                msg = "Improper file in {}\n{}".format(', '.join(paths), traceback.format_exc())
```
fbertsch: It may be worth considering a bit more deeply what to do with emails (if anything). In general I'd expect any issues with a `metrics.yaml` file to arise during build, because we're all using the `glean_parser`; in that case, we can perhaps just fail hard (for a specific repo).

probe_scraper/transform_probes.py
```diff
@@ -189,68 +190,85 @@ def transform(probe_data, node_data, break_by_channel):
...
         definition[COMMITS_KEY]["last"] = commit
+        definition[DATES_KEY]["last"] = commit_timestamps[commit]
 
     return definition
 
 
-def update_or_add_probe(all_probes, repo_name, commit_hash, probe, probe_type, definition):
-    probe_id = get_probe_id(probe_type, probe)
+def metrics_equal(def1, def2):
+    return all((
```
mdboom: I think because these are `dataclasses`, you can just do `def1 == def2`.  Did you try that and it didn't work?

fbertsch: No, I didn't! That sounds fantastic.

fbertsch:
Added Glean to `repositories.yaml`, so it now scrapes the actual `metrics.yaml` file. There is an unresolved issue on how to solve dependent definitions but this is a good start for scraping the data.

fbertsch:
@mdboom right. There's no need to scrape those legacy definition files so I removed those pieces as much as possible.

probe_scraper/parsers/metrics.py
```diff
@@ -0,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+from glean_parser.parser import parse_metrics
+from pathlib import Path
+
+
+class MetricsParser:
```
Dexterp37: nit: can we call this something like `GleanMetricsParser` or, at least, mention that's used for glean in the docstring for this class?

probe_scraper/runner.py
Dexterp37: Would you mind documenting this in the [readme](https://github.com/mozilla/probe-scraper#accessing-the-data-files)? I think the `README` requires an update for both the file format and the REST-like API format for glean data.

probe_scraper/runner.py
Dexterp37: Since we're talking about GH repos, I wonder if there's an API to file issues :) (definitely not blocking, just a loud thought)

mdboom:
Dexterp37 wrote:
> Since we're talking about GH repos, I wonder if there's an API to file issues :) (definitely not blocking, just a loud thought)

There is an API, and a Python library (used it before for some things).  That would be killer (but I agree shouldn't hold up this PR).

fbertsch:
@Dexterp37, this is RFAL. The readme has been updated and tests are now passing.

probe_scraper/transform_probes.py
fbertsch: This didn't work because I'm calling `.serialize()` on the objects before it gets to this point :/. What would be ideal, I think, is to extend those classes with ones that include the new fields, add a comparator in that class, and use that here.

In general though this library could use a lot of refactoring...

mdboom: This solution here is fine given all that...

fbertsch: I've opened https://github.com/mozilla/probe-scraper/issues/74 to call attention to a refactoring effort. If you find some time it may be helpful to write some thoughts on what you think a good refactor would look like here.

PR #73, opened 2019-02-27
Order revisions by date when possible

fbertsch:
The revisions we get from Buildhub have the published date,
which is the sort order we want when getting the earliest and
latest revisions (rather than sorting by version).


fbertsch:
@Dexterp37 I'm not sure it's necessary, considering the default behavior hasn't changed? All this guarantees is that what we _thought_ are the earliest/latest dates actually _are_.

PR #77, opened 2019-03-15
Mark use counters as enabled on release (opt-out)

badboy:
This was done on desktop in [bug 1510566](https://bugzilla.mozilla.org/show_bug.cgi?id=1510566), but we forgot to import it here
as well.

### DIFF
--- a/probe_scraper/parsers/histograms.py
+++ b/probe_scraper/parsers/histograms.py
@@ -52,7 +52,7 @@ def extract_histogram_data(histogram, version):
 
     # Use Counters are shipped on release since 65.
     # If the parsers would set this flag, we couldn't differentiate between versions.
-    if version >= "65":
+    if int(version) >= 65:
         if histogram.name().startswith("USE_COUNTER2_"):
             optout = True
 
--- a/probe_scraper/parsers/histograms.py
+++ b/probe_scraper/parsers/histograms.py
@@ -6,7 +6,7 @@ from .third_party import histogram_tools
 from .utils import set_in_nested_dict, get_major_version
 
 
-def extract_histogram_data(histogram):
+def extract_histogram_data(histogram, version):
     props = {
         # source_field: target_field
         "cpp_guard": "cpp_guard",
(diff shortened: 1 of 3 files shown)


Dexterp37:
Interesting. Some failures are unrelated here, e.g. the ones related to the git repository and the required 'expired' key. Let me fix this in a separate PR.

badboy:
Yeah, looks like the glean parser recently turned `expires` into a required attribute. Had the same test failures locally.
Leaving that for you to fix here. :)

Dexterp37:
#78 was merged. You're good to go!

georgf:
Cheers for following up!
Note that the use counters are only enabled on release from Firefox 65+.
Is this reflected in the data here?

fbertsch:
@badboy @Dexterp37 do we still want to get this in?

badboy:
Yes, let's get that in. I'm not sure how the probe-scraper determines the first version a probe was introduced in though. Georg is right that we enabled release use counters in 65. Is there any way we can add a small special case into the code there?

fbertsch:
@badboy just to be clear: you'd like all use-counters to have a definition from _before_ 65 that marks it as `opt-in`, and a definition _after_ 65 that marks it as `opt-out` (for the ones that were available before 65)?

badboy:
that would be the technically correct solution, yes.

fbertsch:
Okay, it should be relatively straightforward. It will happen [in this piece](https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/parsers/histograms.py#L9). You will need to add the `version` as a parameter to that function, from [here](https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/parsers/histograms.py#L71).

Then in the [code that checks for opt-in/opt-out](https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/parsers/histograms.py#L48), check if it's a use-counter and if it's before/after 65.

badboy:
He, I even ran it across a wider range! Looks good though:

fbertsch:
That's great! Let's get this in.

PR #78, opened 2019-03-15
Update the test files to add the required 'expires' key

Dexterp37:
The glean scraper were updated, and so we broke this consumer. Nice? No :(  I wonder if:

- we should version pin the parsers, so that don't break on new releases;
- have a "relaxed" mode in the parsers, like the desktop ones, which doesn't fail hard and is happy to ingest historical data that might not have all the latest properties.

### DIFF
--- a/tests/resources/metrics.yaml
+++ b/tests/resources/metrics.yaml
@@ -12,6 +12,7 @@ example:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
     notification_emails:
       - telemetry-client-dev@mozilla.com
+    expires: '2015-07-11'
 
   os:
     type: string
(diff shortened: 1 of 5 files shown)


badboy:
One thing I proposed in IRC:

* Have the main CI job use a pinned version of the glean parser
* Have an (allowed-to-fail?) CI job run against the latest version
* Always deploy with the pinned version.

mdboom:
Yeah, :+1: on pinning from me.  Then we can explicitly update and test things over here as part of a PR every time there is an update to `glean_parser` (which isn't very frequent).

badboy:
We should do that in a follow-up though and merge this for now.

Dexterp37:
Ok, filed #79

PR #80, opened 2019-03-26
Add &-browser and Fenix

fbertsch:


probe_scraper/scrapers/git_scraper.py
```diff
@@ -13,7 +13,17 @@
 
 MIN_DATES = {
     # Previous versions of the file were not schema-compatible
-    "glean": "2019-01-25 00:00:00"
+    "glean": "2019-03-15 00:00:00"
+}
+
+IGNORE_COMMITS = {
```
fbertsch: This was required because these commits all came on the same day, so simply limiting by date wasn't enough. That should be the only case when `IGNORE_COMMITS` should be used and not `MIN_DATES`.

probe_scraper/scrapers/git_scraper.py
```diff
@@ -13,7 +13,17 @@
 
 MIN_DATES = {
     # Previous versions of the file were not schema-compatible
-    "glean": "2019-01-25 00:00:00"
+    "glean": "2019-03-15 00:00:00"
```
fbertsch: This will fix the `expires` notifications

probe_scraper/scrapers/git_scraper.py
```diff
@@ -13,7 +13,17 @@
...
+}
+
+IGNORE_COMMITS = {
+    "fenix": [
+        "0e9400730b1168396ed4a19a5e3bae87300a1b75",
```
mdboom: These are just commits where the content in the fenix repo is somehow broken such that probe scraper will fail?

I wonder if there's a way we could add a test to fenix so that committing such things to the repo would be less likely.  (But that's just a wondering -- not a reason to hold up this PR).

fbertsch: Right, you can see there were 6 commits for that file that day, and it looks like the first 5 were not compatible: https://github.com/mozilla-mobile/fenix/commits/master/app/metrics.yaml

My gut says it'd be hard because of versioning, but worth considering.

PR #84, opened 2019-04-18
Use updated metrics.yaml only

fbertsch:


requirements.txt
```diff
@@ -1,6 +1,6 @@
 GitPython
 boto3
-glean_parser
+glean_parser>=0.23.1
```
mdboom: You probably want to make this `==0.23.1`.  There are some [very incompatible changes](https://github.com/mozilla/glean_parser/pull/58) coming along soon, and I'd hate to break `probe_scraper` before I have a chance to make a PR to fix it.

fbertsch:
This will fix all those error emails.

mdboom:
> This will fix all those error emails.

I don't think I get the error emails.  I should consider myself lucky?

requirements.txt
fbertsch: Cool, good idea.

Basically, we have those historical metrics files (from after 3-15 but before 4-12, when you removed `labeled` and added `labeled_*` metrics), which all failed with an error like below. Long-term we need to solve this with versioning:

Error in processing commit 5abfc5723d9730cb813137f8710d6dc857ee459a
Errors: [/mnt/analyses/probe-scraper/probe_cache/glean/5abfc5723d9730cb813137f8710d6dc857ee459a/components/service/glean/metrics.yaml:
    ```
    glean.error:
      invalid_label:
        bugs:
        - 1499761
        data_reviews:
        - https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5
        description: Counts the number of times a metric was set with an invalid
          label. The labels are the `category.name` identifier of the metric.
        expires: never
        labeled: true
        notification_emails:
        - telemetry-client-dev@mozilla.com
        type: counter
    ```
    
    Additional properties are not allowed ('labeled' was unexpected)

mdboom: Ah, thanks.

fbertsch:
@mdboom added your email, and updated the requirements.

PR #85, opened 2019-04-18
Update for glean_parser changes

mdboom:
These are just updates for changes to `glean_parser` in https://github.com/mozilla/glean_parser/pull/58.  That other PR is likely to go through some changes before landing, but I thought it wouldn't hurt to have this in the wings.

Basically it amounts to:
- `parse_metrics` is now `parse_objects` (because it parses metrics and ping objects)
- all `metrics.yaml` files *must* have a `$schema` key so we can tell them apart from `pings.yaml` files.


mdboom:
@Dexterp37 : I think this should fix the probe_scraper error e-mail you just forwarded me.

Issue #76, opened 2019-03-05
Glean probes should use the latest commit in addition to file changes


fbertsch:
Using the latest commit will allow us to get an accurate date range for a probe, since even if it changed 5 commits ago, it may still be available, but the `date` will be for the last commit the `metrics.yaml` file changed on.

This isn't urgent, since we can order glean probes by `first`.

fbertsch:
This change involves including the latest commit [in the dictionary of commits that we return.](https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/scrapers/git_scraper.py#L20).

fbertsch:
Looks like This was fixed.

PR #87, opened 2019-05-20
1542709: Add support for reading glean dependencies to the probe_scraper

mdboom:
This [PR to Fenix](https://github.com/mozilla-mobile/fenix/pull/1996) will need to be merged before the final version of this can land.  In the meantime, I've put a hardcoded version of the dependencies output in `dependencies.txt` here which at least puts it at a fetchable URL somewhere, but this should be removed before this is merged.

Despite that, I thought it would be worth filing this PR for an initial review in the meantime.  There is a corresponding PR to mozilla_schema_generator that I will post shortly which will actually *use* this data.


README.md
```diff
@@ -18,7 +18,45 @@ To scrape a git repository for probe definitions, an entry needs to be added in
 - `notification_emails`: Where emails about probe-scraper failures and improper files will be forwarded to. These
 will be just about your specific repository.
 - `url`: The URL of the repository to scrape. It should be able to be cloned directly from that URL.
-- `metrics_files`: A list of relative paths to `metrics.yaml` files
+- `metrics_files**: A list of relative paths to `metrics.yaml** files
+
+### Adding an application
```
Dexterp37: Should this be mentioned in the Glean SDK docs as well? Or at least reference this document. Or, better, should we have "one true doc" somewhere about "Using Glean" (SDK, Pipeline and whatnot). Not for this PR, but we need to track this somewhere I think.

repositories.yaml
```diff
@@ -14,6 +16,10 @@ fenix:
   url: 'https://github.com/mozilla-mobile/fenix'
   metrics_files:
     - 'app/metrics.yaml'
+  # TODO: This is a stub URL with example content.  This should be replaced with the
```
Dexterp37: Dropping this comment here so that it doesn't fall through the cracks :)

probe_scraper/runner.py
```diff
@@ -20,7 +20,7 @@
 from .parsers.metrics import GleanMetricsParser
 from .parsers.repositories import RepositoriesParser
 from .parsers.scalars import ScalarsParser
-from .scrapers import git_scraper, moz_central_scraper
+from .scrapers import dependency_scraper, git_scraper, moz_central_scraper
```
Dexterp37: Where is `dependency_scraper` implemented?

README.md
mdboom: We have [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1529761) about unifying Glean documentation.  I added a note that we have this documentation here as well that should be pulled in.

probe_scraper/runner.py
mdboom: Gah -- forgot to `git add` it.  Should be here now.

fbertsch:
@mdboom looks like the dependencies test failed!

probe_scraper/scrapers/git_scraper.py
```diff
@@ -29,8 +29,11 @@
 
 def get_commits(repo, filename):
     sep = ":"
-    commits = repo.git.log('--format="%H{}%ct"'.format(sep), filename)
-    with_ts = dict((c.strip('"').split(sep) for c in commits.split("\n")))
+    log_format = '--format="%H{}%ct"'.format(sep)
+    change_commits = repo.git.log(log_format, filename).split('\n')
+    most_recent_commit = repo.git.log('-n', '1', log_format).split('\n')
+    commits = set(change_commits) | set(most_recent_commit)
+    with_ts = dict((c.strip('"').split(sep) for c in commits))
```
mdboom: Fixes #76.

README.md
```diff
@@ -224,6 +262,43 @@ first and last commits that definition has been seen in, and when those commits
...
+in the `history` log is `version`, which is the version number of the
+dependency. At the top-level of each entry is:
+
+- `name`: the name of the dependency (for Android, this is a Maven
+  package name).
+- `type`: Always `"dependency"`.
+
+```
+{
+  "<library name>": {
```
fbertsch: Should this be `<application name>`?

probe_scraper/runner.py
```diff
@@ -220,11 +225,18 @@ def load_glean_metrics(cache_dir, out_dir, repositories_file, dry_run):
                         "subject": "Probe Scraper: Error on Metric Parsing",
                         "message": msg
                     })
+            dependencies[repo_name][commit_hash] = dependency_scraper.parse_dependencies(
```
fbertsch: I _think_ the correct thing to do here would be to only check commit hashes on commits where the dependencies file changed. In this case since the dependency file isn't actually in git, we would need the original dependencies file (`build.gradle`?).

Since at this point we're not expecting: many libraries, or many additions/removals of libraries, this may be an optional change, but it's worth noting that we could theoretically miss dependencies with this approach.

README.md
mdboom: Each entry here is a library that the application depends on.  I could probably make that clearer in the docs.

probe_scraper/runner.py
mdboom: Ah, that's a good point.  Changes can actually occur across multiple files -- there are multiple `build.gradle` files, and technically, everything in `buildSrc` could affect the dependencies.

I guess we should add a `dependencies_files` key to `repositories.yaml` and treat it similarly to `metrics_files` with respect to how the git commit hashes are determined?

fbertsch: I think that's the way. Things do seem to be getting a little bloated, but I'm not sure of a good alternative.

probe_scraper/scrapers/dependency_scraper.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+import functools
+import re
+
+import requests
+from requests_file import FileAdapter
+
+session = requests.Session()
+session.mount('file://', FileAdapter())
```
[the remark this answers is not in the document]
mdboom: Yes -- I'll add a comment.  It basically adds support for `file://` URLs to the requests library (which by default only supports `http(s)`.

schemas/repositories.json
```diff
@@ -22,6 +22,22 @@
           "type": "string",
           "pattern": "metrics\\.yaml$"
         }
+      },
+      "dependencies_url": {
+        "type": "string",
+        "format": "uri"
+      },
+      "dependencies_format": {
```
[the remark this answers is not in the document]
mdboom: Definitely for iOS and desktop the formats will be significantly different.  Since that's already on the roadmap, I thought we should plan for it so we don't end up in a situation where gradle is the default or something confusing...

probe_scraper/parsers/repositories.py
```diff
@@ -22,10 +22,17 @@ def __init__(self, name, definition):
...
         self.metrics_file_paths = definition.get("metrics_files", [])
+        self.library_names = definition.get("library_names", None)
+        self.dependencies_url = definition.get("dependencies_url", None)
+        self.dependencies_format = definition.get("dependencies_format", None)
+        self.dependencies_files = definition.get("dependencies_files", [])
 
     def get_metrics_file_paths(self):
         return self.metrics_file_paths
 
+    def get_change_files(self):
```
fbertsch: There is some unnecessary work now happening, where metrics and dependencies are being read in commits where they didn't change. This isn't a big deal but could be a future point of optimization.

mdboom: That's a good point.  I'm not seeing a *simple* way to address that now, since it tracks dependencies and probes together.  Are you comfortable with coming back to this later if it becomes a performance problem?  (I don't think any of these files change terribly often, relative to other files in the repo...)

PR #88, opened 2019-06-10
1557843: Add lib-crash to probe_scraper

mdboom:
This will be required to get crash counts once https://github.com/mozilla-mobile/android-components/pull/3099 lands.

Cc: @travis79

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -46,3 +46,15 @@ firefox-for-fire-tv:
   notification_emails:
     - frank@mozilla.com
   url: 'https://github.com/mozilla-mobile/firefox-tv'
+lib-crash:
+  app_id: lib-crash
+  notification_emails:
+    - frank@mozilla.com
+    - mdroettboom@mozilla.com
+    - android-components-team@mozilla.com
+  url: 'https://github.com/mozilla-mobile/android-components'
+  metrics_files:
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -28,3 +28,14 @@ reference-browser:
     - frank@mozilla.com
     - mdroettboom@mozilla.com
   url: 'https://github.com/mozilla-mobile/reference-browser'
+lib-crash:
+  app_id: lib-crash
+  notification_emails:
```
fbertsch: Are these the correct email addresses?

mdboom: That's a good policy question.  I guess maybe someone from telemetry/Glean and someone from the library would be a good way to address this.  On this, the natural choices are probably @travis79 (who did the Glean integration) and @pocmo (the owner of lib-crash).  

(And by that argument, it would be great to have someone from the Fenix team in the Fenix section, but that's for a separate bug...)

Dexterp37: I feel like this should probably match what's in the [metrics.yaml](https://github.com/mozilla-mobile/android-components/pull/3099/files#diff-4accd56399d25ff7394107fe66647031) for mozilla-mobile/android-components#3099 . I don't think there should be any of us in there. Is there any reason why you think that might be useful?

PR #91, opened 2019-07-02
Add fenix-nightly and fire-tv

fbertsch:


fbertsch:
Apparently fenix-nightly will begin reporting as soon as tomorrow!

PR #95, opened 2019-07-25
1568485: Add engine-gecko dependencies to repositories.yaml

mdboom:
Even though there is a scraper for mozilla-central, the repositories.yaml
feature for Glean is currently limited to git repositories.  Therefore, this
uses the read-only mirror of the gecko-dev part of mozilla-central on Github.

There was a small code change required to support only looking on a specific
branch of the repo.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -72,6 +72,8 @@ engine-gecko:
   notification_emails:
     - frank@mozilla.com
     - mdroettboom@mozilla.com
+    - android-components-team@mozilla.com
+    - geckoview-team@mozilla.com
   url: 'https://github.com/mozilla/gecko-dev'
   branch: release
   metrics_files:
@@ -83,6 +85,8 @@ engine-gecko-beta:
   notification_emails:
     - frank@mozilla.com
     - mdroettboom@mozilla.com
+    - android-components-team@mozilla.com
+    - geckoview-team@mozilla.com
   url: 'https://github.com/mozilla/gecko-dev'
   branch: beta
   metrics_files:
@@ -94,6 +98,8 @@ engine-gecko-nightly:
   notification_emails:
     - frank@mozilla.com
     - mdroettboom@mozilla.com
+    - android-components-team@mozilla.com
+    - geckoview-team@mozilla.com
   url: 'https://github.com/mozilla/gecko-dev'
   metrics_files:
     - 'toolkit/components/telemetry/geckoview/streaming/metrics.yaml'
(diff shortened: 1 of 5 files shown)


repositories.yaml
```diff
@@ -57,4 +57,35 @@ lib-crash:
     - 'components/lib/crash/metrics.yaml'
   library_names:
     - org.mozilla.components:lib-crash
-  
+engine-gecko:
+  app_id: engine-gecko
+  notification_emails:
+    - frank@mozilla.com
```
Dexterp37: What are the notification emails used for in the probe-scraper? For general contacts? If so, I think that `android-components-team@mozilla.com` and `geckoview-team@mozilla.com` should be there, in addition to you two ;)

repositories.yaml
```diff
@@ -57,4 +57,35 @@ lib-crash:
...
+  url: 'https://github.com/mozilla/gecko-dev'
+  branch: release
+  metrics_files:
+    - 'toolkit/components/telemetry/geckoview/streaming/metrics.yaml'
+  library_names:
+    - org.mozilla.components:browser-engine-gecko
+engine-gecko-beta:
+  app_id: engine-gecko-beta
+  notification_emails:
+    - frank@mozilla.com
```
Dexterp37: As above, with the emails

repositories.yaml
```diff
@@ -57,4 +57,35 @@ lib-crash:
...
+  url: 'https://github.com/mozilla/gecko-dev'
+  branch: beta
+  metrics_files:
+    - 'toolkit/components/telemetry/geckoview/streaming/metrics.yaml'
+  library_names:
+    - org.mozilla.components:browser-engine-gecko-beta
+engine-gecko-nightly:
+  app_id: engine-gecko-nightly
+  notification_emails:
+    - frank@mozilla.com
```
Dexterp37: As above, with the emails

mdboom:
> Did you try this locally? How did it perform?

The bottleneck seems to be the initial clone.  Once checked out, it's really fast because it only looks at commits where `metrics.yaml` has changed (and that's currently only 1).

@fbertsch: There might be legitimate concerns there.  The `mozilla/gecko-dev` mirror is 3GB, but (AFAICT) that only has to be done once (or three times here, on for each branch).  One way around this might be to do a shallow clone from some known commit (when the `metrics.yaml` was first added).

mdboom:
Before I merge -- @fbertsch : Can you confirm that the extra time and disk space requirements would be an issue?  (This downloads a 3GB repo three times...)

fbertsch:
I'm not _overly_ concerned about timing out. I would say let's see how this runs and if it does time out, or takes too long, we can reconsider.

PR #101, opened 2019-07-31
WIP: 1569184: Automatically extract dependencies from gradle apps

mdboom:
This is a first pass at automatically extracting the dependencies for gradle-based Android products within probe-scraper itself, rather than depending on the app uploading an artifact.

The biggest structural change is that the handling of dependencies happens as a subtask of the `git_scraper`, since it needs a full git clone to do its work, and we wouldn't want to clone these twice.

The biggest problem with this approach is how slow it is.  It takes a really long time to run gradle on what is currently for Fenix 204 commits.  It seems like caching might be the answer (since the dependencies should be constant for a given commit), but I'm not sure how best to handle persistent caching in the EMR environment (or whatever problems that would create).

This also needs updated installation  docs for local testing -- for now see https://github.com/mozilla/telemetry-airflow/pull/571 -- and needs updated tests.  But I'd put this up for discussion first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ will be just about your specific repository.
 
 ### Adding an application
 
-All **applications** in `repositories.yaml` must also define `dependencies_url`,
+All **applications** in `repositories.yaml` must also define
 `dependencies_format` and `dependencies_files`.
 
 Glean metrics are emitted by the application using Glean, any libraries it uses
@@ -33,21 +33,11 @@ that application.
 Currently, probe-scraper has support for reading dependencies from the following
 platforms and build systems:
 
-- **gradle for Android:** Obtain the dependencies for your application using
-  `./gradlew app:dependencies --configuration implementation`
+- `gradle`: `probe_scraper` will obtain the dependencies for your
+  application using `./gradlew app:dependencies --configuration implementation`
 
-Configure the application's CI system to store the output of one of the above
-commands at a publicly accessible URL that contains the git commit hash of the
-application that generated it.
-
-Set the `dependencies_url` parameter for the application in `repositories.yaml`
-to this URL, using the `{commit_hash}` marker to indicate the part that should be replaced
-with a git commit hash. Also set the `dependencies_format` parameter to the name
+Set the `dependencies_format` parameter to the name
 of the build system in use (currently only `gradle` is supported).
-
-For example, [here were the
-changes](https://github.com/mozilla-mobile/fenix/pull/1996) to make this work
-for Fenix, which uses Taskcluster for CI.
 
 Set the `dependencies_files` parameter to a list of files that change when
 dependencies of of the application change. For example, in an Android project,
--- a/probe_scraper/scrapers/git_scraper.py
+++ b/probe_scraper/scrapers/git_scraper.py
@@ -10,6 +10,8 @@ import tempfile
 import traceback
 from datetime import datetime, timedelta
 
+from . import dependency_scraper
+
 
 MIN_DATES = {
     # Previous versions of the file were not schema-compatible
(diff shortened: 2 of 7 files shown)


probe_scraper/scrapers/git_scraper.py
```diff
@@ -71,13 +74,27 @@ def retrieve_files(repo_info, cache_dir):
...
+
+        all_hashes = {}
+        for rel_path in repo_info.dependencies_files:
+            all_hashes.update(get_commits(repo, rel_path))
+        for _hash, ts in all_hashes.items():
+            if (min_date and ts < min_date):
+                continue
+
+            if repo_info.dependencies_format is not None:
+                repo.git.checkout(_hash)
```
Dexterp37: Do we really need this for all the hashes (e.g. can we shallow clone the 1 latest master commit?)? What happens if the application uses a component that publishes Glean metrics and then removes it 2 versions later?

README.md
```diff
@@ -33,22 +33,12 @@ that application.
 Currently, probe-scraper has support for reading dependencies from the following
 platforms and build systems:
 
-- **gradle for Android:** Obtain the dependencies for your application using
-  `./gradlew app:dependencies --configuration implementation`
+- `gradle`: `probe_scraper` will obtain the dependencies for your
```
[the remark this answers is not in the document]
mdboom: That's a good point, but I don't actually know the answer.  @fbertsch ?

probe_scraper/scrapers/git_scraper.py
mdboom: It runs gradle on every commit where the dependencies *may* have changed.  (Based on when the files specified in `dependencies_files` have changed, which includes certain key build.gradle files as well as Dependencies.kt.  This has already happened 204 times since Glean was integrated into Fenix.

This is done exactly for the case you mention -- if a dependency goes in and later comes out -- we still need to catch it and document it because it may have sent data at some point in the past.

It's possible we could get away with fewer checks on beta and release and just track release tags, or something.  But for nightly we're probably still stuck doing something like this.

Dexterp37: I see, thanks for clarifying. Looks like we're stuck with this, then :(

README.md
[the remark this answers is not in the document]
fbertsch: This does run once a day

probe_scraper/scrapers/git_scraper.py
fbertsch: We could cache these in s3, like we're doing with the nightly `histogram.json` files.

mdboom: @fbertsch: Can you point me at where that happens?  Is that the requests_cache [here](https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/scrapers/moz_central_scraper.py#L219)?

fbertsch:
Okay, I wrote up a short proposal on moving to GCP, where we can use GKE to get dependencies within the probe-scraper: https://github.com/mozilla/probe-scraper/issues/104. It should be useful for any kind of analysis we want to do on applications.

Dexterp37:
> 
> 
> Okay, I wrote up a short proposal on moving to GCP, where we can use GKE to get dependencies within the probe-scraper: #104. It should be useful for any kind of analysis we want to do on applications.

@fbertsch the proposal makes much sense! Is this PR blocked by that proposal?

mdboom:
> @fbertsch the proposal makes much sense! Is this PR blocked by that proposal?

I think regardless, let's hold off on this PR until we have a chance to discuss.  In my own proposal around different ways to handle dependencies, it came to light that there may be better options than this.

mdboom:
Closing.  We've decided to no longer do automatic dependency determination.

PR #103, opened 2019-08-05
Add sync Glean repo

fbertsch:


fbertsch:
Got the dependency name from here: https://taskcluster-artifacts.net/afeCANcvTlaihUgkd4QFFw/0/public/dependencies.txt

PR #113, opened 2019-08-22
Fix #93: Improve repositories schema

### LINKED ISSUE #93: Add a schema for repositories.yaml
It's easy to make small, not noticeable mistakes. A schema that validates would not allow this.

mdboom:
This tightens things up by adding `"additionalProperties": false`, which lead to noticing that `dependencies_files` was never declared.

Also adds the ability to specify `pings.yaml` files, which would be required by libraries that specify custom pings.

### DIFF
--- a/schemas/repositories.json
+++ b/schemas/repositories.json
@@ -23,10 +23,16 @@
         "type": "array",
         "items": {
           "type": "string",
-          "pattern": "metrics\\.yaml$"
+          "pattern": "(metrics|pings)\\.yaml$"
         }
       },
       "dependencies": {
+        "type": "array",
+        "item": {
+          "type": "string"
+        }
+      },
+      "dependencies_files": {
         "type": "array",
         "item": {
           "type": "string"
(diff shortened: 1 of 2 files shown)


mdboom:
Thinking about your comment, @fbertsch: You're right we definitely don't need to accept `pings.yaml` on `metrics_files`.  To the extent that new pings are added, those would usually have new metrics associated with then anyway...

Whether we even need to track `pings.yaml` is sort of an open question, and I'd argue we don't bother until we have a use case.

mdboom:
> Should we have naming conventions around the top-level names? Note that they are used in the URLs, e.g. https://probeinfo.telemetry.mozilla.org/glean/glean/metrics

Is this a place where we want to use kebab-case to be consistent with the `docType`s?

fbertsch:
@mdboom I think so. Want to add that here?

fbertsch:
Worst case is something comes along that _needs_ a different name and we can relax it. This will prevent any egregious errors until then.

mdboom:
Updated this to enforce `kebab-case`.

PR #114, opened 2019-08-28
1577253: Remove automatic dependency determination

mdboom:
This hardcodes the dependencies and removes all of the code for automatic dependency handling.


mdboom:
Yep -- output is identical before and after this change (except after is missing the `history` entry).

PR #116, opened 2019-08-30
1523977: Detect and alert about duplicate Glean metrics

mdboom:
This performs a check for duplicate metrics for Glean-using applications, sends an e-mail if it finds any, and removes all metrics for an app where duplicate metrics were found.

@fbertsch: In particular, please confirm that the way the error is handled seems right.  If it finds an app with duplicate metrics, it removes all metrics for that app.  That's the same behavior as if the `metrics.yaml` is invalid against the schema.  But maybe that's too heavy-handed and I don't know the implications of saying an app has no metrics on the operations downstream of this.


mdboom:
I just realized this should e-mail the addresses associated with the metrics involved, not just the addresses associated with the repositories.  Let me see if I can get that working...

tests/test_git_scraper.py
```diff
@@ -57,6 +60,12 @@ def get_repo(repo_name):
     directory = os.path.join(test_dir, repo_name)
     repo = Repo.init(directory)
 
+    # We need to synthesize the time stamps of commits to each be a second
+    # apart, otherwise the commits may be at exactly the same second, which
+    # means they won't always sort in order, and this the merging of identical
+    # metrics in adjacent commits may not happen correctly.
+    base_time = time.time()
```
mdboom: @fbertsch: This was a fun one!  I think it's only a bug in the way things are tested, though, not in probe scraper itself.

fbertsch: Yes, I agree! Good catch on that one - much more performant than a `sleep` command at the end of each iteration :)

tests/test_git_scraper.py
```diff
@@ -132,14 +156,17 @@ def test_normal_repo(normal_repo):
     duration = 'example.duration'
     os_metric = 'example.os'
 
-    # duration has 1 definition
-    assert len(metrics[duration][HISTORY_KEY]) == 1
+    # duration has 2 definition
```
fbertsch: The most nit-picky nit in the history of nits: `definitions`

probe_scraper/glean_checks.py
```diff
@@ -0,0 +1,81 @@
...
+        for dependency in dependencies:
+            for metric in metrics_by_repo[dependency].keys():
+                metric_sources.setdefault(metric, []).append(dependency)
+
+        duplicate_sources = dict(
+            (k, v) for (k, v) in metric_sources.items() if len(v) > 1
+        )
+
+        if len(duplicate_sources):
+            for name, sources in duplicate_sources.items():
```
fbertsch: We _could_ send a single email with all dupes, but that wouldn't really apply to probe-owners (who aren't repo owners).

mdboom:
> Q: What will happen on the client if there are duplicate metrics in a single ping?

It's pretty bad.  Essentially, the metrics will be stored in the same location in the client database, so they will "fight" over the same storage.  Whichever metric records last will be what's sent in the ping, or in the case of something like a counter the counts will be summed together.  In short, it's really bad.  We'd love to find a way to detect this at build time before these mistakes even get merged, but haven't found a way as of yet, so this PR represents doing at least *something*, even though it's later than we'd like.  But hopefully it detects the problem on "nightly/master" prior to a proper release.

> Q: Will it write nothing to the probe-info-service (making that repo a 404), or will it write an empty JSON file?

It writes an empty JSON file -- here, I was just following the example of what happens when the `metrics.yaml` is invalid.  I'm not sure that's the correct behavior.  Like you say, if that results in deleting a bunch of columns that sounds **very bad**.  If the probe_scraper raised an exception (or otherwise quit with a UNIX error) is that better?  What happens in that case? -- we'd continue to get the yesterday's output flowing downstream?

> If we run into a situation with a duplicate metric we'll need to hash out how to fix it. My gut says we'll have to special-case it in the probe-scraper since it sees all of git history, so we can never remove it from the client repository and fix things.

Oof.  That's indeed tricky.  I'd like to find a way to solve it in the repos themselves if we can.  Off the top of my head, I was thinking:

1. We only send emails/error out when the same metric exists in the latest revision of two repositories.
2. Otherwise, the metric in the later revision "wins out".

This isn't great because apps are likely to depend on historical revisions of libraries for some time, and we'd like to detect when things are "correct" again.

Alternatively, we could put the special casing in the repos (the edges), with some sort of flag "retroactively_removed".  If we see this on a metric for a given repo, we (effectively) delete it in probe_scraper in that revision and all revisions prior.  We then do the duplicate detection / generate output from that.  It feels like a giant hack, but somehow feels better than a bunch of allow/block lists that live in probe_scraper.

Lot's of imperfect solutions to choose from, I guess. Happy to discuss.

Dexterp37:
> > If we run into a situation with a duplicate metric we'll need to hash out how to fix it. My gut says we'll have to special-case it in the probe-scraper since it sees all of git history, so we can never remove it from the client repository and fix things.
> 
> Oof. That's indeed tricky. I'd like to find a way to solve it in the repos themselves if we can. Off the top of my head, I was thinking:
> 
>     1. We only send emails/error out when the same metric exists in the latest revision of two repositories.

I'd go for the first one. I think we should go for at least a warning here, while still trying to fail at build time.

probe_scraper/glean_checks.py
```diff
@@ -0,0 +1,81 @@
...
+                    addresses.update(repo_by_name[source].notification_emails)
+
+                    # Also send to the metric's contacts
+                    for history_entry in metrics_by_repo[source][name]["history"]:
+                        addresses.update(history_entry["notification_emails"])
+
+                emails[name] = {
+                    "emails": [
+                        {
+                            "subject": "Probe scraper: Duplicate metric identifiers",
```
Dexterp37: nit: I'd lean a bit more into leveraging our "brand". `Glean: duplicate metric identifiers detected`

probe_scraper/glean_checks.py
Dexterp37: I'd really expect dupes to be a low frequency exception, not the rule: maybe receiving a few emails about it isn't that bad

probe_scraper/glean_checks.py
```diff
@@ -0,0 +1,81 @@
...
+            for metric in metrics_by_repo[dependency].keys():
+                metric_sources.setdefault(metric, []).append(dependency)
+
+        duplicate_sources = dict(
+            (k, v) for (k, v) in metric_sources.items() if len(v) > 1
+        )
+
+        if len(duplicate_sources):
+            for name, sources in duplicate_sources.items():
+                msg = "Duplicate metric: {!r}: exists in {}".format(
```
Dexterp37: What about crafting this a bit differently, following the great example in mozilla/medusa#31 from @chutten ? Something along the lines of:

```
Glean has detected duplicated metric identifiers coming from your product, {repo_name}.

- {metric_identifier} defined in {repo_a} and {repo_b};
- {metric_identifier} defined in {repo_a} and {repo_b};
- ...

What to do about this:

1. File a bug to track your investigation. You can just copy this email into the bug Description to get you started.
2. Reply-All to this email to let the list know that you are investigating. Include the bug number so we can help out.
3. Rename the most recently added {metric_identifier} to be more specific. See [1]

If you have any problems, please ask for help on the #glean Slack channel. We'll give you a hand.

What this is:

We have a system called probe-scraper [2] that scrapes the metric information from all Mozilla products using the Glean SDK. All the scraped data is available on the probeinfo service [3]. The scraped definition is used to build things such as the probe-dictionary [4] and other data tools. It detected that one metric that was recently added has an identifier collision with some metric that already existed in the application namespace. So it sent this email out, encouraging you to fix the problem.

What happens if you don't fix this:

TODO?

You can do this!

Your Friendly, Neighbourhood Glean Team

[1] - https://mozilla.github.io/glean/book/user/adding-new-metrics.html#naming-things
[2] - https://github.com/mozilla/probe-scraper
[3] - https://probeinfo.telemetry.mozilla.org/
[4] - https://telemetry.mozilla.org/probe-dictionary/
```

probe_scraper/glean_checks.py
mdboom: I feel like the best compromise here might be to: Send a single e-mail with all dupes, that goes to all repo owners and all metric owners involved.  It does mean that the e-mail will contain metrics that some of the metric owners don't care about, but given that there will be other metrics in there that they *do* care about, it seems like a small price to pay.  And I think I like that better than the current behavior which has the potential (in rare cases) to send many e-mails to the same individual in a single run.

probe_scraper/transform_probes.py
```diff
@@ -353,8 +353,20 @@ def transform_by_hash(commit_timestamps, metric_data):
         sorted_commits = sorted(iter(commits.items()),
                                 key=lambda x_y: commit_timestamps[repo_name][x_y[0]])
 
+        # Not only do we need to know when metrics were added or changed, we also
```
fbertsch: Can we add a test for this change? It's not clear to me what the expected outcome here is:
- Should there be an entry in `history` with just `{"disabled": True}`?
- Should this probe stop being reported in the probe-scraper altogether?

mdboom: Yes -- if a probe is removed, it should have an entry with the commit where it was removed, and the only contents are `{"disabled": True}`.

However, I don't think it should stop being reported in the probe-scraper altogether, since we want to retain the database columns for historical data before the metric was expired (I think) -- though that kind of relates to questions of data retention and when we would go back and cull columns for expired metrics.  Regardless, if we don't make this change, the probe-scraper can't communicate downstream to any tools that a metric was removed (and thus disabled).

mdboom: I've added a test here.

fbertsch: Great, thanks, this helps me understand what's happening. So the idea is that we'll remove a duplicate probe, and then it will show up as `"disabled": True` in the probe-info-service?

There's an additional hiccup, in that MSG doesn't know anything about this non-definitioned probe (a disabled one). It will try to get definition for it to fill-in the schema, and fail.

Question then: How do we decipher between probes that are _just_ disabled, and probes that were duplicates and disabled? The former should show up in the schema, but the latter should not.

If we do something fancy like "if a dupe is found in MSG, take the one that is not disabled", how do we handle the future case where dupe happens, one metric is disabled, and then sometime down the line, the other one is removed/disabled?

probe_scraper/glean_checks.py
```diff
@@ -0,0 +1,131 @@
...
+
+We have a system called probe-scraper [2] that scrapes the metric information from all Mozilla products using the Glean SDK. All the scraped data is available on the probeinfo service [3]. The scraped definition is used to build things such as the probe-dictionary [4] and other data tools. It detected that one metric that was recently added has an identifier collision with some metric that already existed in the application namespace. So it sent this email out, encouraging you to fix the problem.
+
+What happens if you don't fix this:
+
+The metrics will compete to send their data in pings, making the data unreliable at best.
+
+You can do this!
+
+Your Friendly, Neighbourhood Glean Team
```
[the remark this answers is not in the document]
chutten: ... :canada:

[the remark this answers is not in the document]
Dexterp37: For the record, that's also British English, which is what I usually tend to write :P

probe_scraper/transform_probes.py
mdboom: Given these issues, and what we discussed in the meeting today, I think (a) we want to have *something* to start detecting this problem sooner than later, but (b) finding a good way forward to resolve these issues once they are created is going to be difficult.

Based on that, I propose:

1) I back out the changes here that actually modify the output (the insertion of `"disabled": True` entries in the history list).  That solution is perhaps half-baked.
2) Word the e-mail carefully such that it's clear all developers should reach out for help.
3) Deal with the problem of how to actually resolve any conflicts when we know we actually have them (since the solution may be somewhat case-by-case anyway).
4) Continue to tackle this problem on better fronts -- either by "fully namespacing" or build-time detection.

What do you think?

fbertsch: I think what you've outlined is wise. If we're lucky we won't have to deal with this problem :)

mdboom:
As discussed, this has been simplified to just give us emails about any duplicates.  @fbertsch : Significant enough changes that maybe it should have one final look before merging?

fbertsch:
@mdboom apologies, this requires a rebase now :/

mdboom:
@fbertsch : Rebased.  Hopefully this is down to the minimal "do no harm" version now...

PR #119, opened 2019-09-20
Include artificial tags in probe-scraper json-tags

fbertsch:
More info here: https://github.com/mozilla/mozilla-schema-generator/issues/70#issuecomment-533168438

We solve this by simply specifying commits we guarantee we want to scrape.

### DIFF
--- a/probe_scraper/scrapers/moz_central_scraper.py
+++ b/probe_scraper/scrapers/moz_central_scraper.py
@@ -31,6 +31,13 @@ CHANNELS = {
     'nightly': {
         'base_uri': 'https://hg.mozilla.org/mozilla-central/',
         'tag_regex': '^FIREFOX_(AURORA|BETA)_[0-9]+_BASE$',
+        'artificial_tags': [
+            {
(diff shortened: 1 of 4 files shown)


probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -44,9 +52,24 @@
...
 ERROR_CACHE_FILENAME = 'probe_scraper_errors_cache.json'
+ARTIFICIAL_TAG = 'artificial'
 
 
 def load_tags(channel):
+    """
+    A bit of info about `json-tags`
+
+    They are a set of tags that represent the first and
+    last commit hat a release_channel went out on. For example,
```
chutten:
```suggestion
    last commit that a release_channel went out on. For example,
```

But now I want a commit hat.

fbertsch:
@chutten RFAL

probe_scraper/scrapers/moz_central_scraper.py
chutten: Still says "hat"

PR #121, opened 2019-09-25
1578719: Do not disable expired metrics

mdboom:
Requires https://github.com/mozilla/glean_parser/pull/108/files to be merged first.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 GitPython
 boto3
-glean_parser==1.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 GitPython
 boto3
-glean_parser==1.1
+glean_parser==1.8
```
fbertsch: Any other changes worth worrying about with this version bump?

PR #122, opened 2019-09-27
1584182: Move browser-storage-sync to support-sync-telemetry

mdboom:
This is in support of the changes in https://github.com/mozilla-mobile/android-components/pull/4480.

tl;dr: The metrics that were formerly sent by `browser-storage-sync` have been moved to a helper component in `support-sync-telemetry` (which will in the future be used by other sync products).

None of the metrics were actually renamed, so I think this is a safe, non-schema-changing transformation, but could you please confirm, @fbertsch ?  This is the sort of change we haven't really anticipated...

Cc: @grigoryk

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -19,7 +19,7 @@ fenix:
   dependencies:
     - org.mozilla.components:service-glean
     - org.mozilla.components:lib-crash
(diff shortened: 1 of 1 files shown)


fbertsch:
Ah, this was causing the failures over the weekend!

PR #123, opened 2019-10-01
Pull in labels for categorical histograms

### LINKED ISSUE #111: Include `labels` for categorical histograms
MDV2 needs this to display those histograms.

robhudson:
Fixes #111

### DIFF
--- a/probe_scraper/parsers/histograms.py
+++ b/probe_scraper/parsers/histograms.py
@@ -46,6 +46,10 @@ def extract_histogram_data(histogram, version):
             value = defaults[source_field]
         set_in_nested_dict(data, target_field, value)
 
+    # Only include labels if the histogram is categorical.
+    if histogram.kind() == "categorical":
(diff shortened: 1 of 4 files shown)


probe_scraper/parsers/histograms.py
```diff
@@ -46,6 +46,10 @@ def extract_histogram_data(histogram, version):
             value = defaults[source_field]
         set_in_nested_dict(data, target_field, value)
 
+    # Only include labels if the histogram is categorical.
+    if histogram.kind() == "categorical":
+        set_in_nested_dict(data, "details/labels", histogram.labels())
```
fbertsch: What will this value be for categorical histograms that don't have labels defined?

robhudson: There aren't any in the current histograms.json file, but I could guard against that and add an empty list in that case.

PR #128, opened 2019-10-09
1586789: Send e-mails for expired Glean metrics

mdboom:


probe_scraper/glean_checks.py
```diff
@@ -126,3 +127,81 @@ def check_for_duplicate_metrics(repositories, metrics_by_repo, emails):
...
+What to do about this:
+
+1. If the metric is no longer needed, remove it from the `metrics.yaml` file.
+2. If the metric is still required, resubmit a data review [1] and extend its expiration date.
+
+If you have any problems, please ask for help on the #glean Slack channel. We'll give you a hand.
+
+What happens if you don't fix this:
+
+The metrics will automatically stop collecting data after their expiration date, and you will continue to get this e-mail as a daily reminder.
```
badboy: Due to how we do date-expiry, is it worth mentioning that this is affects _builds_ after the expiration date?

probe_scraper/glean_checks.py
```diff
@@ -126,3 +127,81 @@ def check_for_duplicate_metrics(repositories, metrics_by_repo, emails):
...
+
+Your Friendly, Neighborhood Glean Team
+
+[1] https://wiki.mozilla.org/Firefox/Data_Collection
```
badboy: Cerberus explicitly adds this line:

> This is an automated message sent by Cerberus. See https://github.com/mozilla/cerberus for details and source code

Should we have something similar? *bleep* *blop* I'm a bot! 🤖

probe_scraper/glean_checks.py
mdboom: Very good point.

probe_scraper/glean_checks.py
```diff
@@ -126,3 +127,83 @@ def check_for_duplicate_metrics(repositories, metrics_by_repo, emails):
...
+
+
+EXPIRED_METRICS_EMAIL_TEMPLATE = """
+The following metrics in {repo_name} will expire in the next {expire_days} days or have already expired.
+
+{expired_metrics}
+
+What to do about this:
+
+1. If the metric is no longer needed, remove it from the `metrics.yaml` file.
```
Dexterp37: Do we have *any* way to provide a link to the specific `metrics.yaml`?

mdboom: We could figure it out from what we have, if it's on Github (which everything is right now, even for m-c since we go through a github mirror).

mdboom:
> I do think we should only run this once a week, so you would get two advance warning emails about the expiration.

I don't know exactly how `probe_scraper` gets run, but would it be enough to add a check for "is it Monday?"

badboy:
Wouldn't it makes sense to report this if expire is 14 days in the future, then 7 days in the future (under the assumption that probe_scraper runs daily)?
Metrics can expire on any date, so only running that expiration check once a week might miss some.

mdboom:
@badboy: Each e-mail wraps up all of the expired metrics for a particular time window, so it should miss any metrics expiring any time during that window.  I did it this way to prevent being spammed when a ton of metrics all expire on the same day (as would be the case with Fenix, for example).  I think it would be less confusing to get a single email on a fixed day of the week containing all of the metrics set to expire in the next 14 days, rather than separate e-mails for metrics that expire on a Tuesday vs those on a Monday.

PR #129, opened 2019-10-16
Fix unit test for expiry emails on Monday

mdboom:


fbertsch:
Ah, flake8 error.

tests/test_git_scraper.py
```diff
@@ -280,7 +281,17 @@ def test_check_for_expired_metrics(expired_repo):
...
 
-    runner.main(cache_dir, out_dir, None, None, False, True, repositories_file, True, None, None)
+    # Mock `datetime.date.today` so it's a Monday, the only day that
+    # expirations are checked.
+    class MockDate(datetime.date):
+        @classmethod
+        def today(cls):
+            return datetime.date(2019, 10, 14)
+
+    with unittest.mock.patch("probe_scraper.glean_checks.datetime.date", new=MockDate):
```
fbertsch: import `unittest.mock`, but use `unittest.mock.patch`? How is that working?

mdboom: This is IIUC a standard way to import a python subpackage.  Could have alternatively done `from unittest import mock` -- happy to change if that's clearer.

fbertsch: Oh right, that was my confusion. Carry on!

PR #134, opened 2019-10-21
Modifying Probe scraper runner to do the s3 syncs

haroldwoo:
Since we will run in a container via GKEPodOperator instead of EMR -> scriptrunner jar -> bash -> python -> bash

Replaces logic in:
https://github.com/mozilla/telemetry-airflow/blob/master/jobs/probe_scraper.sh

TODO:
- [x]  Create aws creds with aws prod access to both the s3 cloudfront prod bucket and the telemetry-airflow-cache bucket. 
- [x] Copy cache data over from aws dev -> prod bucket
- [x]  Add creds to wtmo, and secret sops repo
- [x]  Test
- [x]  This should merge along with a PR to telemetry-airflow that modifies the probe-scraper.py dag to use GKEPodOperator. Also make sure to sync the cache from aws dev to prod before the first migrated run

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM python:3
 MAINTAINER Harold Woo <hwoo@mozilla.com>
 
+ENV PYTHONUNBUFFERED=1
+
 ARG APP_NAME=probe-scraper
 ENV APP_NAME=${APP_NAME}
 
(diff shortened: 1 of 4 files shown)


haroldwoo:
Planning to do some testing this week.

PR #142, opened 2019-12-09
IOS and Desktop

fbertsch:
Adds both of those apps, also changes the behavior when running on a single app (this should not affect prod).

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -136,6 +136,14 @@ firefox-reality:
   dependencies:
     - org.mozilla.components:service-glean
     - org.mozilla.components:support-sync-telemetry
+lockwise-ios:
+  app_id: org.mozilla.ios.Lockbox
+  notification_emails:
+    - frank@mozilla.com
+    - tlong@mozilla.com
+  url: 'https://github.com/mozilla-lockwise/lockwise-ios'
+  dependencies:
+    - org.mozilla.components:service-glean
 logins-store:
   app_id: logins-store
   notification_emails:
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -136,6 +136,14 @@ firefox-reality:
   dependencies:
     - org.mozilla.components:service-glean
     - org.mozilla.components:support-sync-telemetry
+fogotype:
+  app_id: org.mozilla.fogotype
+  notification_emails:
+    - frank@mozilla.com
+    - chutten@mozilla.com
+  url: 'https://github.com/mozilla/gecko-dev'
+  dependencies:
+    - org.mozilla.components:service-glean
 lockwise-ios:
   app_id: org.mozilla.ios.Lockbox
   notification_emails:
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -144,6 +144,13 @@ fogotype:
   url: 'https://github.com/mozilla/gecko-dev'
   dependencies:
     - org.mozilla.components:service-glean
+lockwise-android:
+  app_id: mozilla.Lockbox
+  notification_emails:
+    - frank@mozilla.com
+  url: 'https://github.com/mozilla-lockwise/lockwise-android'
+  dependencies:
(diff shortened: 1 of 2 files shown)


repositories.yaml
```diff
@@ -123,3 +123,19 @@ firefox-reality:
   dependencies:
     - org.mozilla.components:service-glean
     - org.mozilla.components:support-sync-telemetry
+lockwise-ios:
+  app_id: org.mozilla.ios.Lockbox
+  notification_emails:
+    - frank@mozilla.com
+    - tlong@mozilla.com
```
Dexterp37: I think this should contain also people from the Lockwise team, shouldn't it?

repositories.yaml
```diff
@@ -123,3 +123,19 @@ firefox-reality:
   dependencies:
     - org.mozilla.components:service-glean
     - org.mozilla.components:support-sync-telemetry
+lockwise-ios:
```
Dexterp37: Will you be tackling Android as a separate PR? (https://bugzilla.mozilla.org/show_bug.cgi?id=1600613)

repositories.yaml
Dexterp37: Maybe `lockwise-dev@mozilla.com`?

repositories.yaml
fbertsch: Sure, we can add Android while we're at it

repositories.yaml
fbertsch: @travis79 can you confirm who the recipients should be?

repositories.yaml
```diff
@@ -123,3 +123,26 @@ firefox-reality:
...
+  notification_emails:
+    - frank@mozilla.com
+    - chutten@mozilla.com
+  url: 'https://github.com/mozilla/gecko-dev'
+  dependencies:
+    - org.mozilla.components:service-glean
+lockwise-android:
+  app_id: mozilla.Lockbox
+  notification_emails:
+    - frank@mozilla.com
```
Dexterp37: I think we should add `lockwise-dev@mozilla.com` here as well

repositories.yaml
travis79: lockwise-dev@mozilla.com is good and maybe ddurst@mozilla.com?

repositories.yaml
fbertsch: Cool, added

repositories.yaml
fbertsch: Done and done

repositories.yaml
```diff
@@ -123,3 +123,28 @@ firefox-reality:
...
+lockwise-ios:
+  app_id: org.mozilla.ios.Lockbox
+  notification_emails:
+    - frank@mozilla.com
+    - tlong@mozilla.com
+    - lockwise-dev@mozilla.com
+  url: 'https://github.com/mozilla-lockwise/lockwise-ios'
+  dependencies:
+    - org.mozilla.components:service-glean
+firefox-desktop:
```
chutten: The fogotype is a "burn after using" prototype. Will attaching it to the "firefox-desktop" name here cause naming collisions down the line when we remove this one after fogotype's end, and want to add it back when we integrate the Glean SDK for real?

fbertsch: It will indeed cause naming collisions. If this is a prototype let's use a different name. @chutten opinions on what it should be? `fog` seems short and unclear.

chutten: `fogotype` works for me, as that's the "application" (heavy use of airquotes) that's sending the "prototype" pings with the app id `org.mozilla.fogotype`. Sound good, @fbertsch ?

chutten:
@fbertsch Anything I can help unblock on this?

fbertsch:
@chutten we should be good to go to merge, I was waiting for someone to approve :) @Dexterp37?

Dexterp37:
Redirecting to @mdboom as I'm at the end of my day

PR #144, opened 2019-12-10
Parse pings.yaml files and write historic information 

### LINKED ISSUE #143: Support pings.yaml in the scraper
We do have pings without additional metrics, which means the only way to find them is by their pings.yaml

The scraper needs to scrape those.

badboy:
Fixes #143

---

Here's a sample output: https://tmp.fnordig.de/priv/mozilla/pings.json

It is parsed from [pings.yaml](https://github.com/mozilla/glean/blob/master/glean-core/pings.yaml).


probe_scraper/parsers/pings.py
```diff
@@ -0,0 +1,29 @@
...
+
+from glean_parser.parser import parse_objects
+from pathlib import Path
+
+
+class GleanPingsParser:
+    """
+    Use the [Glean Parser]
+    (https://mozilla.github.io/glean_parser)
+    to parse the metrics.yaml files.
```
mdboom:
```suggestion
    to parse the pings.yaml files.
```

probe_scraper/transform_pings.py
```diff
@@ -0,0 +1,140 @@
...
+        def1.get(l) == def2.get(l)
+        for l in {
+            'bugs',
+            'data_reviews',
+            'description',
+            'notification_emails',
+            'include_client_id',
+            'send_if_empty',
+        }
+     ))
```
mdboom: Can't we just compare the dictionaries here?  The schema doesn't allow extra keys, so there shouldn't be anything about this that would be different from just `def1 == def2`.

probe_scraper/transform_pings.py
```diff
@@ -0,0 +1,140 @@
...
+            for ping, definition in pings.items():
+                repo_pings = update_or_add_ping(repo_pings,
+                                                commit_hash,
+                                                ping,
+                                                definition,
+                                                commit_timestamps[repo_name])
+
+        all_pings[repo_name] = repo_pings
+
+    return all_pings
```
mdboom: This seems really redundant with transform_metrics.py.  Maybe we can just put the pings as in the `pings` category in the metrics data structure (this is what glean_parser does).  Then it can all be handled by the same code...

probe_scraper/transform_pings.py
badboy: We probably can! I mostly copy-pasted the existing implementation. Will simplify.

probe_scraper/transform_pings.py
badboy: Unfortunately the transform_probes part hard-codes the `TYPE_KEY`, which pings don't have. 
It might be easier to just refactor out that small part and otherwise reuse the code.

mdboom: :+1:

schemas/repositories.json
```diff
@@ -26,7 +26,7 @@
         "type": "array",
         "items": {
           "type": "string",
-          "pattern": "metrics\\.yaml$"
+          "pattern": "(metrics|pings)\\.yaml$"
```
fbertsch: This shouldn't go under `metrics_files`, but should go under `ping_files`

probe_scraper/runner.py
```diff
@@ -233,11 +246,27 @@ def load_glean_metrics(cache_dir, out_dir, repositories_file, dry_run, glean_rep
...
+                    results, errs = GLEAN_PINGS_PARSER.parse(pings_files, config)
+                    pings[repo_name][commit_hash] = results
+            except Exception:
+                msg = "Improper file in {}\n{}".format(', '.join(pings_files),
+                                                       traceback.format_exc())
+                emails[repo_name]["emails"].append({
+                    "subject": "Probe Scraper: Improper File",
+                    "message": msg
+                })
+
```
fbertsch: This section very closely mirrors the above for metrics. Let's try and pull this out into shared functionality.

fbertsch: We should also email on errors (as we do above).

probe_scraper/runner.py
```diff
@@ -233,10 +245,34 @@ def load_glean_metrics(cache_dir, out_dir, repositories_file, dry_run, glean_rep
...
+                    "message": msg
+                })
+            else:
+                if errs:
+                    msg = ("Error in processing commit {}\n"
+                           "Errors: [{}]").format(commit_hash, ".".join(errs))
+                    emails[repo_name]["emails"].append({
+                        "subject": "Probe Scraper: Error on Ping Parsing",
+                        "message": msg
+                    })
```
fbertsch: We should be able to combine this with the above metrics file processing.

probe_scraper/transform_probes.py
```diff
@@ -275,48 +275,74 @@ def metrics_equal(def1, def2):
...
+        def1.get(l) == def2.get(l)
+        for l in {
+            'bugs',
+            'data_reviews',
+            'description',
+            'notification_emails',
+            'include_client_id',
+            'send_if_empty',
+        }
+    ))
```
fbertsch: We should make this compare all fields. In the past we've had issues where a new field is added and then a change in it isn't present in the probe-info-service.

probe_scraper/transform_probes.py
```diff
@@ -275,48 +275,74 @@ def metrics_equal(def1, def2):
...
+
+
+def metric_ctor(defn, metric):
```
fbertsch: Weakly held opinion that we shouldn't use `ctor`, the full `constructor` seems more obvious.

badboy:
@fbertsch, can we land this?

PR #145, opened 2019-12-10
Handle the multiple metrics files in glean

mdboom:
https://github.com/mozilla/glean/pull/448 refactored the metrics.yaml files in
glean into a cross-platform file and an Android-specific file. Both are
required to know all of the metrics that Glean sends.

We were "getting lucky" before now since we didn't add any metrics, so they
were getting included by being in the historical versions of
`glean-core/android/metrics.yaml`.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -6,6 +6,7 @@ glean:
   url: 'https://github.com/mozilla/glean'
   metrics_files:
     - 'glean-core/android/metrics.yaml'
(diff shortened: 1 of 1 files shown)


mdboom:
I tested it locally an it seems to do the right thing.  However, I'm not sure how to write a unit test for it...

fbertsch:
No need for a unit test, locally testing should be fine.

PR #148, opened 2019-12-19
Add appservices:logins component.

rfk:
Over in https://github.com/mozilla/application-services/pull/2225 we landed some metrics for the application-services "logins" component. The new release hasn't worked its way into Fenix yet, but it will shortly, and I'm told that we need to get ourselves an entry in the file here before submitted metrics will start showing up for analysis.

Does this look about right?

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -26,6 +26,7 @@ fenix:
     - org.mozilla.components:lib-crash
     - org.mozilla.components:support-sync-telemetry
     - org.mozilla.components:browser-engine-gecko-beta
+    - org.mozilla.appservices:logins
 reference-browser:
   app_id: org-mozilla-reference-browser
   notification_emails:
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -26,6 +26,7 @@ fenix:
     - org.mozilla.components:lib-crash
     - org.mozilla.components:support-sync-telemetry
     - org.mozilla.components:browser-engine-gecko-beta
+    - org.mozilla.appservices:logins
```
mdboom: This line also needs to be added to fenix-nightly below (I assume this data is collected by fenix nightly as well).

rfk: Thanks, added.

vladikoff:
@mdboom Ryan is out on PTO, could you please merge when you possible?

vladikoff:
Thank you!!

PR #150, opened 2020-01-07
Bug 1607225

fbertsch:


probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -175,8 +176,14 @@ def relative_path_is_in_version(rel_path, version):
     return True
 
 
-def download_files(channel, node, temp_dir, error_cache, version):
-    base_uri = CHANNELS[channel]['base_uri'] + 'raw-file/' + node + '/'
+def download_files(channel, node, temp_dir, error_cache, version, tree=None):
+
+    if tree is None:
```
Dexterp37: Just to be sure: when is `tree` expected to be `None`? Is this an error condition?

fbertsch: It's `None` here: https://github.com/mozilla/probe-scraper/pull/150/files#diff-00aa03ed5ca38a33de9a85416a639da6L282

PR #153, opened 2020-01-09
Bug 1511035 - Add probe expiry alerting

Ben-Wu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1511035

This sends emails to probe owners two weeks before the release of a firefox version in which a probe would expire.
Much of the release calendar scraping logic was copied from https://github.com/mozilla/cerberus/blob/master/alert/expiring.py

The idea is to run this as a task in the probe scraper dag after the scrape task so I think it makes sense to keep it in this repo rather than separately.


probe_scraper/probe_expiry_alert.py
```diff
@@ -0,0 +1,106 @@
...
+
+def send_emails_for_expiring_probes(target_date, expiring_histograms_by_channel, dryrun=True):
+    histograms_by_email = defaultdict(lambda: defaultdict(list))
+
+    for channel, histograms in expiring_histograms_by_channel.items():
+        for name, emails in histograms.items():
+            for email in emails:
+                histograms_by_email[email][channel].append(name)
+
+    email_body_format_string = """
```
fbertsch: I think we can clean this message up a bit to make it more actionable. Something like:

```
The following Firefox probes will expire in the next {expire_days} days or have already expired.

What to do about this:
1. If the metric is no longer needed, ignore this warning. You can remove it from `histograms.json`[link] or `scalars.yaml`[link], but then it will not be available to view in telemetry.mozilla.org (however, it will still be queryable).
2. If the metric is still required, resubmit a data review [1] and extend its expiration date.
If you have any problems, please ask for help on the #fx-metrics Slack channel. We'll give you a hand.

Your Friendly, Neighborhood Telemetry Team
[1] https://wiki.mozilla.org/Firefox/Data_Collection
This is an automated message sent from probe-scraper.  See https://github.com/mozilla/probe-scraper for details.
```

cc @chutten

probe_scraper/probe_expiry_alert.py
```diff
@@ -0,0 +1,106 @@
...
+This is an automated message from probe-scraper (https://github.com/mozilla/probe-scraper).
+    """
+    per_channel_format_string = """
+On {}:
+{}
+    """
+
+    for email, histograms_by_channel in histograms_by_email.items():
+        per_channel_format_strings = [
+            per_channel_format_string.format(channel, "\n".join(histogram_names))
```
fbertsch: Might be worth it to make this wordier as well.

probe_scraper/probe_expiry_alert.py
```diff
@@ -0,0 +1,106 @@
...
+        "--dry-run",
+        help="Whether emails should be sent",
+        action="store_true",
+    )
+    return parser.parse_args()
+
+
+def main(dryrun):
+    probe_info = requests.get(PROBE_INFO_BASE_URL + "firefox/all/main/all_probes").json()
+    target_date = datetime.date.today() + EMAIL_TIME_BEFORE
```
fbertsch: I'm considering whether we should have multiple reminders, at 2 weeks and 1 week.

probe_scraper/scrapers/release_calendar.py
```diff
@@ -0,0 +1,105 @@
...
+from bs4 import BeautifulSoup
+import requests
+
+from probe_scraper.parsers.utils import get_major_version
+
+RELEASE_CALENDAR_URL = ("https://wiki.mozilla.org/api.php"
+                        "?action=parse&format=json&page=Release_Management/Calendar")
+
+
+def get_version_table_dates(table):
```
fbertsch: Some info here: https://mozilla.slack.com/archives/C4D5ZA91B/p1576163365064000

I'm not necessarily sure that it's a better alternative, but we could use https://calendar.google.com/calendar/embed?src=bW96aWxsYS5jb21fZGJxODRhbnI5aTh0Y25taGFiYXRzdHY1Y29AZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ

probe_scraper/probe_expiry_alert.py
chutten: For Case 1 I'd like to de-emphasize that "do nothing" is an acceptable choice. We want people to clean up their probes (and in the case of the version increase simulator resulting in sheriffs filing bugs like [this one](https://bugzilla.mozilla.org/show_bug.cgi?id=1607737), this might be working against what we've previous said), so maybe something like.

```
1. If one, some, or all of the metrics are no longer needed, please remove them from their definitions files (Histograms.json, Scalars.yaml, Events.yaml).
2. If one, some, or all of the metrics are still required, please submit a Data Collection Review and patch to extend their expiry.
```

Two things that definitely need to be watched are 
1. Histogram.json/Scalars.yaml are Capital Case
2. It's an expiry, not an expiry date. (Expiry dates are for Glean)

Otherwise: all good.

probe_scraper/probe_expiry_alert.py
chutten: I recommend using the behaviour of the Glean expiry emails (whatever it is) so that we have consistency.

probe_scraper/scrapers/release_calendar.py
Ben-Wu: Parsing through the calendar seems even messier so I'll stick with this for now and maybe revisit one day

fbertsch: That seems fine, can you file an issue for it?

chutten:
> 
> 
> I agree with Chris that we should get this in. We need to decide on two things then:
> 
>     1. Wording - I agree with the points Chris made, and I'll leave it to him to decide on whether to let people know about telemetry-dashboard. Let's include those other points of cleanup

I say if they were relying on TMO dashboards to inform their decisionmaking, having the data disappear on them is an excellent opportunity to let them know it was a mistake to do so : D

(( and I volunteer to write the SQL to get them up and running using sql.tmo ))

Ben-Wu:
I'm updating the email now and comment it here in a bit.  I think it makes sense to run it weekly with a 14 day notification window like glean.  This way we can send another notification at the 1 week mark.

Ben-Wu:
Is this text fine to go with? @fbertsch @chutten

```
The following Firefox probes will expire in the next 14 days or have already expired.


These probes are expiring on the beta channel:
expiring_probe_1
expiring_probe_2


These probes are expiring on the release channel:
expiring_probe_2


What to do about this:
1. If one, some, or all of the metrics are no longer needed, please remove them from their definitions files (Histograms.json, Scalars.yaml, Events.yaml).
2. If one, some, or all of the metrics are still required, please submit a Data Collection Review and patch to extend their expiry.

If you have any problems, please ask for help on the #fx-metrics Slack channel. We'll give you a hand.

Your Friendly, Neighborhood Telemetry Team

[1] https://wiki.mozilla.org/Firefox/Data_Collection

This is an automated message sent from probe-scraper.  See https://github.com/mozilla/probe-scraper for details.
```

chutten:
LGTM!

fbertsch:
@Ben-Wu that looks great. 🚢

PR #155, opened 2020-01-14
Kebab appids

fbertsch:
This will fix the schema deploys.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -137,7 +137,7 @@ firefox-reality:
     - org.mozilla.components:service-glean
     - org.mozilla.components:support-sync-telemetry
 fogotype:
-  app_id: org.mozilla.fogotype
+  app_id: org-mozilla-fogotype
   notification_emails:
     - frank@mozilla.com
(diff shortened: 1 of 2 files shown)


fbertsch:
Fixes https://github.com/mozilla/mozilla-schema-generator/issues/91

repositories.yaml
```diff
@@ -137,23 +137,23 @@ firefox-reality:
     - org.mozilla.components:service-glean
     - org.mozilla.components:support-sync-telemetry
 fogotype:
-  app_id: org.mozilla.fogotype
```
Dexterp37: fly-by: can't the probe-scraper do the transformation itself? It might be clearer to specify the same app ids that the app advertised in this repository file...

fbertsch: Yeah, that is how we should handle these. However let's first get this moving again then we can move this responsibility to MSG.

fbertsch: https://github.com/mozilla/probe-scraper/issues/156

PR #158, opened 2020-01-14
Make probe expiry alerts only run on mondays

Ben-Wu:
To copy glean alerts behaviour.  I thought it would be easy to do this in airflow but it is not because the probe scraper dag runs daily.


chutten:
Since version increases happen on Mondays (last one happened at 3pm UTC Monday Jan 6), maybe choose Tuesday or Wednesday instead?

Ben-Wu:
OK changed to tuesday

PR #162, opened 2020-01-15
Make app_ids lowercase

fbertsch:
Also added the length requirement.

### DIFF
--- a/schemas/repositories.json
+++ b/schemas/repositories.json
@@ -7,7 +7,7 @@
     "properties": {
       "app_id": {
         "type": "string",
-        "format": "^[a-zA-Z][a-zA-Z-]*$"
+        "pattern": "^[a-z][a-z-]{0,1023}$"
       },
       "notification_emails": {
         "type": "array",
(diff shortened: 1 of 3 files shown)


schemas/repositories.json
```diff
@@ -7,7 +7,7 @@
     "properties": {
       "app_id": {
         "type": "string",
-        "format": "^[a-zA-Z][a-zA-Z-]*$"
```
Dexterp37: Is this something we should enforce in the SDK as well?

fbertsch: I don't think so. App id is transformed in the pipeline so it would be different constraints. What's the likelihood of someone requiring an app id that doesn't match those requirements?

mdboom: I'm not aware of any requirements -- these specific ids in these changes came from iOS where maybe upper case letters are a thing (they aren't on Android), but I doubt they are case-sensitive -- even if they were, Mozilla controls the names so it's very unlikely that differences only in case would matter.

PR #163, opened 2020-01-16
Change probe expiry alerts to weekly for prev and next versions

### LINKED ISSUE #157: Use a more robust firefox release date data source
In probe expiry alerts, we are currently scraping the firefox release calendar webpage for the release dates.  This isn't robust because if the html structure changes it will either fail or return incorrect data.  If there is a better alternate data source (e.g. json api), that should be used instead.

[Current code](https://github.com/mozilla/probe-scraper/blob/e99d2e4c50813d9b43929c57b6f564391cb49156/probe_scraper/scrapers/release_calendar.py)

Release dates are in this [google calendar](https://calendar.google.com/calendar/embed?src=bW96aWxsYS5jb21fZGJxODRhbnI5aTh0Y25taGFiYXRzdHY1Y29AZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ) but it also isn't an ideal source for the same reasons above.

Ben-Wu:
(closes #157)

Now sends weekly emails for probes expiring in the next version and probes expired in the latest version.  Also adds some logging for airflow

Email body look like this:
```
The following Firefox probes have either expired or will expire in the next major Firefox release: version 74 on the release channel, 75 on the beta channel, and 76 on the nightly channel.


The following probes are expiring on the beta channel in Firefox version 76:
expiring_probe_1
expiring_probe_2


The following probes are expiring on the release channel in Firefox version 75:
expiring_probe_2


The following probes have expired on the nightly channel in Firefox version 76:
expired_probe_2


The following probes have expired on the release channel in Firefox version 74:
expired_probe_1


What to do about this:
1. If one, some, or all of the metrics are no longer needed, please remove them from their definitions files (Histograms.json, Scalars.yaml, Events.yaml).
2. If one, some, or all of the metrics are still required, please submit a Data Collection Review and patch to extend their expiry.

If you have any problems, please ask for help on the #fx-metrics Slack channel. We'll give you a hand.

Your Friendly, Neighborhood Telemetry Team

[1] https://wiki.mozilla.org/Firefox/Data_Collection

This is an automated message sent from probe-scraper.  See https://github.com/mozilla/probe-scraper for details.
```

cc @chutten


chutten:
By splitting all of the channels into separate paragraphs... does that mean if I have a probe expiring in version N, then when Nightly hits N-1 I'll start getting emails and then I won't stop getting emails until Release hist N+1 ?

Ben-Wu:
Yes I made it this way because I noticed some probes that have different expiry versions depending on channel (although it's rare).  But now that I think about it, that can be annoying if someone doesn't extend the expiry because they don't need it anymore.  Do you think it should notify on nightly only?

chutten:
I say nightly only. If a probe has different expiry versions on different channels, that might be the sign of a bug. To my knowledge there is no mechanism by which you can specify different expiry per channel.

...unless there's a patch to renew the probe and the patch hasn't made it through all the channels yet?

(( and if that's the case, we don't want to alert on that probe any more because it _has_ been renewed. Beta and Release just haven't gotten the news yet ))

Ben-Wu:
Sounds good to me.  Now checking nightly only.  the text looks like this:

```
The following Firefox probes have either expired or will expire in the next major Firefox nightly release: version 76.


The following probes are expiring in Firefox version 76:
expiring_probe_1
expiring_probe_2


The following probes have expired in Firefox version 75:
expired_probe_1
expired_probe_2

...
```

probe_scraper/probe_expiry_alert.py
```diff
@@ -5,21 +5,21 @@
...
 
 EMAIL_BODY_FORMAT_STRING = """
-The following Firefox probes will expire in the next {expire_days} days or have already expired.
+The following Firefox probes have either expired or will expire in the next major Firefox nightly release: version {next_version}.
 
-{per_channel_probes}
+{probes}
 
 What to do about this:
 1. If one, some, or all of the metrics are no longer needed, please remove them from their definitions files (Histograms.json, Scalars.yaml, Events.yaml).
```
chutten: Huh, missed this last time, but we have a `[1]` footnote for data collection review, but we don't reference it in "What to do about this" number 2.

Ben-Wu:
OK added the release calendar as a footnote and added the references

PR #172, opened 2020-02-06
Get expiring probes from hg.moz.org instead of probeinfo

### LINKED ISSUE #171: Telemetry Probe Expiry emails are sent for probes that have been removed from the code
At the beginning of the 74 cycle we got an email about some expired/expiring probes:
```
The following probes are expiring in Firefox version 75:
NETWORK_CROSS_ORIGIN_STYLESHEET_CONTENT_TYPE


The following probes have expired in Firefox version 74:
NETWORK_CACHE_ISOLATION_ENTRY_COUNT_INCREASE
NETWORK_CACHE_ISOLATION_SIZE_INCREASE
NETWORK_CACHE_ISOLATION_UNIQUE_SITE_ACCESS_COUNT
```

These probes were removed in [bug 1607468](https://bugzil.la/1607468) and [bug 1606207](https://bugzil.la/1606207) but we are still getting emails about them every week or so. I believe this is a bug in the probe scraper.

Ben-Wu:
(fixes #171)

Instead of using the probe info service, now `Events.yaml`, `Histograms.json`, and `Scalars.yaml` from the latest revision are used so that probes that have been removed aren't included.


probe_scraper/probe_expiry_alert.py
```diff
@@ -109,6 +110,12 @@ def send_emails_for_expiring_probes(expired_probes, expiring_probes,
     logging.info(f"Sent emails to {len(probes_by_email_by_state)} recipients")
 
 
+def download_file(url, output_filename):
+    content = requests.get(url).text
+    with open(output_filename, "w") as scalars_file:
+        scalars_file.write(content)
```
relud: I feel like this should use a temp file, e.g. use [NamedTemporaryFile](https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile) and accept a parse method:
```python
from functools import partial
from tempfile import NamedTemporaryFile

...

def download_file(url, parse):
    content = requests.get(url).text
    with NamedTemporaryFile(text=True) as temp_file:
        temp_file.write(content)
        return parse([temp_file.name])

...

def main(current_date, dryrun):
    current_version = get_latest_nightly_version()
    next_version = str(int(current_version) + 1)

    events = download_file(BASE_URI + EVENTS_FILE, EventsParser().parse)

    histograms = download_file(
        BASE_URI + HISTOGRAMS_FILE,
        partial(HistogramsParser().parse, version=next_version),
    )

    scalars = download_file(BASE_URI + SCALARS_FILE, ScalarsParser().parse)
```

Ben-Wu: The histogram parser only accepts certain filenames.  I do agree that something temporary should be used though.  Instead I used `gettempdir()`

probe_scraper/probe_expiry_alert.py
```diff
@@ -120,20 +129,37 @@ def parse_args():
...
-    probe_info = requests.get(PROBE_INFO_BASE_URL + "firefox/all/main/all_probes").json()
-
     current_version = get_latest_nightly_version()
     next_version = str(int(current_version) + 1)
 
-    expired_probes = find_expiring_probes(probe_info, current_version)
-    expiring_probes = find_expiring_probes(probe_info, next_version)
+    events_file_path = os.path.join(tempfile.gettempdir(), "Events.yaml")
+    download_file(BASE_URI + EVENTS_FILE, events_file_path)
+    events = EventsParser().parse([os.path.join(tempfile.gettempdir(), events_file_path)])
```
relud: this should reuse `events_file_path`
```suggestion
    events = EventsParser().parse([events_file_path])
```

probe_scraper/probe_expiry_alert.py
```diff
@@ -120,20 +129,37 @@ def parse_args():
...
+
+    histograms_file_path = os.path.join(tempfile.gettempdir(), "Histograms.json")
+    download_file(BASE_URI + HISTOGRAMS_FILE, histograms_file_path)
+    histograms = HistogramsParser().parse(
+        [os.path.join(tempfile.gettempdir(), histograms_file_path)],
```
relud: this should reuse `histograms_file_path`
```suggestion
        [histograms_file_path],
```

probe_scraper/probe_expiry_alert.py
```diff
@@ -120,20 +129,37 @@ def parse_args():
...
+        version=next_version
+    )
+
+    scalars_file_path = os.path.join(tempfile.gettempdir(), "Scalars.yaml")
+    download_file(BASE_URI + SCALARS_FILE, scalars_file_path)
+    scalars = ScalarsParser().parse([os.path.join(tempfile.gettempdir(), scalars_file_path)])
```
relud: this should reuse `scalars_file_path`
```suggestion
    scalars = ScalarsParser().parse([scalars_file_path])
```

probe_scraper/probe_expiry_alert.py
```diff
@@ -120,20 +129,37 @@ def parse_args():
...
 
 def main(current_date, dryrun):
-    probe_info = requests.get(PROBE_INFO_BASE_URL + "firefox/all/main/all_probes").json()
-
     current_version = get_latest_nightly_version()
     next_version = str(int(current_version) + 1)
 
-    expired_probes = find_expiring_probes(probe_info, current_version)
-    expiring_probes = find_expiring_probes(probe_info, next_version)
+    events_file_path = os.path.join(tempfile.gettempdir(), "Events.yaml")
```
relud: this should use `EVENTS_FILE`, and a context manager with a `TemporaryDirectory` rather than creating one for each file and not cleaning up:
```suggestion
    with tempfile.TemporaryDirectory() as tempdir:
        events_file_path = os.path.join(tempdir, EVENTS_FILE)
```
(and indent down to `scalars = ...`)

probe_scraper/probe_expiry_alert.py
```diff
@@ -120,20 +129,37 @@ def parse_args():
...
+    download_file(BASE_URI + EVENTS_FILE, events_file_path)
+    events = EventsParser().parse([os.path.join(tempfile.gettempdir(), events_file_path)])
+
+    histograms_file_path = os.path.join(tempfile.gettempdir(), "Histograms.json")
```
relud: this should reuse `tempdir` and `HISTOGRAMS_FILE`
```suggestion
    histograms_file_path = os.path.join(tempdir, HISTOGRAMS_FILE)
```

probe_scraper/probe_expiry_alert.py
```diff
@@ -120,20 +129,37 @@ def parse_args():
...
+    download_file(BASE_URI + HISTOGRAMS_FILE, histograms_file_path)
+    histograms = HistogramsParser().parse(
+        [os.path.join(tempfile.gettempdir(), histograms_file_path)],
+        version=next_version
+    )
+
+    scalars_file_path = os.path.join(tempfile.gettempdir(), "Scalars.yaml")
```
relud: this should reuse `tempdir` and `SCALARS_FILE`
```suggestion
    scalars_file_path = os.path.join(tempdir, SCALARS_FILE)
```

PR #173, opened 2020-02-11
Add tip revision tag to general

Ben-Wu:
This seems to be how transform probes finds last revision (by sorting by version)

Output looks something like:
```
{
  "lastUpdate": "2020-02-11T12:50:48.299557-05:00",
  "tip": {
    "nightly": "11c9c5ce395543b8584722a96e52655277ec49ee"
  }
}
```

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -51,10 +51,22 @@ GLEAN_METRICS_FILENAME = 'metrics.yaml'
 GLEAN_PINGS_FILENAME = 'pings.yaml'
 
 
-def general_data():
-    return {
+def general_data(revisions=None):
+    output_data = {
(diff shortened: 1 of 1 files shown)


fbertsch:
Huh, looks like we can get the tip revision from https://probeinfo.telemetry.mozilla.org/firefox/revisions?

fbertsch:
Given that, do we still need this?

Ben-Wu:
I suggested that yesterday but you said that it's not guaranteed to be the latest.  Although after digging through the code more I think `/revisions` is doing roughly the same thing as the changes in this PR so I assume it will give the tip revision.  So I don't think this PR is needed

fbertsch:
@Ben-Wu I must have misunderstood, I agree you can do this from `/firefox/revisions`.

PR #174, opened 2020-02-11
Look at last revision to determine if probe still exists

Ben-Wu:
@fbertsch This assumes the highest version is unique in https://probeinfo.telemetry.mozilla.org/firefox/revisions.  But I noticed that in nightly there are two entries for 71.  Any ideas for the cause and could this be a problem?

### DIFF
--- a/probe_scraper/probe_expiry_alert.py
+++ b/probe_scraper/probe_expiry_alert.py
@@ -112,12 +112,9 @@ def send_emails_for_expiring_probes(expired_probes, expiring_probes,
 
 
 def get_latest_revision():
-    revision_info = requests.get(PROBE_INFO_BASE_URL + "firefox/revisions").json()["nightly"]
-    sorted_revisions = sorted(
-        [{'tag': tag, 'version': info['version']} for tag, info in revision_info.items()],
-        key=lambda x: x['version'], reverse=True
-    )
-    return sorted_revisions[0]['tag']
+    revision_info = requests.get(
+        f"{PROBE_INFO_BASE_URL}firefox/revisions?p={datetime.datetime.now()}").json()["nightly"]
+    return max(revision_info.items(), key=lambda x: x[1]['version'])[0]
 
 
 def parse_args():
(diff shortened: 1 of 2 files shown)


probe_scraper/probe_expiry_alert.py
```diff
@@ -112,10 +111,13 @@ def send_emails_for_expiring_probes(expired_probes, expiring_probes,
...
-    content = requests.get(url).text
-    with open(output_filepath, "w") as output_file:
-        output_file.write(content)
+def get_latest_revision():
+    revision_info = requests.get(PROBE_INFO_BASE_URL + "firefox/revisions").json()["nightly"]
+    sorted_revisions = sorted(
+        [{'tag': tag, 'version': info['version']} for tag, info in revision_info.items()],
+        key=lambda x: x['version'], reverse=True
+    )
+    return sorted_revisions[0]['tag']
```
fbertsch: Could use `max` here instead.
```suggestion
    return max(revision_info.items(), key=lambda x: x[1]['version'])[0]
```

probe_scraper/probe_expiry_alert.py
```diff
@@ -129,28 +131,14 @@ def parse_args():
 
 
 def main(current_date, dryrun):
+    probe_info = requests.get(PROBE_INFO_BASE_URL + "firefox/all/main/all_probes").json()
```
fbertsch: For the schema pipeline, we realized that there was a delay from cloudfront of when the probe-info-service actually updated. To accommodate that we added a parameter to the request with a timestamp, which now forces cloudfront to return the current copy. We should probably do this above as well (for /revisions)
```suggestion
    probe_info = requests.get(f"{PROBE_INFO_BASE_URL}firefox/all/main/all_probes?p={datetime.datetime.now()}").json()
```

Ben-Wu:
Closing because very outdated (before switch to creating bugs).  Problem is that latest revision from probe info service is potentially not determinable, and effort to fix is high.

PR #176, opened 2020-02-19
Ensure contents served at probes.tmo are gzipped

wlach:
Cloudfront doesn't automatically apply compression if the file is
larger than 10 megabytes (our probes file is currently 25M), but we
can get around this by gzipping on our end.

This helps make mozilla/probe-dictionary#80 less severe.

I haven't tried this end-to-end, but have tested that the python code + s3 commandline locally and it seems to work correctly.

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -7,6 +7,7 @@ from collections import defaultdict
 import datetime
 from dateutil.tz import tzlocal
 import errno
+import gzip
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


probe_scraper/runner.py
```diff
@@ -310,15 +311,26 @@ def sync_output_and_cache_dirs(output_bucket, cache_bucket, out_dir, cache_dir,
...
+
+        # cloudfront is supposed to automatically gzip objects, but it won't do that
+        # if the object size is > 10 megabytes (https://webmasters.stackexchange.com/a/111734)
+        # which our files sometimes are. to work around this, we'll regzip the contents into a
+        # temporary directory, and upload that with a special content encoding
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            for filename in os.listdir(out_dir):
+                with open(f'{out_dir}/{filename}', 'rb') as f1:
+                    with gzip.open(f'{tmpdirname}/{filename}', 'wb') as f2:
+                        f2.write(f1.read())
```
robhudson: Do we want to check the size before gzipping and only gzip the big files?

Also, consistent quoting would be good here.

probe_scraper/runner.py
```diff
@@ -310,15 +311,26 @@ def sync_output_and_cache_dirs(output_bucket, cache_bucket, out_dir, cache_dir,
...
-            f"--acl public-read"
-        )
-        os.system(sync_cft_cmd)
+
+        # cloudfront is supposed to automatically gzip objects, but it won't do that
+        # if the object size is > 10 megabytes (https://webmasters.stackexchange.com/a/111734)
+        # which our files sometimes are. to work around this, we'll regzip the contents into a
+        # temporary directory, and upload that with a special content encoding
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            for filename in os.listdir(out_dir):
```
jasonthomas: `os.listdir` will include both files and directories which will be problematic later. I think `os.walk` might be better suited here which will split out directories and files.

probe_scraper/runner.py
wlach: I would vote that we just compress everything, as doing it selectively would require a lot of extra code. What do you mean by consistent quoting?

wlach:
Build issues have nothing to do with this PR, they are fixed in the (just filed) #177

probe_scraper/runner.py
robhudson: Okay, I didn't know if cloudfront was fine zipping everything. Sounds good. Above you use single quotes and below you use double quotes. It's a nitpick.

probe_scraper/runner.py
wlach: this is a bit of a pain but you are probably right... fixing

wlach: Done, it is a little complex so would be good to test again before deploying

probe_scraper/runner.py
wlach: Gotcha! Easy to fix. We should probably use black to enforce this sort of thing...

wlach:
On second thought, I'm fairly confident this will work (since I tested locally with a script). I'll monitor airflow and backout if there are issues in production.

PR #180, opened 2020-03-09
Keep git commits sorted in their log order

mdboom:
probe-scraper currently sorts commits based on the committer time, which is the time that commits in a PR are merged into master.  While this is mostly correct, all commits in a PR end up having the exact same timestamp, so their sorting order is non-deterministic.

This retains the order that the commits were reported from git-log (in addition to the timestamp), so that the correct sorting of commits *within* PRs is retained.

This should resolve the problem we were seeing [here](https://github.com/mozilla/mozilla-schema-generator/issues/118), where commits that changed a metric type within a PR would non-deterministically select the ultimate metric type.

### DIFF
--- a/probe_scraper/glean_checks.py
+++ b/probe_scraper/glean_checks.py
@@ -178,7 +178,7 @@ def check_for_expired_metrics(
     for repo_name, commits in repos_metrics.items():
         repo = repo_by_name[repo_name]
         timestamps = list(commit_timestamps[repo_name].items())
-        timestamps.sort(key=lambda x: x[1], reverse=True)
+        timestamps.sort(key=lambda x: (-x[1][0], x[1][1]))
         last_commit_hash = timestamps[0][0]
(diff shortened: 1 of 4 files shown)


Dexterp37:
(wow, good catch)

mdboom:
> We need to use the index (in addition to ts) when sorting the probes in transform_probes

This is already the case...  I updated the docstring to make it clear that's the new expected input.

> We should add the index as another output, so we can sort independently of the probe-scraper's ordering. We could add another field to each history entry that is e.g. reflog-index, git-index, or commit-index and looks something like: {'first': 3, 'last': 5}. We would guarantee that no history entries would overlap, so the next entry would have {'first': 4, ...}.

That's easy enough to do.  Unlike the date stamps, these are likely to change on every run of probe_scraper.  Do we think that will be a problem?

> Tests need to be updated with this new data structure

Indeed.

PR #183, opened 2020-04-07
FOGotype is dead, long live Project FOG

badboy:
Is it safe to just remove it?
We deactivated it on Firefox Nightly a while ago and are not interested in any of that data anymore.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -138,16 +138,6 @@ firefox-reality:
   dependencies:
     - org.mozilla.components:service-glean
     - org.mozilla.components:support-sync-telemetry
-fogotype:
-  app_id: org-mozilla-fogotype
-  notification_emails:
-    - frank@mozilla.com
-    - chutten@mozilla.com
-  url: 'https://github.com/mozilla/gecko-dev'
-  dependencies:
-    - org.mozilla.components:service-glean
-  ping_files:
-    - 'toolkit/components/telemetry/fog/pings.yaml'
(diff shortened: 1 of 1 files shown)


chutten:
If pipeline says we can remove it, then we should.

@fbertsch can we also deep-six the data that's already in BQ? Is there a process for that?

fbertsch:
@chutten that will happen automatically! This is a bit of new territory for us, so no official process in place.

jklukas:
> Can we also deep-six the data that's already in BQ? Is there a process for that?

cc @whd. I don't believe we've yet done this in prod, and there may be some hand-holding needed to get out terraform state into shape to handle the deletion of tables/datasets.

chutten:
Oh wait I see that this will happen automagically in the bug. Whoops.

whd:
> cc @whd. I don't believe we've yet done this in prod, and there may be some hand-holding needed to get out terraform state into shape to handle the deletion of tables/datasets.

We haven't done this in prod, but the process is the same as stage where we've done this three or four times. I've mentioned it in https://bugzilla.mozilla.org/show_bug.cgi?id=1626490#c8 and if @mreid-moz signs off I will allow this deletion to propagate to stage and prod.

whd:
@mreid-moz signed off on the deletion in https://bugzilla.mozilla.org/show_bug.cgi?id=1626490#c12 so I think we're good to proceed from a procedural perspective (I see tests are failing however).

PR #184, opened 2020-04-08
Bug 1628320 - Add mozregression

wlach:


wlach:
> Hey William, just want to clarify a few things:
> 
>     1. Do we not want to add the service-glean dependency? That enables the default pings and metrics.

I don't think so, that stuff won't play well with a commandline application from what I understand.

>     2. [We're seeing](https://console.cloud.google.com/bigquery?project=moz-fx-data-ingesti-prod-579d&folder=&organizationId=&j=bq:US:bquxjob_4a5b94ac_1715ad135ad&page=queryresults) a `mozphab` namespace with 227 `usage` pings. Is that related to this?

Nope, that's an unrelated project (also a command-line application using the Glean python bindings though). See: https://bugzilla.mozilla.org/show_bug.cgi?id=1621025

Dexterp37:
>     1. Do we not want to add the service-glean dependency? That enables the default pings and metrics.

@wlach @fbertsch (cc @mdboom ) I think we need a dependency on glean here. *However*, [this dependency](https://github.com/mozilla/probe-scraper/blob/2d7b2d935a960e2c7f744b97f18e7ca9e4ac9f7d/repositories.yaml#L9) includes both the `core` metrics and the `android specific` metrics.

Going forward we might want to split that definition. `service-glean` is specifically for Android applications.

We should probably create an entry that leaves out Android specific metrics for clarity.


>     2. [We're seeing](https://console.cloud.google.com/bigquery?project=moz-fx-data-ingesti-prod-579d&folder=&organizationId=&j=bq:US:bquxjob_4a5b94ac_1715ad135ad&page=queryresults) a `mozphab` namespace with 227 `usage` pings. Is that related to this?

@fbertsch that's another python application being gleanified by the build system team. It's not in production yet, they'll go through this process once it is about to.

mdboom:
> @wlach @fbertsch (cc @mdboom ) I think we need a dependency on glean here. However, this dependency includes both the core metrics and the android specific metrics.

I don't think this dependency is needed for the core metrics (the ones in client_info and ping_info), since those are hardcoded in a different part of the JSON schema.  So I think we're ok with leaving this out.  However, renaming that to something that makes it more obvious it's for android is probably a good idea (but separate from this PR/issue).

mdboom:
Yes.  I think a PR to separate out platform-independent from Android-specific dependencies could be done as a follow-up.  (And unless I'm missing something, could be done entirely in probe_scraper).  For now, this PR is fine as-is and should work correctly for mozregression.

fbertsch:
@mdboom Are those dependencies already separated out? If so then yes we can add them in here.

mdboom:
I think we'd just need to add a new library in the metadata here with just the android `metrics.yaml` file [here](https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml#L8).  And then update all of the existing apps and libraries (except mozregression) to *also* have that one.  Feels a little finicky to get right, which is why I thought we could merge this as-is in the meantime.

fbertsch:
Agreed, let's merge this as-is and fix that later. @mdboom can you open an issue describing how you see it working?

mdboom:
Got it.  Just files #185.  I can take that (but probably later today at the earliest).

PR #188, opened 2020-05-12
Bug 1637065 - Add Firefox iOS App

fbertsch:
Don't merge until https://github.com/mozilla-mobile/firefox-ios/pull/6577 is merged.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -245,4 +245,4 @@ firefox-ios:
     frank@mozilla.com
   url: https://github.com/mozilla-mobile/firefox-ios
   metrics_files:
-    - 'Client/Generated/LegacyIds.swift'
+    - ' Client/metrics.yaml'
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -238,3 +238,11 @@ mozregression:
     - 'mozregression/metrics.yaml'
   ping_files:
     - 'mozregression/pings.yaml'
+firefox-ios:
+  app_id: org-mozilla-ios-fennec
+  notification_emails:
+    tlong@mozilla.com
+    frank@mozilla.com
+  url: https://github.com/mozilla-mobile/firefox-ios
+  metrics_files:
+    - 'Client/Generated/LegacyIds.swift'
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -238,8 +238,24 @@ mozregression:
     - 'mozregression/metrics.yaml'
   ping_files:
     - 'mozregression/pings.yaml'
-firefox-ios:
+firefox-ios-dev:
   app_id: org-mozilla-ios-fennec
+  notification_emails:
+    - tlong@mozilla.com
+    - frank@mozilla.com
+  url: https://github.com/mozilla-mobile/firefox-ios
+  metrics_files:
+    - 'Client/metrics.yaml'
+firefox-ios-beta:
+  app_id: org-mozilla-ios-firefoxbeta
+  notification_emails:
+    - tlong@mozilla.com
+    - frank@mozilla.com
+  url: https://github.com/mozilla-mobile/firefox-ios
+  metrics_files:
+    - 'Client/metrics.yaml'
+firefox-ios-release:
+  app_id: org-mozilla-ios-firefox
   notification_emails:
     - tlong@mozilla.com
     - frank@mozilla.com
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -245,4 +245,4 @@ firefox-ios:
     - frank@mozilla.com
   url: https://github.com/mozilla-mobile/firefox-ios
   metrics_files:
-    - ' Client/metrics.yaml'
+    - 'Client/metrics.yaml'
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -241,8 +241,8 @@ mozregression:
 firefox-ios:
   app_id: org-mozilla-ios-fennec
   notification_emails:
-    tlong@mozilla.com
-    frank@mozilla.com
+    - tlong@mozilla.com
+    - frank@mozilla.com
   url: https://github.com/mozilla-mobile/firefox-ios
   metrics_files:
     - ' Client/metrics.yaml'


fbertsch:
cc @travis79, there's no `pings.yaml` yet?

repositories.yaml
```diff
@@ -238,3 +238,11 @@ mozregression:
...
   ping_files:
     - 'mozregression/pings.yaml'
+firefox-ios:
+  app_id: org-mozilla-ios-fennec
+  notification_emails:
+    tlong@mozilla.com
+    frank@mozilla.com
+  url: https://github.com/mozilla-mobile/firefox-ios
+  metrics_files:
+    - 'Client/Generated/LegacyIds.swift'
```
mdboom, line 248: I think this should point to the metrics.yaml, not the generated code.

fbertsch: Indeed it definitely should... strange that I missed that 🤦‍♀️

travis79:
> cc @travis79, there's no `pings.yaml` yet?

No pings.yaml yet.

repositories.yaml
```diff
@@ -238,3 +238,11 @@ mozregression:
...
   ping_files:
     - 'mozregression/pings.yaml'
+firefox-ios:
+  app_id: org-mozilla-ios-fennec
+  notification_emails:
+    - tlong@mozilla.com
+    - frank@mozilla.com
+  url: https://github.com/mozilla-mobile/firefox-ios
+  metrics_files:
+    - ' Client/metrics.yaml'
```
mdboom, line 248:
```suggestion
    - 'Client/metrics.yaml'
```

fbertsch:
@travis79 sounds good, should be fine. Don't forget that when they add it, we need to add it here or those pings won't be available (though it should come up in our health meeting monitoring).

travis79:
FYI, @fbertsch I recently received updated information on the application ids used by Firefox for iOS:

Firefox Beta - org.mozilla.ios.FirefoxBeta
Firefox Release - org.mozilla.ios.Firefox
Development - org-mozilla-ios-fennec

I also mentioned this in [Bug 1637065](https://bugzilla.mozilla.org/show_bug.cgi?id=1637065)

fbertsch:
@travis79 what will Glean be sending in the pings for Beta and Release? `org-mozilla-ios-FirefoxBeta` and `org-mozilla-ios-Firefox`?

travis79:
> @travis79 what will Glean be sending in the pings for Beta and Release? `org-mozilla-ios-FirefoxBeta` and `org-mozilla-ios-Firefox`?

Correct, sorry I should have switched out for dashes when I commented earlier.

travis79:
I really wish that they were using channels rather than app ids to differentiate, but I guess we are dealing with this already with Fenix.

fbertsch:
@travis79 no reason we can't make that change for ios. I agree that these apps are the same, they should be flowing to the same tables. Any reason why we can't use the same app ids?

travis79:
It would be awkward since we don't allow applications to set this information (GleanSDK sets it automatically).  So we would need to allow the app to set this metric, or pass it in via the configuration.  I think that we didn't consider enough about how application ids are used in mobile or we would have made this a little less painful.

fbertsch:
I don't think it's out of the question to combine this data, in fact it's probably the right decision. We can even move Android to do the same thing.

Dexterp37:
> 
> 
> I really wish that they were using channels rather than app ids to differentiate, but I guess we are dealing with this already with Fenix.

This is a limitation of the stores (Google Play and Apple Store) and the way _they_ handle channels.
 
> I don't think it's out of the question to combine this data, in fact it's probably the right decision. We can even move Android to do the same thing.

@fbertsch I think that's a product decision and that *heavily* depends on products, so we can't really provide a "one solution for all". Given that, I'm not sure we should tackle this for this PR, specifically, or in Glean, in general, without a proposal :)

> It would be awkward since we don't allow applications to set this information (GleanSDK sets it automatically). So we would need to allow the app to set this metric, or pass it in via the configuration. I think that we didn't consider enough about how application ids are used in mobile or we would have made this a little less painful.

We [carefully considered](https://bugzilla.mozilla.org/show_bug.cgi?id=1520741) the implications of using app ids in the mobile world and not allowing them to be changed. That's also used as a defence mechanism (that seems to be working) against forks who mistakenly send telemetry.

travis79:
> > I really wish that they were using channels rather than app ids to differentiate, but I guess we are dealing with this already with Fenix.
> 
> This is a limitation of the stores (Google Play and Apple Store) and the way _they_ handle channels.
> 

I know, I just wish there was a better way to deal with it :/

> > It would be awkward since we don't allow applications to set this information (GleanSDK sets it automatically). So we would need to allow the app to set this metric, or pass it in via the configuration. I think that we didn't consider enough about how application ids are used in mobile or we would have made this a little less painful.
> 
> @travis79 we [carefully considered](https://bugzilla.mozilla.org/show_bug.cgi?id=1520741) the implications of using app ids in the mobile world and not allowing them to be changed. That's also used as a defence mechanism (that seems to be working) against forks who mistakenly send telemetry.

Maybe I should have reworded this to better indicate that this was just my frustration at how app ids work in the mobile world.  I know that we considered this carefully, I just wish that we could make better use of the app_channel.

jklukas:
> what will Glean be sending in the pings for Beta and Release? `org-mozilla-ios-FirefoxBeta` and `org-mozilla-ios-Firefox`?

Small nit - document namespaces are limited to `[a-z-]` so these should not include capital letters. I believe glean handles lowercasing, so I would expect these to be sent as `org-mozilla-ios-firefoxbeta` and `org-mozilla-ios-firefox` unless the Glean logic detects camelCase and breaks `firefoxbeta` into `firefox-beta`.

Dexterp37:
> 
> 
> > what will Glean be sending in the pings for Beta and Release? `org-mozilla-ios-FirefoxBeta` and `org-mozilla-ios-Firefox`?
> 
> Small nit - document namespaces are limited to `[a-z-]` so these should not include capital letters. I believe glean handles lowercasing, so I would expect these to be sent as `org-mozilla-ios-firefoxbeta` and `org-mozilla-ios-firefox` unless the Glean logic detects camelCase and breaks `firefoxbeta` into `firefox-beta`.

You are correct. The SDK [makes everything lowercase](https://searchfox.org/glean/rev/4e1882a2d0b6adb6ba72c943f4300ce435b9de96/glean-core/src/util.rs#201), so `org-mozilla-ios-firefoxbeta`  is the one to expect.

PR #193, opened 2020-06-04
Avoid ni? for disabled accounts and "see also" for confidential bugs

Ben-Wu:
Errors in bug creation happen if we needinfo an inactive account or if we try to see also a confidential bug 🤷‍♂️.  So instead of see also, put it in the description if confidential.  Could be simplified by always using description instead of see also but I think it's insignificant.

example bug https://bugzilla-dev.allizom.org/show_bug.cgi?id=1628814

### DIFF
--- a/probe_scraper/probe_expiry_alert.py
+++ b/probe_scraper/probe_expiry_alert.py
@@ -73,14 +73,15 @@ class ProbeDetails:
     product: str
     component: str
     emails: List[str]
-    previous_bug: Union[int, None]
+    # int will be put in the "see also" field, string will be in description due to permissions
+    previous_bug: Union[int, str, None]
 
 
 def bugzilla_request_header(api_key: str) -> Dict[str, str]:
     return {"X-BUGZILLA-API-KEY": api_key}
 
 
-def get_bug_component(bug_id: int, api_key: str) -> Tuple[str, str]:
+def get_bug_component(bug_id: int, api_key: str) -> Tuple[Union[str, None], Union[str, None]]:
     response = requests.get(BUGZILLA_BUG_URL + "/" + str(bug_id),
                             headers=bugzilla_request_header(api_key))
     try:
@@ -88,7 +89,7 @@ def get_bug_component(bug_id: int, api_key: str) -> Tuple[str, str]:
     except requests.exceptions.HTTPError as e:
         print(f"Error getting component for bug {bug_id}: {e}")
         if e.response.status_code == 401:  # Some confidential security bugs are not accessible
-            return BUG_DEFAULT_PRODUCT, BUG_DEFAULT_COMPONENT
+            return None, None
         else:
             raise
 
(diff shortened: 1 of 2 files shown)


probe_scraper/probe_expiry_alert.py
```diff
@@ -73,22 +73,23 @@ class ProbeDetails:
...
+    # int will be put in the "see also" field, string will be in description due to permissions
+    previous_bug: Union[int, str, None]
 
 
 def bugzilla_request_header(api_key: str) -> Dict[str, str]:
     return {"X-BUGZILLA-API-KEY": api_key}
 
 
-def get_bug_component(bug_id: int, api_key: str) -> Tuple[str, str]:
+def get_bug_component(bug_id: int, api_key: str) -> Tuple[Union[str, None], Union[str, None]]:
```
fbertsch, line 84: Python is really starting to feel like Scala.

probe_scraper/probe_expiry_alert.py
```diff
@@ -158,16 +159,28 @@ def create_bug(probes: List[ProbeDetails], version: str, api_key: str) -> int:
     probe_names = [probe.name for probe in probes]
     probe_prefix = get_longest_prefix(probe_names, tolerance=1)
 
+    see_also_bugs = list(set([probe.previous_bug for probe in probes
+                              if type(probe.previous_bug) == int]))
```
fbertsch, line 163: nit: `isinstance`

probe_scraper/probe_expiry_alert.py
```diff
@@ -189,7 +206,8 @@ def check_bugzilla_user_exists(email: str, api_key: str):
     user_response = requests.get(BUGZILLA_USER_URL + "?match=" + email,
                                  headers=bugzilla_request_header(api_key))
     user_response.raise_for_status()
-    return len(user_response.json()["users"]) > 0
+    users = user_response.json()["users"]
+    return len(users) > 0 and users[0]["can_login"]
```
fbertsch, line 210: What is `users[1]` can login?

probe_scraper/probe_expiry_alert.py
```diff
@@ -228,6 +238,10 @@ def find_expiring_probes(probes: dict, target_version: str,
             else:
                 last_bug_number = max(details["bug_numbers"])
                 product, component = (get_bug_component(last_bug_number, bugzilla_api_key))
+                if product is None:
```
fbertsch, line 241: Just for completeness, check that both product and component are `None`.

probe_scraper/probe_expiry_alert.py
Ben-Wu: `can_login` tells us if the account is disabled or not

probe_scraper/probe_expiry_alert.py
Ben-Wu: Oops I completely blanked when I was trying to remember that

probe_scraper/probe_expiry_alert.py
fbertsch: Right, but what if e.g. the second account is not disabled?

Ben-Wu: Oh I see what you mean.  Since this uses the `match` query param this can include multiple users like `abc@a.com` and `abc@a.com2`.  What should be done is to use `name` so that `len(users)` can only be 1 or 0.  We can raise an exception if not 0 or 1 to ensure this

PR #197, opened 2020-06-22
Fix yaml.load safety warning.

mreid-moz:
When running the probe scraper, I saw a message like:
  `/path/to/probe-scraper/probe_scraper/parsers/repositories.py:58: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.`

The recommended fix is to change a call like
  `yaml.load('myfile.yaml')`
to
  `yaml.load('myfile.yaml', Loader=yaml.FullLoader)`

So here we make this change for each `yaml.load` call in the tests, and use the `SafeLoader` in the main code.

### DIFF
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -55,7 +55,7 @@ class RepositoriesParser(object):
             filename = REPOSITORIES_FILENAME
 
         with open(filename, 'r') as f:
-            repos = yaml.load(f, Loader=yaml.FullLoader)
+            repos = yaml.load(f, Loader=yaml.SafeLoader)
 
         return repos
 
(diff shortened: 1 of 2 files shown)


probe_scraper/parsers/repositories.py
```diff
@@ -55,7 +55,7 @@ def _get_repos(self, filename=None):
             filename = REPOSITORIES_FILENAME
 
         with open(filename, 'r') as f:
-            repos = yaml.load(f)
+            repos = yaml.load(f, Loader=yaml.FullLoader)
```
fbertsch, line 58:
```suggestion
            repos = yaml.load(f, Loader=yaml.SafeLoader)
```

PR #202, opened 2020-07-13
Update application-services repo definitions to use branch `main`.

### LINKED ISSUE #200: Received email about expired metrics even after renewing them
Ref [this slack conversation](https://mozilla.slack.com/archives/CEE12R4E8/p1594000563222000).

We recently renewed some glean metrics for our "logins" component, thanks to the emails from probe-scraper about their approaching expiry. The pull request was at https://github.com/mozilla/application-services/pull/3290 and was merged five days ago.

However, I got another email yesterday from probe-scraper reminding me the metrics were about to expire, despite the PR above having already been merged. I don't *think* there's anything else we need to do to update the metrics, so maybe this is a bug in the email-notification logic?

rfk:
We recently switch from `master` to `main` as the default branch for the `application-services` repo, this updates the probe-scraper repository definitions to look in the right place. I expect this will fix #200.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -164,10 +164,11 @@ logins-store:
   notification_emails:
     - rfkelly@mozilla.com
     - lina@mozilla.com
-    - application-services@mozilla.com
+    - sync-team@mozilla.com
   url: 'https://github.com/mozilla/application-services'
   metrics_files:
     - 'components/logins/android/metrics.yaml'
+  branch: main
   library_names:
     - org.mozilla.appservices:logins
 support-migration:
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -164,10 +164,11 @@ logins-store:
   notification_emails:
     - rfkelly@mozilla.com
     - lina@mozilla.com
-    - application-services@mozilla.com
+    - sync-team@mozilla.com
```
rfk, line 167: We don't really use the `application-services@` email address any more, so I took the liberty of updating it while I was in here.

repositories.yaml
```diff
@@ -282,9 +283,11 @@ android-places:
   app_id: android-places
   notification_emails:
     - frank@mozilla.com
+    - sync-team@mozilla.com
```
rfk, line 286: And took the liberty of sharing the notification email load with poor frank, since the sync team is ultimately responsible for these metrics :-)

fbertsch:
For future reviewers of PRs like these, when we change to a different branch we risk losing a history of probes. The most important thing is to ensure that the new list of metrics is the same as the existing metrics in `probeinfo.telemetry.mozilla.org/glean/{repo_name}/metrics`. If any are dropped from there, it will result in an incompatible schema change in BQ.

PR #203, opened 2020-07-20
bug 1650782 - Reroute emails previously sent to dev-telemetry-alerts to glean-team

chutten:
dev-telemetry-alerts is to be decommissioned.


Dexterp37:
@chutten I'm not sure this is a good idea :) "glean-team" is really meant for internal comms and might not be the best audience for this kind of email. Isn't fx-team or dev-platform best suited?

chutten:
I don't think either of those mailing lists' admins will consent to receive automated email. They are both too large.

The only viable alternative I came up with in my investigation was to stop sending it to a default address at all. That's a bigger change in behaviour and code, so I went with the simpler course.

Dexterp37:
Ok ok... !

chutten:
I just checked and you are indeed on glean-team@

PR #211, opened 2020-07-31
Issue#52

### LINKED ISSUE #52: Deal with EOL Whitespace in outputted JSON
According to the [`json.dump` docs](https://docs.python.org/3.5/library/json.html#json.dump), the default `separators` value for Python 2 is `(', ', ': ')`, even with a non-default `indent` value.

This means before the `\n` at the end of lines with terminal commas there is a space. EOL Whitespace! Sound the alarm!

This was fixed in Python v3.4, but as the docs recommend and the current production job use `python` instead of `python3` this means we need to specify `separators=(',',': ')` in `json.dump`

jhabarsingh:
Fixes #52

### DIFF
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -1,6 +1,8 @@
 from probe_scraper import runner
 from copy import deepcopy
 from datetime import datetime
+import os
+import shutil
 
 
 def test_add_first_appeared_dates():
@@ -69,3 +71,31 @@ def test_add_first_appeared_dates():
     }
 
     assert runner.add_first_appeared_dates(probes_by_channel, first_appeared_dates) == expected
+
+
+def test_trailing_space(tmp_path):
+    '''Test cases to check the output of json.dumps has no trailing spaces
+    '''
+    test_cases = [
+        {
+            "test1": 12,
+            "test2": 31,
+        }
+    ]
+
+    DIR_NAME = tmp_path / 'output_dir'
+    FILE_NAME = 'file_name.txt'
+
+    for test_case in test_cases:
+        trailing_spaces = 0     # Counts no of traling spaces
+        runner.dump_json(test_case, DIR_NAME, FILE_NAME)
+        path = os.path.join(DIR_NAME, FILE_NAME)
+        with open(path, 'r') as file:
+            for line in file.readline():
+                if line[-1] == ' ':
+                    trailing_spaces += 1
+
+        assert not trailing_spaces
+
+    # Deleting the files created [ TEAR-DOWN ]
+    shutil.rmtree(DIR_NAME)
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -2,7 +2,6 @@ from probe_scraper import runner
 from copy import deepcopy
 from datetime import datetime
 import os
-import shutil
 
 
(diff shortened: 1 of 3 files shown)


jhabarsingh:
okay sir, can you tell me the file name where I have to write tests .

jhabarsingh:
```
{
   "test1": 1,
   "test2": 2,
   "test3": 3
}
```
this is the output it will give.
it's giving space after colon, is that okay or not ?

I wanted to confirm it once before writing the tests.

mdboom:
You can put the test here https://github.com/mozilla/probe-scraper/blob/master/tests/test_runner.py

jhabarsingh:
> ```
> {
>    "test1": 1,
>    "test2": 2,
>    "test3": 3
> }
> ```
> 
> this is the output it will give.
> it's giving space after colon, is that okay or not ?
> 
> I wanted to confirm it once before writing the tests.

jhabarsingh:
Sir I have added tests to check trailing spaces in the output of  json.dump(), can you review it and let me know if any changes needed

tests/test_runner.py
```diff
@@ -69,3 +71,35 @@ def test_add_first_appeared_dates():
     }
 
     assert runner.add_first_appeared_dates(probes_by_channel, first_appeared_dates) == expected
+
+
+def test_trailing_space():
```
mdboom, line 76: This test dumps files in the current directory, which is generally not a good idea.  Let's use the [tmp_path](https://docs.pytest.org/en/stable/tmpdir.html#the-tmp-path-fixture) fixture here instead.

tests/test_runner.py
```diff
@@ -69,3 +71,35 @@ def test_add_first_appeared_dates():
...
+    '''
+    test_cases = [
+        {
+            "test1": 12,
+            "test2": 31,
+        },
+        {
+            "test1": 12,
+            "test2": 31
+        },
```
mdboom, line 87: Both of these test cases are the same.  I think it's fine to just have one test case.

jhabarsingh: yeah I will remove this portion

tests/test_runner.py
jhabarsingh: Okay sir I will use tmp_path intead

jhabarsingh:
@mdboom,  I have had the changes you told me can you review it once ?

tests/test_runner.py
```diff
@@ -69,3 +71,31 @@ def test_add_first_appeared_dates():
...
+        path = os.path.join(DIR_NAME, FILE_NAME)
+        with open(path, 'r') as file:
+            for line in file.readline():
+                if line[-1] == ' ':
+                    trailing_spaces += 1
+
+        assert not trailing_spaces
+
+    # Deleting the files created [ TEAR-DOWN ]
+    shutil.rmtree(DIR_NAME)
```
mdboom, line 101: This isn't necessary if using the `tmp_path` fixture.

jhabarsingh:
I have removed the  lines that were using shutil.

jhabarsingh:
Sir, this was my first contribution to the Open source and I am glad you were there to help me. Thanks a lot for being so helpful.

Issue #185, opened 2020-04-09
Separate Android metrics from platform-independent metrics in Glean repositories.yaml


mdboom:
Currently the "glean" dependency as specified in the repositories.yaml includes metrics.yaml files both for Android and that apply to all platforms.

We should separate this out into two dependencies for Android and platform-independent, and the adjust all of the applications in the repositories.yaml to use the appropriate dependencies.

fbertsch:
The plan is to have two libraries:
- `glean`, which will have just the core glean metrics
- `glean-android`, which will have the core glean metrics as well as the android-specific metrics

`glean` can be added to all non-Android apps, while `glean-android` replaces the existing `glean` library.

@mdboom @Dexterp37 which color should our bikeshed be, `glean` or `glean-core`?

Dexterp37:
> 
> 
> The plan is to have two libraries:
> 
>     * `glean`, which will have just the core glean metrics
> 
>     * `glean-android`, which will have the core glean metrics as well as the android-specific metrics
> 
> 
> `glean` can be added to all non-Android apps, while `glean-android` replaces the existing `glean` library.
> 
> @mdboom @Dexterp37 which color should our bikeshed be, `glean` or `glean-core`?

I'd favour `glean-core` for consistency with our SDK naming :)

fbertsch:
> I'd favour `glean-core` for consistency with our SDK naming :)

Sold!

The next question is about pings. Are the four default pings what we want for `glean-core`? Do those need to be split into separate files? (are non-mobile apps sending `baseline` pings, for example?) @Dexterp37

Dexterp37:
> 
> 
> > I'd favour `glean-core` for consistency with our SDK naming :)
> 
> Sold!
> 
> The next question is about pings. Are the four default pings what we want for `glean-core`? Do those need to be split into separate files? (are non-mobile apps sending `baseline` pings, for example?) @Dexterp37

Yes, non-mobile apps can send `baseline` pings (at least with reason `dirty_startup`). There are no plans right now to split the pings in core vs non-core.

fbertsch:
Fixed with https://github.com/mozilla/probe-scraper/pull/217.

PR #217, opened 2020-08-08
Bug 1656649 - Add Fx Reality PC

fbertsch:
Is Glean a dependency for the C# apps?

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -25,24 +25,6 @@ glean-android:
     - 'glean-core/pings.yaml'
   library_names:
     - org.mozilla.components:service-glean
-fenix:
-  app_id: org-mozilla-fenix
-  notification_emails:
-    - frank@mozilla.com
-    - mdroettboom@mozilla.com
-  url: 'https://github.com/mozilla-mobile/fenix'
-  metrics_files:
-    - 'app/metrics.yaml'
-  ping_files:
-    - 'app/pings.yaml'
-  dependencies:
-    - org.mozilla.components:service-glean
-    - org.mozilla.components:lib-crash
-    - org.mozilla.components:support-sync-telemetry
-    - org.mozilla.components:browser-engine-gecko-beta
-    - org.mozilla.appservices:logins
-    - org.mozilla.components:support-migration
-    - org.mozilla.components:places
 glean-deprecated:
   app_id: glean
   notification_emails:
@@ -73,7 +55,6 @@ fenix:
     - org.mozilla.appservices:logins
     - org.mozilla.components:support-migration
     - org.mozilla.components:places
-
 reference-browser:
   app_id: org-mozilla-reference-browser
   notification_emails:
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -10,7 +10,7 @@ glean-core:
   ping_files:
     - 'glean-core/pings.yaml'
   library_names:
-    - org.mozilla.glean-core
+    - glean-core
 glean-android:
   app_id: glean-android
   notification_emails:
@@ -359,4 +359,4 @@ firefox-reality-pc:
   ping_files:
     - 'Source/FirefoxRealityUnity/pings.yaml'
   dependencies:
-    - org.mozilla.glean-core
+    - glean-core
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -324,3 +324,12 @@ firefox-for-echo-show:
     - 'app/metrics.yaml'
   dependencies:
     - org.mozilla.components:service-glean
+firefox-reality-pc:
+  app_id: org-mozilla-firefoxreality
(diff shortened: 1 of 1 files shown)


Dexterp37:
> 
> 
> Is Glean a dependency for the C# apps?

Yes, Glean is, but *not* Glean Android. Should we split [this](https://github.com/mozilla/probe-scraper/blob/5c9d656fb42c5d9f02cf85d26d156f221c9e329e/repositories.yaml#L1) in two? `glean` and `glean_android`. With `glean_android` having a dependency on `glean`? @mdboom @fbertsch ?

fbertsch:
@Dexterp37 it sounds like that is the correct way to approach this, however all dependencies have to be explicitly defined, so we would need to add both `glean` and `glean_android` for android apps.

Dexterp37:
> 
> 
> @Dexterp37 it sounds like that is the correct way to approach this, however all dependencies have to be explicitly defined, so we would need to add both `glean` and `glean_android` for android apps.

I see. Maybe `glean_core` and `glean_android` then, but that's indeed the way forward.

mdboom:
> @Dexterp37 it sounds like that is the correct way to approach this, however all dependencies have to be explicitly defined, so we would need to add both `glean` and `glean_android` for android apps.

It will be hard to remember, but I think this is the correct approach.

fbertsch:
Alternatively, we could have `glean-android` include both those metrics files. Then we would only need to include the one dependency for Android apps. Thoughts?

Dexterp37:
> 
> 
> Alternatively, we could have `glean-android` include both those metrics files. Then we would only need to include the one dependency for Android apps. Thoughts?

Oh, that's even better :) We'll also need a comment section at the top explaining why. With that, I think we should be ok.

fbertsch:
Great. I'll add that to this PR.

fbertsch:
@Dexterp37 @mdboom follow-up, will Glean then also be a dependency for the Python apps?

mdboom:
Yeah -- I'm wondering a bit why it's not already a dependency for iOS and Python apps. This would mainly impact the error metrics.  Looking at those schemas/tables now to confirm/deny my worst fears...

Dexterp37:
> 
> 
> @Dexterp37 @mdboom follow-up, will Glean then also be a dependency for the Python apps?

Yes.

mdboom:
> Yeah -- I'm wondering a bit why it's not already a dependency for iOS and Python apps. This would mainly impact the error metrics. Looking at those schemas/tables now to confirm/deny my worst fears...

So -- somehow the Glean core metrics are ending up in the tables for the iOS apps, even though it's not a dependency there. We should at least remind ourselves and understand why that is first.

fbertsch:
> So -- somehow the Glean core metrics are ending up in the tables for the iOS apps, even though it's not a dependency there. We should at least remind ourselves and understand why that is first.

@mdboom is there a bug or issue filed for this? This seems important to track down.

I'll summarize this in https://github.com/mozilla/probe-scraper/issues/185 and assign that to myself.

fbertsch:
Now fixes #185.

repositories.yaml
```diff
@@ -1,5 +1,18 @@
...
+    - frank@mozilla.com
+    - mdroettboom@mozilla.com
+  url: 'https://github.com/mozilla/glean'
+  branch: main
+  metrics_files:
+    - 'glean-core/metrics.yaml'
+  ping_files:
+    - 'glean-core/pings.yaml'
+  library_names:
+    - org.mozilla.glean-core
```
Dexterp37, line 13: nit: is this required to be `org.mozilla.glean-core` (android specific?)? Can this be `glean-core` or `lib-glean-core`? We don't need to change it, just curious.

repositories.yaml
```diff
@@ -12,6 +25,18 @@ glean:
     - 'glean-core/pings.yaml'
   library_names:
     - org.mozilla.components:service-glean
+glean-deprecated:
```
Dexterp37, line 28: Why are we leaving this around?

repositories.yaml
fbertsch: Nope, it can be any valid string, really. Would you prefer `glean-core`?

repositories.yaml
fbertsch: There's some description in this commit: https://github.com/mozilla/probe-scraper/pull/217/commits/1db46bea9dca73a55c384b10b25536cb43203af6

repositories.yaml
Dexterp37: Slight preference for that, yes, which would give this a less Android-specific flavour.

repositories.yaml
Dexterp37: Oh, shame on me for not reading the commit message. Sorry!

Dexterp37 (approved):

PR #218, opened 2020-08-10
Add main as an alternative default branch

fbertsch:
We just ran into a failure yesterday on https://github.com/MozillaReality/FirefoxReality, which just moved to `main` from `master`.

This adds `main` as the fallback; once everyone is switched over we can make `master` the fallback (and hopefully deprecate it).


fbertsch:
That's a good point - that error case should be handled by adding the branch directly. If `main` was branched from `master`, then we should be good schema-compatible-wise when that happens; when they have disparate histories is when we run into trouble.

We can perhaps just _hope_ that doesn't happen.

PR #220, opened 2020-09-01
Add descriptions to repositories

wlach:
We currently have no easy way of understanding what application
an individual repository corresponds to. For the purposes of the
Glean dictionary and other tools, we really should. Even
internally inside the probe scraper, having this metadata makes
it easier to understand what data we are collecting for auditing
purposes.

/cc @robhudson

This PR:

* Add descriptions to all repositories
* Makes the description field mandatory on all repository
* (Attempts to) incorporate a description field to the repositories generated

Still todo: 

- [x] Refine the description fields (this is an initial attempt based on scraping the repos on Github and is incomplete)
- [x] Validate that the generated repository json contains the repository fields (https://probeinfo.telemetry.mozilla.org/glean/repositories)


repositories.yaml
```diff
@@ -1,5 +1,6 @@
 glean:
   app_id: glean
+  description: Modern Firefox Telemetry for mobile platforms
```
wlach, line 3: https://github.com/mozilla/glean/ repository description might need updating 😄

repositories.yaml
```diff
@@ -43,6 +46,7 @@ reference-browser:
     - org.mozilla.components:lib-crash
 fenix-nightly:
   app_id: org-mozilla-fenix-nightly
+  description: Firefox for Android
```
wlach, line 49: For these types of cases, where one repo corresponds to multiple apps, what do we do? We could probably just do something like:

```suggestion
  description: Firefox for Android (nightly version)
```

While we're here it would be nice to sort out which versions of Fenix and other products are still valid. I might want to explicitly tag deprecated stuff in the future, but at least updating the description would be a start.

repositories.yaml
```diff
@@ -69,6 +74,7 @@ firefox-for-fire-tv:
     - org.mozilla.components:service-glean
 lib-crash:
   app_id: lib-crash
+  description: A collection of Android libraries to build browsers or browser-like applications
```
wlach, line 77: It would be nice to give these types of component repositories sensible names.

mdboom: Each Android component has its own README and tagline, for example for this one: https://github.com/mozilla-mobile/android-components/tree/master/components/lib/crash

repositories.yaml
wlach: Yeah. On second thought, what do you think about adding an (optional) deprecated field right away? It would be nice to have an authoritative way of marking a repository as deprecated.

repositories.yaml
wlach: Awesome! Integrated those.

repositories.yaml
wlach: Done

repositories.yaml
wlach: And actually, looking through it again-- how about an (optional) channel field?

fbertsch: +1 on channel. It fits our deploy model, no reason to exclude it because it's specific.

fbertsch (approved):
This is a good addition, thanks @wlach. A nice followup here would be to include this in the dataset description in BQ.

Related to the channels piece, @mreid-moz and I have been discussing having a `product-family` entry here which would indicate that the apps are related (e.g. Fenix, Firefox iOS).

wlach:
> This is a good addition, thanks @wlach. A nice followup here would be to include this in the dataset description in BQ.

That would be cool, do you have any pointer on where that would live? I'm not sure what component/code actually generates/updates the bq datasets.

> Related to the channels piece, @mreid-moz and I have been discussing having a `product-family` entry here which would indicate that the apps are related (e.g. Fenix, Firefox iOS).

Could be useful, but let's save for that a follow-up. YAGNI and all that

repositories.yaml
wlach: Ok, I'll go ahead and add channel and deprecated to the schema and set it where appropriate.

fbertsch:
Oh, one more follow-up: let's update the bug filing process to include this description

https://github.com/mozilla/glean/blob/main/docs/user/adding-glean-to-your-project.md#adding-metadata-about-your-project-to-the-pipeline

fbertsch:
> That would be cool, do you have any pointer on where that would live? I'm not sure what component/code actually generates/updates the bq datasets.

We would need to confer with @whd here, ops code reads the generated schemas and decides if new datasets need to be created.

schemas/repositories.json
```diff
@@ -9,6 +9,17 @@
         "type": "string",
         "pattern": "^[a-z][a-z-]{0,1023}$"
       },
+      "description": {
+        "type": "string"
+      },
+      "channel": {
+        "type": "string",
+        "enum": ["release", "beta", "nightly", "esr"]
```
wlach, line 17: Opting to be very strict about this right now (corresponding to: https://docs.telemetry.mozilla.org/concepts/channels/channel_normalization.html) since it covers the existing set of possible channels. We can always update this in the future if required, but this will hopefully act as a nudge to use consistent terminology across Mozilla, which I think is desirable.

wlach:
> Oh, one more follow-up: let's update the bug filing process to include this description
> 
> https://github.com/mozilla/glean/blob/main/docs/user/adding-glean-to-your-project.md#adding-metadata-about-your-project-to-the-pipeline

See mozilla/glean#1199

wlach:
Ready for review I think.

fbertsch (approved):
🚢

mdboom (approved):

PR #224, opened 2020-09-09
Clean up yaml and add a yaml-lint check

wlach:
* Add yamllint check and fix various minor issues it found
* Remove unneeded quotes from repositories.yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ clean-pyc:
 
 lint: build
 	docker-compose run app flake8 --max-line-length 100 .
-	docker-compose run app yamllint --max-line-length 100 repositories.yaml .circleci
+	docker-compose run app yamllint repositories.yaml .circleci
 
 test: build
 	docker-compose run app pytest tests/ --run-web-tests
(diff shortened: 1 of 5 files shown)


fbertsch (changes requested):
@wlach did you mean to add this to CI? It would be very useful to ensure that the files have been linted on every commit.

wlach:
> @wlach did you mean to add this to CI? It would be very useful to ensure that the files have been linted on every commit.

Oops yes! This was supposed to be part of the pull request but it somehow got lost. Added it back.

Makefile
```diff
@@ -24,6 +24,7 @@ clean-pyc:
 
 lint: build
 	docker-compose run app flake8 --max-line-length 100 .
+	docker-compose run app yamllint repositories.yaml .circleci
```
fbertsch, line 27: Do we need to add this as a dependency?

fbertsch: I didn't know that. Probably not a big deal either way then

fbertsch (approved):
LGTM, thank you! We should do this to bigquery-etl tests and metadata as well

Makefile
wlach: It's easy to add, so I added it. Will save us if glean ever loses the dep.

PR #227, opened 2020-09-14
Add a temporary app id for testing glean.js

mdboom:
During the Glean.js workweek (and probably subsequent weeks for early
development), it would be useful to have a real table to send data to.

Eventually, we will delete this and its associated table (which I assume is
possible).

This will be *very* low volume.


fbertsch (approved):
Very exciting!

PR #228, opened 2020-09-14
Use probe data from all buildhub revisions instead of release tags

### LINKED ISSUE #225: P-S not properly parsing difference in `record_in_processes`
See https://github.com/mozilla/mozilla-schema-generator/issues/147

[Here](https://phabricator.services.mozilla.com/D78033) is where this field was added, and [here](https://phabricator.services.mozilla.com/D89401) is where it changed. The probe-info-service should have an entry for each of those, but instead it only has an entry for the latter.

Ben-Wu:
(fixes #225)

~~Seems to work from some quick tests so putting a draft up first.  I'll add unit tests and test more before marking as ready for review.~~ 👍

This is the output for the probe in question: https://gist.github.com/Ben-Wu/76f8eacf2da9afff7e2d9e80ea7ad3fb

### DIFF
--- a/probe_scraper/scrapers/moz_central_scraper.py
+++ b/probe_scraper/scrapers/moz_central_scraper.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import datetime
 import re
 import os
 import json
@@ -132,7 +133,7 @@ def adjust_version(channel, version):
 
 
 def extract_tag_data(tag_data, channel, min_fx_version, max_fx_version):
-    tag_regex = CHANNELS[channel]['tag_regex']
+    tag_regex = CHANNELS[channel]["tag_regex"]
     tip_node_id = tag_data["node"]
     tags = [t for t in tag_data["tags"] if re.match(tag_regex, t["tag"])]
     results = []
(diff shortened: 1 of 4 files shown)


Ben-Wu:
Additional caveat is that the buildhub revisions are per push so the revisions it has are for the last change at the time of each push instead of every single change.  So if there are two patches in a single push that affects a probe, it will be reflected as one change.  This shouldn't matter at all, just some additional context to reason about things.

probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -272,7 +276,7 @@ def scrape(folder=None, min_fx_version=None, max_fx_version=None, channels=None)
         folder = tempfile.mkdtemp()
 
     error_cache = load_error_cache(folder)
-    requests_cache.install_cache('probe_scraper_cache')
+    requests_cache.install_cache("probe_scraper_cache")
     results = defaultdict(dict)
 
     if channels is None:
```
fbertsch, line 282: Looks to me like this is still scraping all the channels. We should explicitly only scrape release.

fbertsch (approved):
Nice, this was much simpler than I anticipated. r+wc

probe_scraper/scrapers/moz_central_scraper.py
fbertsch: My mistake, keep as-is

Ben-Wu:
Created https://github.com/mozilla/probe-scraper/issues/230 for using revisions from buildhub

PR #236, opened 2020-09-16
Use buildhub results for revision output

### LINKED ISSUE #235: Probes have revisions that aren't in the revisions output
With the recent change that takes all revisions from buildhub instead of https://hg.mozilla.org/mozilla-central/json-tags (https://github.com/mozilla/probe-scraper/pull/228), probes now have revisions that aren't included in the revision output https://probeinfo.telemetry.mozilla.org/firefox/revisions

This affect probe dictionary.  We should instead include all the buildhub revisions in the output.

Ben-Wu:
(fixes #235)


probe_scraper/scrapers/moz_central_scraper.py
```diff
@@ -251,56 +249,6 @@ def save_error_cache(folder, error_cache):
...
-    error_cache = load_error_cache(folder)
-    requests_cache.install_cache("probe_scraper_cache")
-    results = defaultdict(dict)
-
-    if channels is None:
-        channels = CHANNELS.keys()
-
-    for channel in channels:
-        tags = load_tags(channel)
-        versions = extract_tag_data(tags, channel, min_fx_version, max_fx_version)
```
fbertsch, line 287: Good news! We can remove _everything_ related to using json-tags above.

fbertsch (approved):
r+wc, let's strip out all that cruft we don't need anymore!

probe_scraper/scrapers/moz_central_scraper.py
Ben-Wu: Oops my code inspection didn't highlight the functions as unused

PR #242, opened 2020-10-07
Upgrade to glean_parser v1.29.0

mdboom:
* **Breaking change:** `glean_parser` will now return an error code when any of the input files do not exist (unless the `--allow-missing-files` flag is passed).
* Generated code now includes a comment next to each metric containing the name of the metric in its original `snake_case` form.
* When metrics don't provide a `unit` parameter, it is not included in the output (as provided by probe-scraper).

### DIFF
--- a/probe_scraper/parsers/pings.py
+++ b/probe_scraper/parsers/pings.py
@@ -27,6 +27,7 @@ class GleanPingsParser:
     def parse(self, filenames, config):
         config = config.copy()
         paths = [Path(fname) for fname in filenames]
+        paths = [path for path in filenames if path.is_file()]
         results = parse_objects(paths, config)
         errors = [err for err in results]
 
--- a/probe_scraper/parsers/pings.py
+++ b/probe_scraper/parsers/pings.py
@@ -27,7 +27,7 @@ class GleanPingsParser:
     def parse(self, filenames, config):
         config = config.copy()
         paths = [Path(fname) for fname in filenames]
-        paths = [path for path in filenames if path.is_file()]
+        paths = [path for path in paths if path.is_file()]
         results = parse_objects(paths, config)
         errors = [err for err in results]
 
--- a/probe_scraper/parsers/metrics.py
+++ b/probe_scraper/parsers/metrics.py
@@ -20,7 +20,7 @@ class GleanMetricsParser:
         config["do_not_disable_expired"] = True
 
         paths = [Path(fname) for fname in filenames]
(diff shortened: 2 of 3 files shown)


probe_scraper/parsers/pings.py
```diff
@@ -27,6 +27,7 @@ class GleanPingsParser:
     def parse(self, filenames, config):
         config = config.copy()
         paths = [Path(fname) for fname in filenames]
+        paths = [path for path in filenames if path.is_file()]
```
fbertsch, line 30: Can we bump to 3.8? Then we can use an assignment expression: `[path for fname in filenames if (path := Path(fname)).is_file()]`

fbertsch (commented):
> Generated code now includes a comment next to each metric containing the name of the metric in its original snake_case form.

Is Glean altering the names of existing metrics? Will this result in duplicates? I think not since we snake-case for col names, but I want to verify.

probe_scraper/parsers/pings.py
mdboom: > > Generated code now includes a comment next to each metric containing the name of the metric in its original snake_case form.
> 
> Is Glean altering the names of existing metrics? Will this result in duplicates? I think not since we snake-case for col names, but I want to verify.

When we generate code for language bindings (like Kotlin), we respell the case so it fits into that language's conventions.  This change is just about adding a comment next to that generated code with the original snake_case name from the metrics.yaml file, to make it easier to find in searchfox, grep, etc.

(So should have no bearing on probe_scraper -- sorry, I just blindly copied the whole changelog).

fbertsch: Hah, no problem. My ears immediately perk up when I hear "original name" w.r.t. probe-scraper.

fbertsch (approved):
No reason to block this on bumping to 3.8. Offhand I don't know if that could result in breaking changes and crazy tests.

probe_scraper/parsers/pings.py
mdboom: > Can we bump to 3.8? Then we can use an assignment expression: `[path for fname in filenames if (path := Path(fname)).is_file()]`

glean_parser supports 3.8, so we can bump probe_scraper to require 3.8 without issue.  Might be best as a separate issue, though, in case that has any bearing on the infrastructure probe_scraper is running on?

fbertsch: That sounds fine. This is all dockerized for running though, so a local test should verify whether it would work.

Issue #243, opened 2020-10-13
Linting improvements


wlach:
We already have some amount of linting in this repository (with flake8), however it would be good to also add [black](https://github.com/psf/black) and [isort](https://pypi.org/project/isort/) to the list and have circleCI check for compliance.

Instructions on how to set up the repository are here: https://github.com/mozilla/probe-scraper#developing-the-probe-scraper

Linting is currently done via the docker container here: https://github.com/mozilla/probe-scraper/blob/master/Makefile

For reference, iodide has a similar linting setup as what we'd want here and it also uses docker: https://github.com/iodide-project/iodide

wlach:
@LilyLME is going to take this one!

PR #246, opened 2020-10-13
added black and isort linters

LilyLME:
Added black and isort linters to implement more vigorous checking of python.
Fixes : #243 
### Notes
* CircleCI workflow left unchanged since `the lint` step is already included.

```
   steps:
      - checkout
      - run: make build
      - run: make lint
      - run: make test
```
* `black` and `isort` dependencies were installed and simply run in the lint section of the `Makefile`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,8 @@ clean-pyc:
 lint: build
 	docker-compose run app flake8 --max-line-length 100 .
 	docker-compose run app yamllint repositories.yaml .circleci
+	docker-compose run app python -m black .
+	docker-compose run app python -m isort .
 
 test: build
 	docker-compose run app pytest tests/ --run-web-tests
--- a/Makefile
+++ b/Makefile
@@ -24,14 +24,14 @@ clean-pyc:
 	find . -name '*~' -exec rm -f {} +
 
 format: 
-	python3 -m black probe_scraper tests setup.py
-	python3 -m isort probe_scraper tests setup.py
+	python3 -m black probe_scraper tests ./*.py
+	python3 -m isort probe_scraper tests ./*.py
 
 lint: build
 	docker-compose run app flake8 --max-line-length 100 .
 	docker-compose run app yamllint repositories.yaml .circleci
-	docker-compose run app python -m black --check probe_scraper tests setup.py
-	docker-compose run app python -m isort --check-only probe_scraper tests setup.py
+	docker-compose run app python -m black --check probe_scraper tests ./*.py
+	docker-compose run app python -m isort --check-only probe_scraper tests ./*.py
 
 test: build
 	docker-compose run app pytest tests/ --run-web-tests
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 help:
 	@echo "  clean                  Remove build artifacts"
 	@echo "  lint                   Check style with flake8"
+	@echo "  format                 Format code with black and isort"
 	@echo "  test                   Run tests quickly with the default Python"
 	@echo "  build                  Builds the docker images for the docker-compose setup"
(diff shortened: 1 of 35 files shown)


Makefile
```diff
@@ -25,6 +25,8 @@ clean-pyc:
 lint: build
 	docker-compose run app flake8 --max-line-length 100 .
 	docker-compose run app yamllint repositories.yaml .circleci
+	docker-compose run app python -m black .
```
acmiyaguchi, line 28: Linting should check for potential errors, instead of actually formatting. There's a `black` command-line flag that will do this, but the behavior here will be to format the files instead.

Makefile
```diff
@@ -25,6 +25,8 @@ clean-pyc:
...
+	docker-compose run app python -m isort .
```
acmiyaguchi, line 29: Ditto, this will format sort imports instead of just checking.

acmiyaguchi (changes requested):
This won't fail CI as expected -- it will apply formatting and return with a success.

It might be useful to also add a `make format` command that you can that runs `black` and `isort`.

Makefile
LilyLME: Yeah, that's true. Let me fix that up fast.

LilyLME: @acmiyaguchi I thought of that but actually wondered if I would have to format the codebase as well before sending this PR.

All other PRs would start failing unless a formatted version of the code is merged.

acmiyaguchi: Yes, the whole code-base would need to be formatted and checked in.

@fbertsch what are your thoughts on formatting the codebase with `black`?

fbertsch: Please do! Keep the reformatting as it's own separate commit.

LilyLME:
@acmiyaguchi you want to take a look again now fails gracefully ;) 
 * https://circleci.com/gh/mozilla/probe-scraper/446?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Thank you.

acmiyaguchi (commented):
Would you mind adding in a `make format` command and formatting all of the files so that CI passes when this gets merged into master?

LilyLME:
right away then.

Makefile
```diff
@@ -22,6 +23,10 @@ clean-pyc:
 	find . -name '*.pyo' -exec rm -f {} +
 	find . -name '*~' -exec rm -f {} +
 
+format: 
+	python3 -m black .
```
acmiyaguchi, line 27: One last thought. It's common to put a virtual environment folder `venv/` at the root of the repository, which contains a lot of internal python files. Is it possible to ignore these folders, since they may contain thousands of python files?

https://github.com/mozilla/probe-scraper/blob/022eb2d102ffcdd726e012dfd3d2a2dc2374b776/.gitignore#L81-L83

If you wanted to test this, run `python3 -m venv; source venv/bin/activate;' pip install -r requirements.txt`, then run `make format` and check what's getting formatted.

Makefile
```diff
@@ -22,9 +23,15 @@ clean-pyc:
...
 	find . -name '*~' -exec rm -f {} +
 
+format: 
+	python3 -m black probe_scraper tests setup.py
+	python3 -m isort probe_scraper tests setup.py
+
 lint: build
 	docker-compose run app flake8 --max-line-length 100 .
 	docker-compose run app yamllint repositories.yaml .circleci
+	docker-compose run app python -m black --check probe_scraper tests setup.py
```
acmiyaguchi, line 33: One last file to check is the `conftest.py` at the root of the project (which should be moved to tests/, but that's unrelated to this PR).

acmiyaguchi (approved):
Thanks for taking time to work on this!

Issue #234, opened 2020-09-16
Validate that repositories exist and yaml files are scrapable


wlach:
At minimum our continuous integration should verify that all repositories listed in https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml are reachable (from an unauthenticated session) and that the files contained therein are reachable. 

For the most part, we can probably do this validation with a simple python script using the requests library by hitting github endpoints (e.g. https://raw.githubusercontent.com/mozilla/mozregression/master/mozregression/metrics.yaml)

edward-hong:
Sorry I'm not as experienced in Python as I am with Javascript and Svelte.
Can you please assign me a frontend task instead?

LilyLME:
@wlach I would be happy to take this since @edward-hong does not want it.

wlach:
> @wlach I would be happy to take this since @edward-hong does not want it.

Sure! This would be a good followup to what you were working on before.

PR #249, opened 2020-10-15
Added to check to validate scrappable repositories.

LilyLME:
As required in #234 this PR addresses the need to ensure that all repositories in [https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml](https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml) are reachable

Fixes : #234 

## Notes
 - The script has correct detected unreachable repositories
 - This PR may address unreachable repositories in a commit if decided as such.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 
 help:
 	@echo "  clean                  Remove build artifacts"
+	@echo "  check-repos            Verify all repositories in repositories.yaml are scrapable"
 	@echo "  lint                   Check style with flake8"
 	@echo "  format                 Format code with black and isort"
 	@echo "  test                   Run tests quickly with the default Python"
@@ -32,6 +33,9 @@ lint: build
 	docker-compose run app yamllint repositories.yaml .circleci
 	docker-compose run app python -m black --check probe_scraper tests ./*.py
 	docker-compose run app python -m isort --check-only probe_scraper tests ./*.py
+
+check-repos: 
+	docker-compose run app python probe_scraper/check_repositories.py
 
 test: build
 	docker-compose run app pytest tests/ --run-web-tests
--- a/probe_scraper/check_repositories.py
+++ b/probe_scraper/check_repositories.py
@@ -1,0 +1,54 @@
+import os
+from pathlib import Path
+
+import requests as reqs
+import yaml
+from glean_parser.lint import lint_yaml_files
+
+GITHUB_RAW_URL = "https://raw.githubusercontent.com"
+REPOSITORIES = os.path.join(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "repositories.yaml"
+)
+validation_errors = []
+with open(REPOSITORIES) as data:
+    repos = yaml.load(data, Loader=yaml.SafeLoader)
+    for repo in repos:
+        repo_url = repos[repo]["url"]
+        branch = "master"
+        if "branch" in repos[repo]:
+            branch = repos[repo]["branch"]
+        metrics_files = []
+        if "metrics_files" in repos[repo]:
+            metrics_files = repos[repo]["metrics_files"]
+        if metrics_files:
+            temp_url = (
+                GITHUB_RAW_URL
+                + repo_url.replace("https://github.com", "")
+                + "/"
+                + branch
+                + "/"
+                + metrics_files[0]
+            )
+            response = reqs.get(temp_url)
+            assert response.status_code == 200
+            with open("temp-metrics.yaml", "w") as filehandle:
+                filehandle.write(response.text)
+            yaml_lint_errors = open("yaml-lint-errors.txt", "w")
+            temp_erros = lint_yaml_files(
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -2,8 +2,8 @@
 
 help:
 	@echo "  clean                  Remove build artifacts"
+	@echo "  check-repos            Verify all repositories in repositories.yaml are scrappable"
 	@echo "  lint                   Check style with flake8"
-	@echo "  format                 Format code with black and isort"
```
wlach, line 6: It looks like you accidentally undid some of your work here?

Makefile
```diff
@@ -2,8 +2,8 @@
 
 help:
 	@echo "  clean                  Remove build artifacts"
+	@echo "  check-repos            Verify all repositories in repositories.yaml are scrappable"
```
wlach, line 5:
```suggestion
	@echo "  check-repos            Verify all repositories in repositories.yaml are scrapable"
```

probe_scraper/check_repositories.py
```diff
@@ -0,0 +1,32 @@
...
+    for repo in repos:
+        repo_url = repos[repo]["url"]
+        branch = "master"
+        if "branch" in repos[repo]:
+            branch = repos[repo]["branch"]
+        metrics_files = []
+        if "metrics_files" in repos[repo]:
+            metrics_files = repos[repo]["metrics_files"]
+        if metrics_files:
+            temp_url = (
```
wlach, line 22: This would probably be a bit clearer with an f-string: https://realpython.com/python-f-strings/

probe_scraper/check_repositories.py
```diff
@@ -0,0 +1,32 @@
...
+                GITHUB_RAW_URL
+                + repo_url.replace("https://github.com", "")
+                + "/"
+                + branch
+                + "/"
+                + metrics_files[0]
+            )
+            print("Checking {} metrics_file at {} ...".format(repo, temp_url))
+        response = reqs.get(temp_url)
+        assert response.status_code == 200
```
wlach, line 32: Let's also check that they are valid yaml.

wlach (changes requested):
A few things to change, but overall this looks great! Can you add this check to circleci? I'd probably create a new job for it (see `.circleci/config.yml`)

probe_scraper/check_repositories.py
```diff
@@ -0,0 +1,52 @@
...
+import yaml
+from glean_parser.lint import lint_yaml_files
+
+GITHUB_RAW_URL = "https://raw.githubusercontent.com"
+REPOSITORIES = os.path.join(
+    os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "repositories.yaml"
+)
+validation_errors = []
+with open(REPOSITORIES) as data:
+    repos = yaml.load(data, Loader=yaml.FullLoader)
```
mdboom, line 14: If possible, can we use `SafeLoader` instead of `FullLoader`.  That will prevent loading YAML from running arbitrary code.

probe_scraper/check_repositories.py
```diff
@@ -0,0 +1,52 @@
...
+            )
+            if temp_erros:
+                print(f"Errors found in {repo} metrics file : {print(temp_erros)}")
+                if "prototype" in repos[repo] and not repos[repo]["prototype"]:
+                    validation_errors.append({"repo": repo, "errors": temp_erros})
+    os.remove("temp-metrics.yaml")
+    os.remove("yaml-lint-errors.txt")
+    if validation_errors:
+        print(
+            f"\nSummary of validation errors \n==========================\n {validation_errors}"
```
mdboom, line 50: Maybe we should make this output a bit easier to read.  If I understand correctly, this will print a dictionary mapping from repository name to lists of errors...  I think it would be better to print each error on its own line, under separate sections for each repository.

probe_scraper/check_repositories.py
```diff
@@ -0,0 +1,52 @@
...
+        if metrics_files:
+            temp_url = (
+                GITHUB_RAW_URL
+                + repo_url.replace("https://github.com", "")
+                + "/"
+                + branch
+                + "/"
+                + metrics_files[0]
+            )
+            print(f"Downloading {repo} metrics_file from {temp_url}...")
```
mdboom, line 32: Ideally, this check won't find any problems (that what we'd like to see). So let's remove these print statements and only have output if we've found errors or something problematic (like a network error) happens.  That way, if something goes wrong it's clear what happened, but if everything is ok we aren't filling the output with "noise".

probe_scraper/check_repositories.py
LilyLME: @mdboom what about a json dump?

probe_scraper/check_repositories.py
LilyLME: ok

mdboom (approved):

wlach (approved):
Thank you!

PR #251, opened 2020-10-16
Bug 1669322: Add support for retention_days

mdboom:


fbertsch (approved):
@mdboom how about adding a `retention_days` for `glean-js`? Would be nice to have one available for the downstream work.

mdboom:
> @mdboom how about adding a `retention_days` for `glean-js`? Would be nice to have one available for the downstream work.

Good idea.  done.

PR #255, opened 2020-12-14
Bug 1681299 - Add Focus app ids

fbertsch:


travis79 (approved):
While I'm not entirely sure I'm the right person to include for notifications on this, I will happily own that for now.  (apparently I am listed in several other too :D)

PR #257, opened 2020-12-15
Bug 1682592: Don't break if expires is not a date

mdboom:


fbertsch (approved):
LGTM, it's the correct move to keep the `never` case as a separate and explicit one.

PR #260, opened 2020-12-23
Implement nested repositories.yaml format

jklukas:
Implements the new structure defined in
https://docs.google.com/document/d/1_zx0cxBnnFOhct1cCb_QnXntV-HN9sjqfGfiBwEBL-8/edit?usp=sharing 
but ensures that we maintain the current output format.

This PR also introduces a `/v2/glean/applications` endpoint. Further work will be required to implement v2 metrics and pings endpoints, especially if we want to include dependency resolution there.

The most important piece for review here is about the new `repositories.yaml` format. There are some additional specifics that differ a bit from the proposal, including introducing top-level `libraries` and `application_families` organization. Now is the time to suggest any final refinements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,8 @@ A list of available repositories is at [`/glean/repositories`](https://probeinfo
 
 ## Using the v2 Glean API
 
+> **⚠** The v2 API is in-development and subject to change without notice. **Do not** depend on it in production!
+
 We are incrementally building a v2 API for Glean applications that will
 provide views of underlying probe scraper data that are more directly
 usable by consuming applications. The general direction is discussed in
--- a/README.md
+++ b/README.md
@@ -14,9 +14,12 @@ A web tool to explore the data is available [here](https://telemetry.mozilla.org
 ## Adding a New Glean Repository
 
 To scrape a git repository for probe definitions, an entry needs to be added in `repositories.yaml`.
+Each application repository should be listed under `application_families`.
+The following fields are supported:
 
+- `app_name` is a lowercase_with_underscores name that is short and unambiguous. It needs to be something that can be used both in a generated table name in BigQuery or as a value in an `app_name` string field in a derived table. BigQuery users might be using this literal name in a `WHERE` clause, so it should feel reasonable for that use case. Examples: `fenix`, `firefox_ios`
+- `canonical_app_name` is how we would formally name this application in marketing copy or in a dashboard. This should not be materialized in ETL, but rather only made available in user-facing views since it may be verbose and it may change over time. Examples: Firefox for Android, Firefox for Android (Legacy Fennec), Firefox for Desktop, Firefox Focus for iOS
 - `description`: A brief description of the repository.
-- `channel`: (optional) The release channel. Must be one of "release", "beta", "nightly" or "esr".
 - `deprecated` (optional) `true` if repository corresponds to a deprecated product or library.
 - `prototype`: (optional) `true` if the application is still under developement.
 - `notification_emails`: Where emails about probe-scraper failures and improper files will be forwarded to. These
@@ -25,6 +28,27 @@ will be just about your specific repository.
 - `branch` (optional): The branch in the repository to use. (Defaults to `master`).
 - `metrics_files`: A list of relative paths to `metrics.yaml` files
 - `ping_files`: A list of relative paths to `pings.yaml` files
+
+There is also an `apps` list to define the concrete applications in the family.
+In most cases, there will be only one entry under `apps`, but more complex situations
+like Fenix have separate apps for different logical release channels.
+Each application can override specific fields like `description` from the family, but
+additionally must provide the following:
+
+- `app_id` is the app’s identifier exactly as it appears in the relevant app store listing (for relevant platforms) or in the app’s Glean initialization call (for other platforms). In the case of the Google Play store, it may contain both dots and dashes. For applicable platforms, you should be able to construct an app store URL from this value.
+- `v1_name` is a unique identifier used to look up the app in the V1 probe scraper API;
+  it appears as `name` in the `/glean/repositories` endpoint.
+  The v2 API will use `document_namespace` for this purpose instead, so `v1_name` may
+  eventually be removed.
+- `app_channel`: (optional) The release channel. Must be one of "release", "beta", "nightly" or "esr".
+  If this is not specified, then we assume the application provides a reasonable channel value in
+  `client_info.app_channel` in its pings.
+
+Note that the output in the `/v2/glean/applications` endpoint includes some
+additional derived fields that aren't defined directly in the yaml:
+
+- `document_namespace` is derived from `app_id`, normalized to contain only dashes as punctuation, and it is what Glean inserts into URLs when submitting telemetry.
+- `bq_dataset_family` is derived from app_id, normalized to contain only underscores as punctuation, and it is the value that appears in the names of BigQuery datasets associated with the given application.
 
 ### Adding an application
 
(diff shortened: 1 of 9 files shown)


jklukas:
Also cc @wlach for review. It's not letting me set you as a reviewer for some reason.

repositories.yaml
```diff
@@ -1,502 +1,460 @@
...
+    notification_emails:
+      - frank@mozilla.com
+      - mdroettboom@mozilla.com
+    url: https://github.com/mozilla/glean
+    branch: main
+    metrics_files:
+      - glean-core/metrics.yaml
+    ping_files:
+      - glean-core/pings.yaml
+    library_names:
```
wlach, line 15: Looking at this file, I wonder if `library_names` is redundant?

I might also suggest putting this field on top.

```suggestion
    names:
```

repositories.yaml
```diff
@@ -1,502 +1,460 @@
...
+      - frank@mozilla.com
+      - sync-team@mozilla.com
+    url: https://github.com/mozilla/application-services
+    metrics_files:
+      - components/places/android/metrics.yaml
+    branch: main
+    library_names:
+      - org.mozilla.components:places
+application_families:
+  - app_name: firefox_desktop
```
wlach, line 164: Wondering the same thing here.

```suggestion
  - name: firefox_desktop
```

repositories.yaml
```diff
@@ -1,502 +1,460 @@
...
+      - toolkit/components/glean/metrics.yaml
+    ping_files:
+      - toolkit/components/glean/pings.yaml
+    dependencies:
+      - org.mozilla.deprecated:glean
+    apps:
+      - v1_name: firefox-desktop
+        app_id: firefox-desktop
+  - app_name: fenix
+    canonical_app_name: Firefox for Android
```
wlach, line 179: And here

```suggestion
    canonical_name: Firefox for Android
```

wlach (changes requested):
Aside from some name bikeshedding and a few doc nits, this looks good to me!

jklukas:
@wlach - Thanks for the feedback on naming. There's an underlying concept here that I probably have not expressed explicitly of what I might call "end-to-end field names". I would like to get to a state where `app_name`, `canonical_app_name`, `app_id`, etc. are entirely unambiguous. If you see a table in BQ with a field named `app_name`, it should be populated with the value defined in repositories.yaml.

From that perspective, I think a bit of redundancy in the names is fine in order to achieve the explicitness that the value shown as `app_name` here will be reflected in BQ tables in fields called `app_name`.

If that makes sense to you, I'll see if I can capture that idea more explicitly in the documentation.

repositories.yaml
jklukas: I want the field name to be fairly self-documenting; I worry that `names` is too general and ambiguous. The real purpose of this field is to provide a link between the names given by apps in their `dependencies` field, so it might be nice to give this some name related to "dependency". Something like `dependency_names`, `declares_dependencies`, `defines_dependencies`, etc.

I think I like `defines_dependencies`. It's a bit verbose, but seems much more descriptive about the purpose of the field.

jklukas:
## What is an application?

The current state of this PR defines a concept of "application families", which contain one or more "apps". In this view of the world, apps are unique per `app_id`, and several apps within a family share the same `app_name`.

I am now having second thoughts about this. All the channels of Fenix make up one logical application, and it seems we should optimize for that. In that view, we would have a top-level `applications` key in repositories.yaml rather than `application_families`. Each app has a single `app_name`. For an app like Fenix with several different app_id values, we would nest those under a `channels` key (rather than `apps`).

From the API standpoint, the name `applications` no longer makes sense for the unnested view. Instead, I think it would have to be `/v2/glean/namespaces`. This is less obvious than `applications` but more true to the real structure we have, where the `v2` API is going to be keyed by `document_namespace`.

fbertsch (changes requested):
In general, these changes align with the proposal you put together, so I'm +1 on that (with most of wlach's comments). However, we're still going to be supporting V1. With this change, we'll be losing the tests of the v1 schema, and complicating the associated code.

For v2, I recommend creating a new `Repository` class that inherits from `Repository`. New test suites can be associated with that new class, so we can cleanly separate the new code/tests from the old, and still run tests on the v1 API.

In fact, it might be best to also create a new function in `runner` which is for v2. That would make it easy for us to iterate on the v2 API separately from v1. (note that overall I'm not a fan of the `runner` code as it exists, so happy to see improvement in this idea.)

jklukas:
> For v2, I recommend creating a new Repository class that inherits from Repository. New test suites can be associated with that new class, so we can cleanly separate the new code/tests from the old, and still run tests on the v1 API.

I'll think some more about this. Taking that approach would require some shim code to "downgrade" the v2 yaml format to the v1 yaml format, but it may indeed make the underlying code cleaner, and make an eventual removal of the v1 API more feasible.

repositories.yaml
```diff
@@ -1,502 +1,460 @@
...
+    notification_emails:
+      - frank@mozilla.com
+      - sync-team@mozilla.com
+    url: https://github.com/mozilla/application-services
+    metrics_files:
+      - components/places/android/metrics.yaml
+    branch: main
+    library_names:
+      - org.mozilla.components:places
+application_families:
```
badboy, line 163: This file is largely for consumption my machines, nonetheless it will be edited by humans. Should we add some newlines to separate out the categories, while we're at it?

```suggestion

application_families:
```

We could even go further and add new lines between individual app definitions

README.md
```diff
@@ -14,9 +14,12 @@ A web tool to explore the data is available [here](https://telemetry.mozilla.org
 ## Adding a New Glean Repository
 
 To scrape a git repository for probe definitions, an entry needs to be added in `repositories.yaml`.
+Each application repository should be listed under `application_families`.
```
badboy, line 17: This is the first mention of `application_families`. Probably fine for people that already knwo about it, but it might not be clear to everyone that a "family" is "same ~code base, but different app ids/builds/..."

I also don't know how to structure this better at the moment, so consider this not fully-blocking, but maybe some future improving/rewording needed.

README.md
```diff
@@ -26,6 +29,27 @@ will be just about your specific repository.
 - `metrics_files`: A list of relative paths to `metrics.yaml` files
 - `ping_files`: A list of relative paths to `pings.yaml` files
 
+There is also an `apps` list to define the concrete applications in the family.
```
badboy, line 32: With regard to my point above, maybe we can start this with something like:

```suggestion
An application family consists of one or more applications built from the same code base. 

There is also an `apps` list to define the concrete applications in the family.
```

And then follow up with the Fenix example. That puts "application family" as a term in the text and explains it upfront and is only then followed by concrete examples.
(wording to be bikeshed)

README.md
```diff
@@ -26,6 +29,27 @@ will be just about your specific repository.
...
+In most cases, there will be only one entry under `apps`, but more complex situations
+like Fenix have separate apps for different logical release channels.
+Each application can override specific fields like `description` from the family, but
```
badboy, line 35: only some or all can be overwritten?
I assume e.g. `url` can't. This should be clarified here.

README.md
```diff
@@ -43,11 +67,12 @@ $ ./gradlew :app:dependencies
 
 ### Adding a library
 
-All **libraries** must define `library_names`.
-
-Probe scraper also needs a way to map dependencies back to an entry in the
-`repositories.yaml` file. Therefore, any libraries defined should also include
-their build-system-specific library names in the `library_names` parameter.
+All **libraries** are defined under the `libraries` top-level key.
+They support a subset of the same fields as applications, and additionally
```
badboy, line 71: Where is this subset defined?

README.md
```diff
@@ -355,3 +380,15 @@ For example, the data for a repository called `fenix` would be found at [`/glean
 
 A list of available repositories is at [`/glean/repositories`](https://probeinfo.telemetry.mozilla.org/glean/repositories).
 
+## Using the v2 Glean API
```
badboy, line 383: Not to be confused with the Glean SDK API (scnr, igore me)

badboy (changes requested):
(Haven't looked at the implementation yet)

See inline comments. Overall a good start, I think we should be more clear in the docs what some parts mean though

badboy:
> ## What is an application?
> 
> [...] I am now having second thoughts about this. 

Of course this came in only after I already started the review.
I _think_ "channels" might be better, but I need to re-read the docs with that in mind (didn't we discuss that in the proposal too?)

jklukas:
>  I think "channels" might be better, but I need to re-read the docs with that in mind (didn't we discuss that in the proposal too?)

There was some discussion there, yes. I will make sure to revisit that, and I will add notes in the proposal about any changes we end up implementing here compared to the format described in the doc.

jklukas:
I've opened a new PR https://github.com/mozilla/probe-scraper/pull/262 which takes a more complete approach here and provides better grounding for a future v2 API. It defines an OpenAPI spec and publishes rendered docs.

PR #262, opened 2021-01-11
Implement nested repositories.yaml with OpenAPI definition

jklukas:
Supersedes #260

This seeks to maintain existing v1 logic as much as possible, separating out the new v2 repostories.yaml and implementing a compatibility layer to "downgrade" it to the old v1 schema.

Additionally, it moves schema definitions to an OpenAPI configuration and publishes API docs to https://mozilla.github.io/probe-scraper ; if this looks good to everyone, we can migrate all schema documentation there and remove most of the README content.


probeinfo_api.yaml
```diff
@@ -0,0 +1,562 @@
...
+    AppName:
+      type: string
+      pattern: "^[a-z][a-z0-9_]{0,39}$"
+      description: |
+        A lowercase_with_underscores name that is short and unambiguous. It needs to be something that can be used both in a generated table name in BigQuery or as a value in an `app_name` string field in a derived table. BigQuery users might be using this literal name in a `WHERE` clause, so it should feel reasonable for that use case. Examples: `fenix`, `firefox_ios`
+      example: firefox_ios
+    CanonicalAppName:
+      type: string
+      description: |
+        How we would formally name this application in marketing copy or in a dashboard. This should not be materialized in ETL, but rather only made available in user-facing views since it may be verbose and it may change over time. Examples: Firefox for Android, Firefox for Android (Legacy Fennec), Firefox for Desktop, Firefox Focus for iOS
```
jklukas, line 462: I want to add in some commentary about app_name, canonical_app_name, etc. being end-to-end names as discussed in https://github.com/mozilla/probe-scraper/pull/260#discussion_r553341010

probeinfo_api.yaml
```diff
@@ -0,0 +1,562 @@
...
+    Description:
+      type: string
+      description: A brief free-text description of the repository
+    AppName:
+      type: string
+      pattern: "^[a-z][a-z0-9_]{0,39}$"
+      description: |
+        A lowercase_with_underscores name that is short and unambiguous. It needs to be something that can be used both in a generated table name in BigQuery or as a value in an `app_name` string field in a derived table. BigQuery users might be using this literal name in a `WHERE` clause, so it should feel reasonable for that use case. Examples: `fenix`, `firefox_ios`
+      example: firefox_ios
+    CanonicalAppName:
```
jklukas, line 459: We should perhaps add more newlines in this file and repositories.yaml per discussion in https://github.com/mozilla/probe-scraper/pull/260#discussion_r553364614

README.md
```diff
@@ -11,20 +11,15 @@ The data is pulled from two different sources:
...
-- `channel`: (optional) The release channel. Must be one of "release", "beta", "nightly" or "esr".
-- `deprecated` (optional) `true` if repository corresponds to a deprecated product or library.
-- `prototype`: (optional) `true` if the application is still under developement.
-- `notification_emails`: Where emails about probe-scraper failures and improper files will be forwarded to. These
-will be just about your specific repository.
-- `url`: The URL of the repository to scrape. It should be able to be cloned directly from that URL.
-- `branch` (optional): The branch in the repository to use. (Defaults to `master`).
-- `metrics_files`: A list of relative paths to `metrics.yaml` files
-- `ping_files`: A list of relative paths to `pings.yaml` files
+Refer the the [schema docs](https://mozilla.github.io/probe-scraper/#operation/getReposYaml).
```
wlach, line 22: As mentioned, I'm not sure about the OpenAPI stuff, but if we're going to use it I would rewrite this to:

```suggestion
For a full list of fields that need to be added (and what they mean), see the [schema documentation](https://mozilla.github.io/probe-scraper/#operation/getReposYaml).
```

probe_scraper/parsers/repositories.py
```diff
@@ -2,15 +2,30 @@
...
 
+from probe_scraper import model_validation
+
 REPOSITORIES_FILENAME = "repositories.yaml"
 REPOSITORIES_SCHEMA = "schemas/repositories.json"
 
 
+def remove_none(obj):
+    """
+    See https://stackoverflow.com/a/20558778
```
wlach, line 15: Can you give a quick inline comment on what this does?

wlach (changes requested):
I'm a little bit torn about the OpenAPI stuff. On the one hand, it definitely does look nice, on the other I think it's confusing for our documentation about how users/developers should define things to be conflated with the interface we're providing to consumers. It's IMO kind of weird that "repositories.yaml" is defined as an endpoint.

I know this is more work, but I wonder if it might make more sense to generate both the OpenAPI documentation and reference documentation for developers based on some kind of underlying representation? I'm kind of imagining that the rendered documentation would look like:

https://mozilla.github.io/probe-scraper --> Introduction to project, link out to API documentation, developer docs
https://mozilla.github.io/probe-scraper/api --> API documentation for users

Thoughts?

jklukas:
> I know this is more work, but I wonder if it might make more sense to generate both the OpenAPI documentation and reference documentation for developers based on some kind of underlying representation?

The model definitions in the OpenAPI config are JSON schemas, so it would be entirely possible to remove the `repositories.yaml` pseudo-endpoint, but use the appropriate schema to generate separate documentation for the yaml input, perhaps with https://pypi.org/project/json-schema-for-humans/ .

jklukas:
> It's IMO kind of weird that "repositories.yaml" is defined as an endpoint.

Yes, it was definitely weird. I figured out, though, how to have ReDoc render a model to the list. I pushed that change and it's now live at https://mozilla.github.io/probe-scraper/#tag/repositories.yaml ; it definitely looks a bit less weird now. It's still not great that the example is in JSON format, though. We could render this to a completely separate page instead (while still using ReDoc). Let me know if this looks any better, @wlach.

wlach:
> > It's IMO kind of weird that "repositories.yaml" is defined as an endpoint.
> 
> Yes, it was definitely weird. I figured out, though, how to have ReDoc render a model to the list. I pushed that change and it's now live at https://mozilla.github.io/probe-scraper/#tag/repositories.yaml ; it definitely looks a bit less weird now. It's still not great that the example is in JSON format, though. We could render this to a completely separate page instead (while still using ReDoc). Let me know if this looks any better, @wlach.

Much better! I think this is good enough for now.

repositories.yaml
```diff
@@ -1,502 +1,471 @@
...
+    url: https://github.com/mozilla-mobile/reference-browser
+    prototype: true
+    metrics_files:
+      - app/metrics.yaml
+    dependencies:
+      - org.mozilla.components:service-glean
+      - org.mozilla.components:lib-crash
+    channels:
+      - v1_name: reference-browser
+        app_id: org-mozilla-reference-browser
```
tdsmith, line 278: drive-by: I think this should be `org.mozilla.reference.browser`; cf https://github.com/mozilla-mobile/reference-browser/blob/349b13324f35bf6357af4a07ae875cc58aec7753/app/src/main/AndroidManifest.xml#L7

jklukas:
@badboy @wlach @fbertsch - I spent a couple hours this morning doing final cleanup here. All outstanding comments have been addressed and this is ready for final review.

The published docs https://mozilla.github.io/probe-scraper/ reflect the current state here, and I've moved all documentation about specific fields/endpoints out of the README and instead into those rendered API docs.

I've double-checked that the `/glean/repostories` output with these changes is identical to the current endpoint modulo a few small details of whitespace (we've removed some trailing newlines in description fields) and order of dependencies in the dependency arrays.

fbertsch (approved):
This is a great addition to probe-scraper. I find the new joining of schemas in `probeinfo_api.yaml` to be confusing (used to both describe the API and, internally, to validate data), so we should add some clear documentation on what the relationship is in the readme.

It's also not clear to me how to handle this version bumping in the codebase. Basically we're keeping around the code to validate/use version 1, but we will only ever be using version 2 moving forward.

@jklukas can you run this and confirm that the output is unchanged, except for the new endpoint?

jklukas:
> @jklukas can you run this and confirm that the output is unchanged, except for the new endpoint?

I posted [a gist](https://gist.github.com/jklukas/5b5f3b5b51283c406de5a4bd33dd93df) that gives the `diff -y` output between the `repositories` json content before and after this change. If you're unfamiliar with `diff -y` format, any lines that differ are marked in the middle with a `|`, `<` or `>`.

jklukas:
> It's also not clear to me how to handle this version bumping in the codebase. Basically we're keeping around the code to validate/use version 1, but we will only ever be using version 2 moving forward.

My take on this is that the existing "v1" endpoints will continue to be built based on the `RepositoriesParser.parse` method which does the downgrade to v1 repositories.yaml format as its first step. As we build out new endpoints, they will use the output of the `RepositoriesParser.parse_v2` method which maintains all the new fields, etc. Does that make sense? I can add some better commentary in the docstrings to make this explicit.

fbertsch:
> > It's also not clear to me how to handle this version bumping in the codebase. Basically we're keeping around the code to validate/use version 1, but we will only ever be using version 2 moving forward.
> 
> My take on this is that the existing "v1" endpoints will continue to be built based on the `RepositoriesParser.parse` method which does the downgrade to v1 repositories.yaml format as its first step. As we build out new endpoints, they will use the output of the `RepositoriesParser.parse_v2` method which maintains all the new fields, etc. Does that make sense? I can add some better commentary in the docstrings to make this explicit.

Ah, I see - that does make sense. This is superior to keeping both because then we may run into copy+paste errors. It might be good to elucidate that.

README.md
```diff
@@ -9,22 +9,18 @@ The data is pulled from two different sources:
...
```
wlach, line 23:
```suggestion
Refer to the [repository schema documentation](https://mozilla.github.io/probe-scraper/#tag/repositories.yaml) for details.
```

README.md
```diff
@@ -9,22 +9,18 @@ The data is pulled from two different sources:
 - From [`hg.mozilla.org`](https://hg.mozilla.org) for Firefox data.
 - From a [configurable set of Github repositories](repositories.yaml) that use [Glean](https://github.com/mozilla-mobile/android-components/tree/master/components/service/glean).
 
-A web tool to explore the data is available [here](https://telemetry.mozilla.org/probe-dictionary/).
+A web tool to explore the Firefox-related data is available at
+[probes.telemetry.mozilla.org](https://probes.telemetry.mozilla.org/).
+
+CI on this repository handles publishing the data to populate the `probeinfo` API
```
wlach, line 15: I wonder if we should reverse the order of this paragraph with the one above it. Normally I would say that the user facing tools are the most important, but if someone's spelunking through this repo they probably care more about the grotty details. Maybe something like:

```md
Probe Scraper outputs JSON to https://probeinfo.telemetry.mozilla.org. Effectively, this creates a REST API which can be used by downstream tools like [mozilla-schema-generator](https://github.com/mozilla/mozilla-schema-generator) and various data dictionary type applications (see below). An [OpenAPI reference](https://mozilla.github.io/probe-scraper/) to this API is available.

A web tool to explore the Firefox-related data is available at [probes.telemetry.mozilla.org](https://probes.telemetry.mozilla.org/). A project to develop a similar view for Glean-based data
is under development in the [Glean Dictionary](https://github.com/mozilla/glean-dictionary).
```

Thoughts?

wlach (changes requested):
Basically just README nits at this point. I feel like there might still be changes we might want as I try to update the glean dictionary to use this API, but I don't see why that should block landing this PR.

README.md
jklukas: Looks great. I'll integrate that text.

wlach (approved):

jklukas:
@badboy - If you have additional feedback, I'm still happy to take feedback that here and I'll address in a follow-up PR.

badboy:
> @badboy - If you have additional feedback, I'm still happy to take feedback that here and I'll address in a follow-up PR.

yeah, didn't manage to get through it yesterday, but I didn't have major concerns against it. Will give it a look and post followups if necessary

badboy:
I really like the new API description. Yey for interactive documentation!
It's a bit unfortunate that due to how examples are specified it mixes names that don't work together:

In the example below names are all over the place because `v1_name`, etc. are re-used across both `libraries` and `applications`.
Not much we can do about that except duplicating the documentation parts I guess, so not sure it's really worth fixing.

<details>

```
{
  "version": "2",
  "libraries": [
    {
      "v1_name": "firefox-android-beta",
      [...]
      "library_names": [
        "org.mozilla.components:service-glean"
      ]
    }
  ],
  "applications": [
    {
      "app_name": "firefox_ios",
      "canonical_app_name": "Firefox for iOS",
      [...]
      "channels": [
        {
          "app_id": "org.mozilla.firefox_beta",
          "v1_name": "firefox-android-beta",
          "app_channel": "release",
          "description": "string",
          "deprecated": false,
          "additional_dependencies": []
        }
      ]
    }
  ]
}
```
</details>

Otherwise I'm quite happy how this turned out and I think the repositories.yaml is easier to read and edit now

PR #266, opened 2021-01-28
Bug 1689513 - Remove glean-js prototype from repositories.yaml

brizental:
Hey folks, that glean-js has been archived a while back so I think it is time to remove it from the repositories list.

This is my first time sending a PR on this repo, let me know if I am missing any steps :)

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -463,25 +463,6 @@ applications:
       - v1_name: mach
         app_id: mozilla-mach
 
-  - app_name: glean_js
-    canonical_app_name: Experimentation with Glean.js
-    description: Experimentation with Glean.js
-    url: https://github.com/brizental/gleanjs
-    prototype: true
-    notification_emails:
-      - brizental@mozilla.com
-      - aplacitelli@mozilla.com
-      - mdroettboom@mozilla.com
-    branch: main
-    metrics_files:
-      - metrics.yaml
-    dependencies:
-      - glean-core
-    retention_days: 90
-    channels:
-      - v1_name: glean-js
-        app_id: glean-js-tmp
-
   - app_name: focus_ios
     canonical_app_name: Firefox Focus for iOS
     description: Firefox Focus on iOS. Klar is the sibling application


fbertsch (commented):
@brizental just to be clear, merging this will result in a total loss of all that data. cc @whd

brizental:
> @brizental just to be clear, merging this will result in a total loss of all that data. cc @whd

That is fine by me. Unless it is problematic for the pipeline, we can go ahead with the deletion even if that means loss of data.

fbertsch (approved):
Okay, I want to get the go-ahead from whd before we merge this.

jklukas:
As a first step, can we just add `deprecated: true` here without deleting the app? That won't have any pipeline implications, but makes the intent clear that it's no longer used.

Then cleaning up the tables and actually removing from repositories.yaml can be a follow-up.

brizental:
Not sure if I understand this correctly, but I see that glean-js has `retention_days: 90`. Does that mean that data is dropped after 90 days? If so then it's possible glean-js doesn't even have data at this point anymore.

jklukas:
> Does that mean that data is dropped after 90 days?

Yes.

> If so then it's possible glean-js doesn't even have data at this point anymore.

Indeed it would appear that the tables are already entirely empty:

```
SELECT date(submission_timestamp) as dt, count(*) 
FROM `moz-fx-data-shared-prod.glean_js_tmp_stable.*`
where submission_timestamp > '2000-01-01'
group by 1 order by 1

-- No results
```

fbertsch:
I'm not sure we need to wait and push this out to multiple steps. Let's get this merged as-is if :whd doesn't have any qualms about it.

whd:
There's a (technically ratified but not moved to mana yet) [document](https://docs.google.com/document/d/1a89FH0jjByAjKkl3BxNtLAgqDtZr2HTp9550t1ZyBww/edit?ts=5f611ad9#heading=h.tvu1uw25tntj) that includes the step of explicitly marking this data as unwanted in beam in advance of table deletes. Let's do that to avoid the (very unlikely) case that we properly decode a message but then be unable to route it to a dataset that has been deleted.

PR #267, opened 2021-01-28
Bug 1689167: Reduce memory usage in probe scraper

mdboom:
Probe scraper loads all probes over all time into memory, before slicing the information in a few different ways and writing it back out.  I measured it using 30GB of RAM before I gave up on it because my machine slowed to a crawl.

There are probably some more clever ways to avoid loading all the data at once, instead streaming data in and out and making greater use of the disk.

However, as a stop gap measure, the approach in this PR seems to work quite well.  Even though there are many (3,400) revisions of the probe metadata files, the individual probes for the most part don't change at every revision.  Therefore, we can save a lot of memory by reusing a probe when we've already seen one with the exact same contents.

To test the impact of this, I ran this over the nightly channel only (running all channels would have taken much more testing time, and nightly represents a "worst case").  With this optimization, the maximum memory use is 1.2 GB vs. 17.5 GB.  Runtime is also slightly lower: 22 min vs. 28 min wall clock time.  Obviously Airflow will behave differently than my local machine, but I think this a solid improvement for something that's pretty localized.

Note: the Glean metrics parsing code path is completely separate.  It will have this same problem *some day* but for now the revisions and number of probes remains much lower.

# Before:

```
1638.12s user 18.49s system 98% cpu 27:57.03 total
```

image

# After:

```
1300.64s user 3.92s system 99% cpu 21:54.72 total
```

image

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -151,12 +151,38 @@ def parse_moz_central_probes(scraped_data):
     }
     """
 
+    lookup_table = {}
+
+    def dedupe_probes(results):
+        # Most probes have exactly the same contents across revisions, so we
+        # can get significant memory savings by deduplicating them across the
+        # entire history.
+        deduped = {}
+        for key, value in results.items():
+            # Get a stable hash for a dict, by sorting the keys when writing
+            # out values.
+            probe_hash = hash(json.dumps(value, sort_keys=True))
+            lookup_for_name = lookup_table.get(key, None)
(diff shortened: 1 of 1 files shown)


mdboom:
Ugh -- sorry -- I shouldn't have pushed right before a meeting.  Try again now -- I think all of the weirdness is fixed and it's passing the tests...

fbertsch (approved):
Okay, this looks good!

PR #268, opened 2021-02-04
Bug 1690112 - Stop sharing probe definitions in final transform

fbertsch:


mdboom (approved):
Got it.  Do the thing to save memory, and then undo the thing :)

PR #271, opened 2021-02-11
Link back to metric and ping definitions in source control

Iinh:
Add a link back to where a metric/ping is originally defined in source control, as the second step of https://github.com/mozilla/glean-dictionary/issues/223. 

@wlach Could you review if this looks ok so far? Thanks!

### DIFF
--- a/probe_scraper/parsers/pings.py
+++ b/probe_scraper/parsers/pings.py
@@ -6,7 +6,7 @@ from pathlib import Path
 
 from glean_parser.parser import parse_objects
 
-from .utils import add_source_url
+from .utils import get_source_url
 
 PING_NAME_NORMALIZATION = {
     "deletion_request": "deletion-request",
@@ -42,6 +42,7 @@ class GleanPingsParser:
 
         if repo_url and commit_hash:
             for v in pings.values():
-                v = add_source_url(v, repo_url, commit_hash)
-
+                v["source_url"] = get_source_url(v["defined_in"], repo_url, commit_hash)
+                # the 'defined_in' structure is no longer needed
+                del v["defined_in"]
         return pings, errors
--- a/probe_scraper/parsers/pings.py
+++ b/probe_scraper/parsers/pings.py
@@ -6,6 +6,8 @@ from pathlib import Path
 
 from glean_parser.parser import parse_objects
 
+from .utils import add_source_url
+
 PING_NAME_NORMALIZATION = {
     "deletion_request": "deletion-request",
     "bookmarks_sync": "bookmarks-sync",
@@ -25,18 +27,21 @@ class GleanPingsParser:
     to parse the pings.yaml files.
     """
 
-    def parse(self, filenames, config):
+    def parse(self, filenames, config, repo_url=None, commit_hash=None):
         config = config.copy()
         paths = [Path(fname) for fname in filenames]
         paths = [path for path in paths if path.is_file()]
         results = parse_objects(paths, config)
         errors = [err for err in results]
 
-        return (
-            {
-                normalize_ping_name(ping_name): ping_data.serialize()
-                for category, pings in results.value.items()
-                for ping_name, ping_data in pings.items()
-            },
-            errors,
-        )
+        pings = {
+            normalize_ping_name(ping_name): ping_data.serialize()
+            for category, pings in results.value.items()
+            for ping_name, ping_data in pings.items()
+        }
+
+        if repo_url and commit_hash:
+            for v in pings.values():
+                v = add_source_url(v, repo_url, commit_hash)
+
+        return pings, errors
--- a/probe_scraper/parsers/pings.py
+++ b/probe_scraper/parsers/pings.py
@@ -20,6 +20,17 @@ def normalize_ping_name(name):
     return PING_NAME_NORMALIZATION.get(name, name)
 
 
+def generate_definition(ping_data, repo_url, commit_hash):
+    serialized = ping_data.serialize()
+    if repo_url and commit_hash:
+        serialized["source_url"] = get_source_url(
+            serialized["defined_in"], repo_url, commit_hash
+        )
+        # the 'defined_in' structure is no longer needed
+        del serialized["defined_in"]
+    return serialized
+
+
 class GleanPingsParser:
     """
     Use the [Glean Parser]
@@ -35,14 +46,11 @@ class GleanPingsParser:
         errors = [err for err in results]
 
         pings = {
-            normalize_ping_name(ping_name): ping_data.serialize()
+            normalize_ping_name(ping_name): generate_definition(
+                ping_data, repo_url, commit_hash
+            )
             for category, pings in results.value.items()
             for ping_name, ping_data in pings.items()
         }
 
-        if repo_url and commit_hash:
-            for v in pings.values():
-                v["source_url"] = get_source_url(v["defined_in"], repo_url, commit_hash)
-                # the 'defined_in' structure is no longer needed
-                del v["defined_in"]
         return pings, errors
--- a/probe_scraper/parsers/utils.py
+++ b/probe_scraper/parsers/utils.py
@@ -26,10 +26,10 @@ def get_major_version(version):
     return version.split(".")[0]
 
 
-def get_source_url(defined_in, repo_url, commit_hash):
+def get_source_url(glean_definition, repo_url, commit_hash):
     """Add source URL where metrics and pings are defined."""
-    line_number = defined_in["line"]
-    file_path = defined_in["filepath"][
-        defined_in["filepath"].find(commit_hash) :  # noqa: E203
+    line_number = glean_definition["line"]
+    file_path = glean_definition["filepath"][
+        glean_definition["filepath"].find(commit_hash) :  # noqa: E203
     ]
     return f"{repo_url}/blob/{file_path}#L{line_number}"
--- a/probe_scraper/parsers/utils.py
+++ b/probe_scraper/parsers/utils.py
@@ -24,3 +24,15 @@ def get_major_version(version):
     :return: a string containing the leftmost number before the first dot
     """
     return version.split(".")[0]
+
+
+def add_source_url(v, repo_url, commit_hash):
+    """Add source URL where metrics and pings are defined."""
+    file_path = v["defined_in"]["filepath"][
+        v["defined_in"]["filepath"].find(commit_hash) :  # noqa: E203
+    ]
+    line_number = v["defined_in"]["line"]
+    v["source_url"] = f"{repo_url}/blob/{file_path}#L{line_number}"
+    # the 'defined_in' structure is no longer needed
+    del v["defined_in"]
+    return v
--- a/probe_scraper/parsers/utils.py
+++ b/probe_scraper/parsers/utils.py
@@ -26,13 +26,10 @@ def get_major_version(version):
     return version.split(".")[0]
 
 
-def add_source_url(v, repo_url, commit_hash):
+def get_source_url(defined_in, repo_url, commit_hash):
     """Add source URL where metrics and pings are defined."""
-    file_path = v["defined_in"]["filepath"][
-        v["defined_in"]["filepath"].find(commit_hash) :  # noqa: E203
+    line_number = defined_in["line"]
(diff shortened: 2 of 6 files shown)


probe_scraper/runner.py
```diff
@@ -257,6 +257,38 @@ def load_moz_central_probes(
...
+    Example input:
+    '/var/folders/yf/q5zg5fjx2kz2lc2sbfdht4bm0000gn/T/tmpt9x11our/glean-android/
+    530a5128526fcf133d57ec3d188590c600895fa1/glean-core/metrics.yaml'
+
+    Output:
+    '530a5128526fcf133d57ec3d188590c600895fa1/glean-core/metrics.yaml'
+    """
+    split_path = "".join(path).split("/")
+    for i, word in enumerate(split_path):
+        if len(word) == 40:  # i.e. a commit hash
```
wlach, line 287: I think it would be better to pass the commit hash (which we have) into this function and splitting on that -- that should be more reliable and simpler.

probe_scraper/runner.py
```diff
@@ -280,16 +312,26 @@ def load_glean_metrics(cache_dir, out_dir, repositories_file, dry_run, glean_rep
         for commit_hash, paths in commits.items():
             metrics_files = [p for p in paths if p.endswith(GLEAN_METRICS_FILENAME)]
             pings_files = [p for p in paths if p.endswith(GLEAN_PINGS_FILENAME)]
-
```
wlach, line 283: Guessing you didn't mean to delete this line.

probe_scraper/runner.py
```diff
@@ -280,16 +312,26 @@ def load_glean_metrics(cache_dir, out_dir, repositories_file, dry_run, glean_rep
...
             try:
                 config = {"allow_reserved": repo_name.startswith("glean")}
+                for d in repositories:
+                    if vars(d)["name"] == repo_name:
+                        repo_dict = d.to_dict()
```
wlach, line 319: I'm pretty sure this should work. Also, I think we should just use the name `repo` as opposed to `repo_dict`.

```suggestion
                repo = next(r for r in repositories if r.name == repo_name)
```

wlach (commented):
This looks pretty reasonable to me. It might be good to get someone like @fbertsch (who AFAIK wrote most of this glean-based code originally) to do the final review.

Iinh:
Okay I think I've addressed all of your review comments @wlach, please let me know if I've missed anything. And yes it'd be great if @fbertsch could do the final review for this PR. Thank you both!

probe_scraper/runner.py
```diff
@@ -257,6 +257,38 @@ def load_moz_central_probes(
...
+    530a5128526fcf133d57ec3d188590c600895fa1/glean-core/metrics.yaml'
+
+    Output:
+    '530a5128526fcf133d57ec3d188590c600895fa1/glean-core/metrics.yaml'
+    """
+    split_path = "".join(path).split("/")
+    for i, word in enumerate(split_path):
+        if word == commit_hash:
+            file_path = "/".join(split_path[i:])
+    return file_path
```
wlach, line 289: I *think* you can replace this with this one-liner:

```suggestion
    return path.split(commit_hash)[-1][1:]
```

wlach (commented):
Can you give some example output from this? E.g. a metrics output for burnham.

Tagging @fbertsch for review as well.

fbertsch (changes requested):
We should push this field addition into the parsers themselves. The runner in general is overloaded, and we're trying to actively avoid adding new logic there.

It should be straightforward, since the parsers already receive a list of files. Note though that it is a _list_, not a single file - each Glean application can have more than one metrics or ping files. The only addition it needs is the url of the repository.

wlach:
> It should be straightforward, since the parsers already receive a list of files. Note though that it is a _list_, not a single file - each Glean application can have more than one metrics or ping files. The only addition it needs is the url of the repository.

Just to be clear, do you mean this change should happen inside [glean_parser](https://github.com/mozilla/glean_parser/)?

fbertsch:
> Just to be clear, do you mean this change should happen inside [glean_parser](https://github.com/mozilla/glean_parser/)?

I suppose it could. Either there or in `parsers/metrics.py`. We should check in with the client team to see if they have an opinion on including it in the glean-parser.

wlach:
> > Just to be clear, do you mean this change should happen inside [glean_parser](https://github.com/mozilla/glean_parser/)?
> 
> I suppose it could. Either there or in `parsers/metrics.py`. We should check in with the client team to see if they have an opinion on including it in the glean-parser.

@mdboom @badboy ^ ?

Iinh:
> @wlach Can you give some example output from this? E.g. a metrics output for burnham.

Here are some example outputs:
https://gist.github.com/Iinh/fa3b196aae8c375b695c560c28c351c3
https://gist.github.com/Iinh/e135dfc90e1b09873f6209c3af7e4e6b

badboy:
> > Just to be clear, do you mean this change should happen inside [glean_parser](https://github.com/mozilla/glean_parser/)?
> 
> I suppose it could. Either there or in `parsers/metrics.py`. We should check in with the client team to see if they have an opinion on including it in the glean-parser.

We already expose the line number, the fact that we don't also expose the file path within the repository seems like an oversight to me. `glean_parser` would indeed be the place where we have the data and can expose it.

badboy:
Actually, I would need to read the [PR on coverage data](https://github.com/mozilla/glean_parser/pull/272) again, but I think that tracks files already

wlach:
> > Actually, I would need to read the [PR on coverage data](https://github.com/mozilla/glean_parser/pull/272) again, but I think that tracks files already
> 
> It does, and it's a pretty simple change. That, of course, only tracks the file path within the repo -- glean_parser doesn't understand repos at all...

Perhaps a hybrid approach, where glean parser gives the path / line number and probe-scraper creates the final URL makes the most sense architecturally. Probably best if glean_parser doesn't *assume* that you're (e.g.) using GitHub.

badboy:
We're gonna land the linked PR, and I'm doing a release by tomorrow so we can use that data here.

wlach (changes requested):
@Iinh I think @fbertsch mentioned he'd prefer to see this logic (aside from what's now in glean_parser) moved into the parsers (`parsers/metrics.py` and `parsers/pings.py`, presumably). Could you give that a try?

fbertsch (commented):
@wlach @Iinh I'm looking at this again, and basically what we have now are 4 pieces:
- Repo URL
- Commit
- File path
- Line number

Since we added the latter two to the metric definitions, we should be able to materialize these URLs downstream from here; i.e. not include them directly in the probe info service.

wlach:
> Since we added the latter two to the metric definitions, we should be able to materialize these URLs downstream from here; i.e. not include them directly in the probe info service.

This is true, but I feel like it would be easiest / best to do this work inside probe-scraper. Otherwise the downstream consumer (the Glean Dictionary currently, but maybe other things later) has to put the information together, which I think is more work since it will involve gluing data together from a bunch of different data sources.

fbertsch:
> This is true, but I feel like it would be easiest / best to do this work inside probe-scraper. Otherwise the downstream consumer (the Glean Dictionary currently, but maybe other things later) has to put the information together, which I think is more work since it will involve gluing data together from a bunch of different data sources.

Got it. If you'd prefer it live here, I don't have any specific objections.

probe_scraper/parsers/utils.py
```diff
@@ -24,3 +24,15 @@ def get_major_version(version):
...
+def add_source_url(v, repo_url, commit_hash):
+    """Add source URL where metrics and pings are defined."""
+    file_path = v["defined_in"]["filepath"][
+        v["defined_in"]["filepath"].find(commit_hash) :  # noqa: E203
+    ]
+    line_number = v["defined_in"]["line"]
+    v["source_url"] = f"{repo_url}/blob/{file_path}#L{line_number}"
+    # the 'defined_in' structure is no longer needed
+    del v["defined_in"]
+    return v
```
Iinh, line 38: @wlach I know how you feel about utils... but I wasn't sure where to put this function. Any advice?

wlach: Since it's already there, I think this is fine.

probe_scraper/parsers/utils.py
```diff
@@ -24,3 +24,15 @@ def get_major_version(version):
     :return: a string containing the leftmost number before the first dot
     """
     return version.split(".")[0]
+
+
+def add_source_url(v, repo_url, commit_hash):
```
wlach, line 29:
```suggestion
def get_source_url(defined_in, repo_url, commit_hash):
```

I think it's probably clearer if we make this function just return a source URL and pass in the `defined_in` part of Glean Parser's output. You can then do the various munging inline (it's a little bit of duplication, but not too much and I think worth it for the additional clarity).

wlach (changes requested):
Looks good to me minus nits. @fbertsch wdyt?

probe_scraper/parsers/utils.py
fbertsch: I might call this data structure `glean_definition`

probe_scraper/parsers/pings.py
```diff
@@ -25,18 +27,21 @@ class GleanPingsParser:
...
-        return (
-            {
-                normalize_ping_name(ping_name): ping_data.serialize()
-                for category, pings in results.value.items()
-                for ping_name, ping_data in pings.items()
-            },
-            errors,
-        )
+        pings = {
+            normalize_ping_name(ping_name): ping_data.serialize()
```
fbertsch, line 38: I would prefer the below addition live inside a separate function alongside this serialized call:
```suggestion
            normalize_ping_name(ping_name): generate_definition(ping_data, repo_url, commit_hash)
```

probe_scraper/parsers/utils.py
fbertsch: What Will is suggesting is to avoid side effects, which is a good pattern that really evolves from functional programming.

probe_scraper/parsers/pings.py
Iinh: Hi @fbertsch could you help elaborate more on this? Since this function only takes `defined_in` and returns the source URL, I think what you recommended might not work. Maybe I'm still not understanding it quite right?

fbertsch: No worries - what I'm proposing is a new function inside the `pings.py` file called `generate_definition`. That function would _call_ `get_source_url`, e.g.:

```
def generate_definition(ping_data, repo_url, commit_hash):
  serialized = ping_data.serialize()
  if repo_url and commit_hash:
    serialized["url"] = get_source_url(...)
  return serialized
```

fbertsch: That way we only need to iterate over the values once, and it makes it obvious where any future changes to the definitions should live.

Iinh: Ah this made a lot of sense! Thank you for explaining!

tests/test_metrics_parser.py
```diff
@@ -21,3 +23,21 @@ def test_metrics_parser():
...
+
+def test_source_url():
+    parser = GleanMetricsParser()
+    parsed_metrics, errs = parser.parse(
+        ["tests/resources/metrics.yaml"], {}, "test.com/foo", "tests"
+    )
+
+    assert (
+        parsed_metrics["example.duration"]["source_url"]
+        == "test.com/foo/blob/tests/resources/metrics.yaml#L4"
```
wlach, line 36: I assume in production, the `repo_url` part includes a `https://` or similar? If it doesn't, maybe we should add it in `get_source_url`? If it does, maybe we should change our example `repo_url` to be more realistic.

wlach (changes requested):
Code looks really good. I just have one last question / nit that I noticed when skimming the tests.

tests/test_metrics_parser.py
Iinh: Yes the `repo_url` from `repositories.yaml` include `https://`. I've updated the test URL to include `https://`

wlach (approved):
💯

fbertsch (approved):
LGTM, thanks for sticking with it @Iinh!

PR #274, opened 2021-03-01
Add a canonical app description field for v2

wlach:
This is derived from the high-level "description" field and
should be applicable across any variants of an application.

We want this for mozilla/glean-dictionary#426

As it stands the front page looks like this because we don't expose this information in the /v2/ API:

image

(note the weird titles for each of the app groupings -- we should put what's currently the description there (e.g. "Firefox for Android") and put in something longer form for the description (which this PR would provide).

### DIFF
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -175,8 +175,8 @@ class RepositoriesParser(object):
             app.pop("app_name")
             app.pop("canonical_app_name", None)
             app.pop("bq_dataset_family")
-            app["description"] = app["app_description"]
-            app.pop("app_description", None)
+            app_description = app.pop("app_description", None)
+            app["description"] = app.get("description", app_description)
             namespace = app.pop("document_namespace")
             app["app_id"] = namespace
             repos[v1_name] = app
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -144,6 +144,7 @@ class RepositoriesParser(object):
                     "additional_dependencies", []
                 )
                 listing = {**app, **channel}
+                listing["canonical_app_description"] = app["description"]
                 listing["dependencies"] = dependencies
                 app_id = listing["app_id"]
                 listing["document_namespace"] = (
@@ -174,6 +175,7 @@ class RepositoriesParser(object):
             v1_name = app.pop("v1_name")
             app.pop("app_name")
             app.pop("canonical_app_name", None)
+            app.pop("canonical_app_description", None)
             app.pop("bq_dataset_family")
             namespace = app.pop("document_namespace")
             app["app_id"] = namespace
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -144,7 +144,6 @@ class RepositoriesParser(object):
                     "additional_dependencies", []
                 )
                 listing = {**app, **channel}
-                listing["canonical_app_description"] = app["description"]
                 listing["dependencies"] = dependencies
                 app_id = listing["app_id"]
                 listing["document_namespace"] = (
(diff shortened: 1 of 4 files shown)


wlach:
@jklukas does this seem like approximately the right approach?

probe_scraper/parsers/repositories.py
```diff
@@ -144,6 +144,7 @@ def parse_v2(self, filename=None) -> dict:
                     "additional_dependencies", []
                 )
                 listing = {**app, **channel}
+                listing["canonical_app_description"] = app["description"]
```
jklukas, line 147: It sounds good to distinguish `canonical_app_description` from description. I think we should update the content of `repositories.yaml` to rename the app-level "description" field to "canonical_app_description" to make this more explicit and avoid having to do this manipulation here.

jklukas:
Yes, this looks like a reasonable approach. Apologies for not commenting sooner.

probeinfo_api.yaml
```diff
@@ -481,7 +483,17 @@ components:
...
-      description: A brief free-text description of the repository
+      description: A brief free-text description of this application variant.
+
+    CanonicalAppDescription:
+      type: string
+      description: |
+        A brief free-text description of an application, applicable across any variants.
+
+        This should not be materialized in ETL, but rather only made available in user-facing views since it may be verbose
+        and it may change over time.
```
fbertsch, line 494: Are these flipped? Something that changes over time and is used in the user-facing view sounds like `Description`. A canonical description sounds like it should be as static, and terse, as possible.

jklukas: The "canon" as to what an app is called and how we describe it may change over time. This field is intended to always reflect the current state of how we think about the application.

In contrast, the `app_name` is a value that must stay stable over time because it's used as a technical hook for filtering.

That said, the `canonical` may not be necessary here. `app_description` would be sufficient to disambiguate from the channel-level `description`. I introduced `canonical_app_name` as a complement to the technical `app_name`. Perhaps it would be obvious to users that a description field is not something that should be materialized in ETL.

wlach: I like the sound of just using `app_description` everywhere. I'll give that a try.

probe_scraper/parsers/repositories.py
```diff
@@ -175,6 +175,8 @@ def _v2_to_v1(self, filename):
             app.pop("app_name")
             app.pop("canonical_app_name", None)
             app.pop("bq_dataset_family")
+            app["description"] = app["app_description"]
+            app.pop("app_description", None)
```
jklukas, line 179:
```suggestion
            app_description = app.pop("app_description", None)
            app["description"] = app.get("description", app_description)
```

To maintain current behavior for v1, I think we want to maintain "description" if it exists at the listing level, and only insert app_description when "description" is otherwise absent.

jklukas (approved):
r+wc

probe_scraper/parsers/repositories.py
wlach: Right, can't believe I missed that!

probeinfo_api.yaml
fbertsch: Thanks for clarifying, @jklukas. That makes sense now.

probeinfo_api.yaml
```diff
@@ -481,7 +483,17 @@ components:
 
     Description:
       type: string
-      description: A brief free-text description of the repository
+      description: A brief free-text description of this application variant.
```
fbertsch, line 486: Can we indicate this is optional now?

fbertsch (approved):

probeinfo_api.yaml
wlach: I'm not sure if we want to do this. I'd encourage people to fill this field out, as it provides useful documentation that we can expose in the Glean Dictionary and possibly elsewhere.

wlach: @fbertsch wdyt? ^

wlach: Talked with Frank, just going to merge this. Can always followup later.

PR #278, opened 2021-03-09
Library names

wlach:
This is a proof of concept for specify library names in a way that is unique per-repository. The idea is that we could then put this parameter into the metrics themselves, so we can route edits to the metric annotations repository correctly (i.e. an edit to a metric annotation for glean-core should be under the glean-core namespace, irrespective of which application you're getting commentary on). See also mozilla/glean-dictionary#457.

It may also be interesting/useful to expose other metadata about libraries in the Glean Dictionary. Not sure if that belongs in this PR or in a followup. At least this gets the ball rolling.

@jklukas @fbertsch curious if you have any early feedback on this approach. Consider the edits to `repositories.yaml` a mini-proposal of sorts.

### DIFF
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -168,13 +168,16 @@ class RepositoriesParser(object):
         for lib in repos_v2["libraries"]:
             variants = lib.pop("variants")
             for variant in variants:
-                newlib = {**lib, **variant}
-                newlib.pop("library_name")
-                newlib["library_names"] = [newlib["dependency_name"]]
-                newlib.pop("dependency_name")
-                v1_name = newlib.pop("v1_name")
-                newlib["app_id"] = v1_name
-                repos[v1_name] = newlib
+                lib_info = {**lib, **variant}
+                v1_name = lib_info["v1_name"]
+                lib_info["library_names"] = [lib_info["dependency_name"]]
+                lib_info["app_id"] = v1_name
+                
+                del lib_info["library_name"]
+                del lib_info["dependency_name"]
+                del lib_info["v1_name"]
+                
+                repos[v1_name] = lib_info
         for app in repos_v2["applications"]:
             app_channel = app.pop("app_channel", None)
             if app_channel is not None:
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -172,11 +172,9 @@ class RepositoriesParser(object):
                 v1_name = lib_info["v1_name"]
                 lib_info["library_names"] = [lib_info["dependency_name"]]
                 lib_info["app_id"] = v1_name
-                
                 del lib_info["library_name"]
                 del lib_info["dependency_name"]
                 del lib_info["v1_name"]
-                
                 repos[v1_name] = lib_info
         for app in repos_v2["applications"]:
             app_channel = app.pop("app_channel", None)
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -136,7 +136,8 @@ class RepositoriesParser(object):
             data = yaml.load(f, Loader=yaml.SafeLoader)
         model_validation.apply_defaults_and_validate(data, "RepositoriesYamlV2")
         repos = data
-        listings = []
+
+        app_listings = []
         for app in repos["applications"]:
             channels = app.pop("channels")
             for channel in channels:
@@ -154,19 +155,26 @@ class RepositoriesParser(object):
                 )
                 listing = remove_none(listing)
                 model_validation.validate_as(listing, "Application")
-                listings.append(listing)
+                app_listings.append(listing)
+
         return {
             "libraries": repos["libraries"],
-            "applications": listings,
+            "applications": app_listings,
         }
 
     def _v2_to_v1(self, filename):
         repos_v2 = self.parse_v2(filename)
         repos = {}
         for lib in repos_v2["libraries"]:
-            v1_name = lib.pop("v1_name")
-            lib["app_id"] = v1_name
-            repos[v1_name] = lib
+            variants = lib.pop("variants")
+            for variant in variants:
+                newlib = {**lib, **variant}
+                newlib.pop("library_name")
+                newlib["library_names"] = [newlib["dependency_name"]]
+                newlib.pop("dependency_name")
+                v1_name = newlib.pop("v1_name")
+                newlib["app_id"] = v1_name
+                repos[v1_name] = newlib
         for app in repos_v2["applications"]:
             app_channel = app.pop("app_channel", None)
             if app_channel is not None:
--- a/probeinfo_api.yaml
+++ b/probeinfo_api.yaml
@@ -649,7 +649,9 @@ components:
     DependencyName:
       type: string
       description: |
-        Name of this dependency, to be referenced by the dependencies section of an application's definition.
+        Build system-specific name of this dependency, to be referenced by the `dependencies` section
+        of an application's definition.
+      example: org.mozilla.components:browser-engine-gecko
 
     Dependencies:
       type: array
--- a/probeinfo_api.yaml
+++ b/probeinfo_api.yaml
@@ -649,7 +649,7 @@ components:
     DependencyName:
       type: string
       description: |
-        Name of this dependency name, to be referenced by dependencies by an application.
+        Name of this dependency, to be referenced by the dependencies section of an application's definition.
 
     Dependencies:
       type: array
--- a/probeinfo_api.yaml
+++ b/probeinfo_api.yaml
@@ -31,7 +31,7 @@ tags:
   - name: library
     x-displayName: The repositories.yaml Library Format
     description: |
-      <SchemaDefinition schemaRef="#/components/schemas/Library" exampleRef="#/components/examples/Library" />
+      <SchemaDefinition schemaRef="#/components/schemas/LibraryYaml" />
   - name: application
     x-displayName: The repositories.yaml Application Format
     description: |
@@ -321,7 +321,7 @@ components:
           type: array
           description: Repositories that define Glean dependencies
           items:
-            $ref: "#/components/schemas/Library"
+            $ref: "#/components/schemas/LibraryYaml"
         applications:
           type: array
           description: Repositories that define Glean applications
@@ -398,34 +398,26 @@ components:
                 items:
                   type: string
 
-    Library:
+    LibraryYaml:
       type: object
       additionalProperties: false
       required:
-        - v1_name
+        - library_name
         - description
         - notification_emails
         - url
-        - library_names
+        - variants
       properties:
-        v1_name:
-          $ref: "#/components/schemas/V1Name"
+        library_name:
+          $ref: "#/components/schemas/LibraryName"
         description:
           $ref: "#/components/schemas/Description"
-        deprecated:
-          $ref: "#/components/schemas/DeprecatedBool"
         notification_emails:
           $ref: "#/components/schemas/NotificationEmails"
         url:
           $ref: "#/components/schemas/RepoUrl"
-        branch:
-          $ref: "#/components/schemas/RepoBranch"
-        metrics_files:
-          $ref: "#/components/schemas/MetricsFiles"
-        ping_files:
-          $ref: "#/components/schemas/PingFiles"
-        library_names:
-          $ref: "#/components/schemas/LibraryNames"
+        variants:
+          $ref: "#/components/schemas/LibraryVariants"
 
     Application:
       type: object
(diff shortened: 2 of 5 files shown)


repositories.yaml
```diff
@@ -6,67 +6,54 @@ version: "2"
...
-      - org.mozilla.deprecated:glean
+    variants:
+      - v1_name: glean-core
+        description: Modern cross-platform telemetry (core library)
+        metrics_files:
+          - glean-core/metrics.yaml
+        ping_files:
+          - glean-core/pings.yaml
+        library_names:
+          - glean-core
```
fbertsch, line 23:
```suggestion
          - glean-core
      branch: main
```

fbertsch: Default is still `master`: https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/parsers/repositories.py#L39

repositories.yaml
```diff
@@ -6,67 +6,54 @@ version: "2"
...
+      - v1_name: glean-android
+        description: Modern cross-platform telemetry (Android-specific)
+        metrics_files:
+          - glean-core/android/metrics.yaml
```
fbertsch, line 27: These metrics are available to `glean-android` but not `glean-core`.

If my mental model is correct, libraries should be grouped as variants if they share _any_ metric definitions. The set of shared metrics is a subset of all metrics for a library variant.

We will run into problems if a variant has shared metrics with _two_ libraries, but I don't see that happening.

fbertsch (commented):
I haven't closely reviewed each repo, but in general this is an r+ from me. Not only does this keep related things together, but this is basically an internal-only API.

Along with this change we should publish libraries to the v2 endpoint, and rely on that for the `library_name` which should be shared among all variants.

jklukas:
## Should we have cascading dependencies?

The geckoview and glean situations feel different to me, and I'm not sure we should handle them both as variants nested within a single repo listing.

For geckoview, the type of nesting we're doing here seems like the only reasonable option. The three different channels correspond to different branches, they have to be scraped separately, and an app only depends on one of the three. But logically most of the metrics are going to be shared across all 3 branches, and it makes sense for glean-annotations to consider them all as one thing.

But for the glean libraries, we're looking at different files within a single branch of a single repository. If an app declares a dependency on `glean-android`, it's also implicitly depending on `glean-core`. It feels like these should be different top-level libraries, and we should understand that `glean-android` depends on `glean-core`. And it would then in glean-annotations, the android-specific metrics would get filed under the `glean-android` library_name and the core metrics would be under `glean-core`. That may be more difficult to achieve in a backwards-compatible way, though.

## library_name vs. library_names

Within the scheme proposed here, my only major concern is that we have an existing field called `library_names` that is linked to the `dependencies` defined for apps. We're introducing a new `library_name` field which is an entirely different concept, which will be confusing. If we pursue this course, I'd recommend renaming the existing `library_names` to something like `dependency_names` in the yaml, and then adding some additional v2 to v1 shim code.

wlach:
Separating out `glean-android` makes logical sense. I can give it a try, it looks like it only defines "special" metrics which get included with all pings:

https://github.com/mozilla/glean/blob/main/glean-core/android/metrics.yaml

Agreed that `library_name` vs. `library_names` is confusing. I'll try renaming `library_names` to `dependencies` and see how it goes.

wlach:
> > I'll try renaming library_names to dependencies and see how it goes.
> 
> Potentially, I think we want a given library to be able to have `dependencies` on other libraries. So `glean-android` would have `dependencies: [glean-core]` but it also needs to be able to declare which dependency identifiers it is defining. So we need a key with a different name that makes it clear that it's providing the handles that other entities in the file can use to tie back to it. Perhaps `defines_dependencies: [org.mozilla.components:service-glean]`.

I don't think this is strictly required, though it might be interesting or useful information to surface in the Glean Dictionary.

Do the long and fancy library names like `org.mozilla.components:service-glean` have any meaning or use elsewhere in the pipeline? Or should I feel free to rename them?

jklukas:
> Do the long and fancy library names like org.mozilla.components:service-glean have any meaning or use elsewhere in the pipeline? Or should I feel free to rename them?

They don't have meaning to the pipeline, but they have meaning to the applications. Per https://mozilla.github.io/probe-scraper/#tag/library

> All the build-specific library names defined in this repository, used as keys to match with the dependencies values declared by applications. A single repository could present multiple library names for different platforms, thus there may be multiple values here.

So `org.mozilla.components:service-glean` is the name of the dependency as declared in an Android application.

fbertsch:
> Potentially, I think we want a given library to be able to have `dependencies` on other libraries. So `glean-android` would have `dependencies: [glean-core]` but it also needs to be able to declare which dependency identifiers it is defining. So we need a key with a different name that makes it clear that it's providing the handles that other entities in the file can use to tie back to it. Perhaps `defines_dependencies: [org.mozilla.components:service-glean]`.

We originally wanted to go this route, but it ended up being too much work for downstream consumers (specifically MSG). Unless we also update this API to include all dependent metrics, I'm against making this change just now.

wlach:
> We originally wanted to go this route, but it ended up being too much work for downstream consumers (specifically MSG). Unless we also update this API to include all dependent metrics, I'm against making this change just now.

I agree with this. In terms of outputs to the glean dictionary there is no benefit to doing this now, let's defer.

jklukas:
> In terms of outputs to the glean dictionary there is no benefit to doing this now, let's defer.

That's fine by me. My primary concern is to make sure the names we're giving to the fields here are clear about their purpose and don't preclude future work like this. We should reserve `dependencies` on a library for that potential future use, and not use that name for what we currently call `library_names`.

wlach:
Ok, so I think I've got the new `repositories.yaml` format right, and I believe we should be producing backwards compatible v1 output. Wouldn't mind a bit of feedback on what's here now along with some thoughts on what I should do next: I guess the two options are:

1. To augment the existing repositories endpoint with this library_name metadata. This is relatively straightforward and would let us move forward with the annotations work (since library metrics could be assigned this unique identifier).
2. To go whole hog and implement `/v2/` metric and ping endpoints with this information.

Writing this out, (1) seems to make the most sense to me, but open to other viewpoints.

I'll be out until Wednesday next week, so no rush on this.

probe_scraper/parsers/repositories.py
```diff
@@ -154,19 +155,26 @@ def parse_v2(self, filename=None) -> dict:
...
-            repos[v1_name] = lib
+            variants = lib.pop("variants")
+            for variant in variants:
+                newlib = {**lib, **variant}
+                newlib.pop("library_name")
+                newlib["library_names"] = [newlib["dependency_name"]]
+                newlib.pop("dependency_name")
+                v1_name = newlib.pop("v1_name")
+                newlib["app_id"] = v1_name
+                repos[v1_name] = newlib
```
fbertsch, line 177: I don't think we have any coding styles I can fall back on here, but something about this felt like it was hard to parse. The following may be easier:
```suggestion
                lib_info = {**lib, **variant}
                v1_name = lib_info["v1_name"]
                lib_info["library_names"] = [lib_info["dependency_name"]]
                lib_info["app_id"] = v1_name
                
                del lib_info["library_name"]
                del lib_info["dependency_name"]
                del lib_info["v1_name"]
                
                repos[v1_name] = lib_info
```

probeinfo_api.yaml
```diff
@@ -622,12 +646,17 @@ components:
         If not specified, retention will be unlimited.
       type: integer
 
+    DependencyName:
+      type: string
+      description: |
+        Name of this dependency name, to be referenced by dependencies by an application.
```
fbertsch, line 652:
```suggestion
        Name of this dependency, to be referenced by the dependencies section of an application's definition.
```

probe_scraper/parsers/repositories.py
fbertsch: Ah, the rest of this file uses `pop`. Feel free to disregard

fbertsch (changes requested):
@wlach how is the glean-dictionary going to know different libraries are variants? It doesn't look like this pushes to v2, and v1 is dropping `library_name`, which was the common identifier for the lib across variants.

I would like to see this push a `/v2/glean/libraries` endpoint, which contains information across variants, but properly identifiers them (as we do channels).

probeinfo_api.yaml
```diff
@@ -604,6 +596,38 @@ components:
...
+            $ref: "#/components/schemas/Description"
+          deprecated:
+            $ref: "#/components/schemas/DeprecatedBool"
+          branch:
+            $ref: "#/components/schemas/RepoBranch"
+          metrics_files:
+            $ref: "#/components/schemas/MetricsFiles"
+          ping_files:
+            $ref: "#/components/schemas/PingFiles"
+          dependency_name:
```
jklukas, line 628: We do not currently have any libraries that define multiple dependencies, but it's my understanding that the intention of having `library_names` be an array is that we could have a glean library that is used by multiple platforms, and so would export multiple names for multiple platforms/build systems. So this should remain plural as `dependency_names` and be an array.

Does that match your understanding, @fbertsch ?

jklukas:
> I would like to see this push a /v2/glean/libraries endpoint, which contains information across variants, but properly identifiers them (as we do channels).

This approach makes sense to me. I don't think we need to invest in adding metrics endpoints for v2 yet, but it would make sense at this point to make a v2 endpoint for listing libraries.

We have `/v2/glean/app-listings` which gives the unnested view of all app channels. If we want to provide an unnested view of libraries, I propose that be called `/v2/glean/library-variants` to make it clear that it's unnested. If we want to expose nested views, then those would be called `applications` and `libraries`.

wlach:
> > I would like to see this push a /v2/glean/libraries endpoint, which contains information across variants, but properly identifiers them (as we do channels).
> 
> This approach makes sense to me. I don't think we need to invest in adding metrics endpoints for v2 yet, but it would make sense at this point to make a v2 endpoint for listing libraries.
> 
> We have `/v2/glean/app-listings` which gives the unnested view of all app channels. If we want to provide an unnested view of libraries, I propose that be called `/v2/glean/library-variants` to make it clear that it's unnested. If we want to expose nested views, then those would be called `applications` and `libraries`.

This is a good point. I'll take a stab at producing this as part of this PR.

probeinfo_api.yaml
fbertsch: Hmm, I'm trying to decide. This may be a leftover from the days of when we were going to try and resolve application dependencies from the repository itself, in which case we would need to know that a single library could be referred to in different ways.

However, we've not taken that approach, and I doubt we ever will. If that's the case we get to say what the dependency names are, and they are really only used by this file. We should then be fine to unite them into one.

jklukas: So it sounds like you're advocating for not having build-system-specific names at all. In which case the applications would list dependencies by `v1_name` (in which case we should not put v1 in the name, but rather use something like `library_variant_id`). This would be somewhat difficult from a backwards-compatibility standpoint, since we'd still need to maintain build-system-specific names to inject into the v1 output.

For the scope of this PR, then, maybe it's fine to keep `dependency_name` as it is defined here.

probe_scraper/parsers/repositories.py
wlach: Your way looks better, I agree. May as well take it.

wlach:
I think this is ready for review now.

probeinfo_api.yaml
```diff
@@ -622,12 +646,17 @@ components:
         If not specified, retention will be unlimited.
       type: integer
 
+    DependencyName:
+      type: string
+      description: |
+        Name of this dependency, to be referenced by the dependencies section of an application's definition.
```
jklukas, line 652:
```suggestion
      description: |
        Build system-specific name of this dependency, to be referenced by the `dependencies` section
        of an application's definition.
      example: org.mozilla.components:browser-engine-gecko
```

jklukas (approved):
r+

Please make sure to run this locally in its final form and verify that the output matches the v1 repositories endpoint before merging.

jklukas:
Diff looks reasonable. Thanks for the validation.

wlach:
Verified that the output of mozilla-schema-generator is the same too, per @fbertsch's request. Just swapped out the repositories.json used, using this technique:

```diff
diff --git a/mozilla_schema_generator/glean_ping.py b/mozilla_schema_generator/glean_ping.py
index b7e9ffa..b31e4ac 100644
--- a/mozilla_schema_generator/glean_ping.py
+++ b/mozilla_schema_generator/glean_ping.py
@@ -25,7 +25,7 @@ class GleanPing(GenericPing):
     )
     probes_url_template = GenericPing.probe_info_base_url + "/glean/{}/metrics"
     ping_url_template = GenericPing.probe_info_base_url + "/glean/{}/pings"
-    repos_url = GenericPing.probe_info_base_url + "/glean/repositories"
+    repos_url = "http://localhost:8000/repositories.json"
     dependencies_url_template = (
         GenericPing.probe_info_base_url + "/glean/{}/dependencies"
     )
```

fbertsch (approved):
Thank you for confirming that Will!

PR #279, opened 2021-03-10
Bug 1696303 - Add bergamot app

fbertsch:
Still waiting on feedback about the app_id.

@wlach will this specific channels usage cause issues for the glean dictionary?

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -519,9 +519,9 @@ applications:
       - epavlov@mozilla.com
     branch: main
     metrics_files:
-      - extension/ts/background-scripts/background.js/telemetry/metrics.yaml
+      - src/core/ts/background-scripts/background.js/telemetry/metrics.yaml
     ping_files:
-      - extension/ts/background-scripts/background.js/telemetry/pings.yaml
+      - src/core/ts/background-scripts/background.js/telemetry/pings.yaml
     dependencies:
       - glean-js
     retention_days: 180
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -499,7 +499,7 @@ applications:
 
   - app_name: bergamot
     canonical_app_name: Bergamot Translator
-    app_description: Web extension for on-device machine translation in Firefox
+    app_description: Web extension for on-device machine translation
     url: https://github.com/mozilla-extensions/bergamot-browser-extension
     notification_emails:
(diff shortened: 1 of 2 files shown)


brizental:
> @brizental is glean-core the correct dependency for gleanjs?

Hm, good point. I don't think so. I'll send a PR adding `glean-js` to the libraries list.

repositories.yaml
```diff
@@ -496,3 +496,25 @@ applications:
...
+    branch: main
+    metrics_files:
+      - extension/ts/background-scripts/background.js/telemetry/metrics.yaml
+    ping_files:
+      - extension/ts/background-scripts/background.js/telemetry/pings.yaml
+    dependencies:
+      - glean-core
+    retention_days: 180
+    channels:
+      - v1_name: bergamot
```
wlach, line 515: We should add descriptions for what these variants are.

eu9ene: bergamot - the extension is Implemented as an infobar, integrated into Firefox

bergamot-cross-browser - the extension is Implemented as a popup, can be installed in Firefox or Chrome

repositories.yaml
```diff
@@ -496,3 +496,25 @@ applications:
...
+        app_id: org.mozilla.bergamot
+        channel: firefox
+      - v1_name: bergamot-cross-browser
```
eu9ene, line 518: @fbertsch now I'm thinking that if those are channels, it would make sense to split this one into `cross-browser-firefox` and `cross-browser-chrome`. Also gleans.js doesn't add browser name in client info at the moment, so it would help us to distinguish them.

fbertsch: That makes sense to me, if it's not a problem on the client end. I'm assuming this hasn't shipped to anyone yet then?

eu9ene: No, I'm just clicking locally to send some data) I already changed it on the client side.  It will be shipped very soon though

repositories.yaml
```diff
@@ -496,3 +496,25 @@ applications:
...
+        app_id: org.mozilla.bergamot.cross-browser
```
eu9ene, line 519: @fbertsch let's drop cross browser one completely for now. We still have internal discussions about channels of distribution. We will add them later.

brizental (approved):

PR #282, opened 2021-03-16
Require unique channel names

### LINKED ISSUE #281: Require unique non-deprecated channel names
For all non-deprecated channels, channel names should be unique. We can add this check to the repositories.yaml check done during CI.

fbertsch:
Fixes #281


relud (approved):
r+wc

Issue #283, opened 2021-03-18
Incorrect "duplicate metrics identifier" checks today (2021-03-18)


badboy:
We just got a whole bunch of emails notifying us:

> Glean: Duplicate metric identifiers detected

Pretty much for any Glean-containing app I think.

Content:

```
Glean has detected duplicated metric identifiers coming from the product 'burnham'.

- 'glean.baseline.duration' defined more than once in glean-android, glean-core
- 'glean.baseline.locale' defined more than once in glean-android, glean-core
- 'os' defined more than once in glean-android, glean-core
- 'os_version' defined more than once in glean-android, glean-core
- 'device_manufacturer' defined more than once in glean-android, glean-core
- 'device_model' defined more than once in glean-android, glean-core
- 'architecture' defined more than once in glean-android, glean-core
- 'client_id' defined more than once in glean-android, glean-core
- 'app_build' defined more than once in glean-android, glean-core
- 'app_display_version' defined more than once in glean-android, glean-core
- 'app_channel' defined more than once in glean-android, glean-core
- 'first_run_date' defined more than once in glean-android, glean-core
- 'glean.error.invalid_value' defined more than once in glean-android, glean-core
- 'glean.error.invalid_label' defined more than once in glean-android, glean-core
- 'glean.error.preinit_tasks_timeout' defined more than once in glean-android, glean-core
```

burnham is a Python tool. It should not depend on `glean-android` at all.
Maybe #278 broke something? (That landed half a day ago, which is why I suspect it)

badboy:
checks are done here: https://github.com/mozilla/probe-scraper/blob/master/probe_scraper/glean_checks.py#L62

jklukas:
It does sound likely this is related to #278 

cc @wlach

wlach:
I am pretty sure this is a false positive, probably due to this bug:

https://github.com/mozilla/probe-scraper/blob/0ddfc78d039a5e2c7b5622156cfda5daa2b75bf9/probe_scraper/runner.py#L285

I think we're incorrectly picking up the metrics from `glean-core` in `glean-android` since we're matching on files that look like metrics.yaml, rather than what's prescribed in the repositories file. This was hidden before because we didn't seperate out the dependencies.

That said, reverting #278 may be the right solution in the short term.

wlach:
[...]

PR #284, opened 2021-03-18
Revert "Library names"

jklukas:
Reverts mozilla/probe-scraper#278 to avoid the failure reported in https://github.com/mozilla/probe-scraper/issues/283


wlach (approved):
I'm ok with reverting this while we figure out a longer-term solution. I'm not sure if I'll have time to fix this properly today.

PR #286, opened 2021-03-19
Failing dryrun demo

wlach:
This PR should fail CI, reproducing #283. Intended to demonstrate new
dryrun check.

PR #287, opened 2021-03-19
Add a dryrun job which scrapes burnham + dependencies

wlach:
This can be completed in a reasonable amount of time and will
give us more confidence that a change will work in production.

You can see an example of it failing in #286.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,9 @@ check-repos:
 test: build
 	docker-compose run app pytest tests/ --run-web-tests
 
+# For this test, we scrape glean-deprecated so we can also test the dependency resolution logic
+# (burnham depends on glean-deprecated) which we use to validate against duplicate metrics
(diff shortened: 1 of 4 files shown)


fbertsch (approved):
LGTM. Will and I discussed offline, and `glean-deprecated` will be removed in libraries v2.

Makefile
```diff
@@ -48,6 +48,9 @@ check-repos:
 test: build
 	docker-compose run app pytest tests/ --run-web-tests
 
+burnham-dryrun:
```
wlach, line 51:
```suggestion
# For this test, we scrape glean-deprecated so we can also test the dependency resolution logic
# (burnham depends on glean-deprecated) which we use to validate against duplicate metrics
# and failed in mozilla/probe-scraper#283
burnham-dryrun:
```

PR #295, opened 2021-03-24
Add an index file to the root of the deploy

wlach:
This will help people understand how to get their bearings when
they reach the URL. It may also assist with making the netlify
deploy work.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,11 +1,11 @@
 [[headers]]
-  for = "/index.html"
-
-  [headers.values]
-    Content-Type = "text/html"
-
-[[headers]]
   for = "/*"
 
   [headers.values]
     Content-Type = "application/json"
+
+[[headers]]
+  for = "/"
+
+  [headers.values]
+    Content-Type = "text/html"
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,11 +1,11 @@
 [[headers]]
-  for = "/*"
-
-  [headers.values]
-    Content-Type = "application/json"
-
-[[headers]]
   for = "/index.html"
 
   [headers.values]
     Content-Type = "text/html"
+
+[[headers]]
+  for = "/*"
+
+  [headers.values]
+    Content-Type = "application/json"
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,6 @@
+[[headers]]
+  # Define which paths this specific [[headers]] block will cover.
+  for = "^(?!\/index\.html).*$"
+
+  [headers.values]
+    content-type = "application/json"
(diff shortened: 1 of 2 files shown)


probe_scraper/runner.py
```diff
@@ -101,6 +101,13 @@ def write_moz_central_probe_data(probe_data, revisions, out_dir):
 
 def write_general_data(out_dir):
     dump_json(general_data(), out_dir, "general")
+    open(os.path.join(out_dir, "index.html"), 'w').write(
+        "<html><head><title>Mozilla Probe Info</title></head>"
+        "<body>This site contains metadata used by Mozilla's data collection "
+        "infrastructure, for more information see "
+        "<a href=\"https://mozilla.github.io/probe-scraper/\">the generated documentation</a>."
+        "</body></html>"
+    )
```
jklukas: To be more idiomatic, I'd like to see this expressed as a context manager and a multiline string:

```
with open(os.path.join(out_dir, "index.html"), 'w') as f:
    f.write("""
    """)
```

I don't see any particular reason to worry about indentation in the output file.

jklukas (approved):
I am very much in favor of this idea! Left some suggestions for code style.

wlach:
Ok, managed to get the content-type headers working right too, I think. @jklukas could you take a look and tell me if anything looks awful? I think this will be a big win for evaluating larger scale changes to the output like you and I have done recently.

https://deploy-preview-295--probe-scraper-dev.netlify.app/
https://deploy-preview-295--probe-scraper-dev.netlify.app/v2/glean/app-listings

netlify.toml
```diff
@@ -0,0 +1,11 @@
+[[headers]]
+  for = "/*"
+
+  [headers.values]
+    Content-Type = "application/json"
```
jklukas, line 5: This wildcard means that even invalid pages are interpreted as JSON.

If I hit this bad path: https://deploy-preview-295--probe-scraper-dev.netlify.app/v2/glean/foo

It returns HTML for "Page Not Found", but it's interpreted as JSON, so the page I get says "SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data" which is confusing.

I don't see this as a blocking issue.

jklukas (approved):
r+wc

netlify.toml
wlach: Yup, I noticed this as well. It's a bit annoying, but the `for:` clauses seem not to support regular expressions. I think the current behaviour is ok, esp. since this isn't a production site or anything.

probe_scraper/runner.py
wlach: Just realized I only half fixed this, will do the other half before landing.

PR #302, opened 2021-04-09
Add nimbus library to repositories.yaml

travis79:
Adding the nimbus metrics.yaml to the repositories

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -145,6 +145,7 @@ libraries:
       - v1_name: glean-js
         branch: main
         dependency_name: glean-js
+
   - library_name: nimbus
     description: Rapid experimentation platform
     notification_emails:
@@ -198,6 +199,7 @@ applications:
       - org.mozilla.appservices:logins
       - org.mozilla.components:support-migration
       - org.mozilla.components:places
+      - org.mozilla.appservices:nimbus
     channels:
       - v1_name: firefox-android-release
         app_id: org.mozilla.firefox
@@ -255,6 +257,7 @@ applications:
       # does not actually depend on glean-android, including for backwards
       # compat
       - org.mozilla.components:service-glean
+      - nimbus
     channels:
       - v1_name: firefox-ios-release
         app_id: org.mozilla.ios.Firefox
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -145,6 +145,18 @@ libraries:
       - v1_name: glean-js
         branch: main
         dependency_name: glean-js
+  - library_name: nimbus
+    description: Rapid experimentation platform
+    notification_emails:
+      - tlong@mozilla.com
+      - sync-team@mozilla.com
+    url: https://github.com/mozilla/application-services
+    metrics_files:
+      - components/nimbus/metrics.yaml
+    variants:
+      - v1_name: nimbus
+        branch: main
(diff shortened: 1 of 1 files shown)


tdsmith:
Should this also be added as a dependency for fenix and firefox_ios?

travis79:
> Should this also be added as a dependency for fenix and firefox_ios?

Probably, I'll fix that on Monday.

repositories.yaml
```diff
@@ -186,6 +199,7 @@ applications:
       - org.mozilla.appservices:logins
       - org.mozilla.components:support-migration
       - org.mozilla.components:places
+      - org.mozilla.appservices:nimbus
```
fbertsch, line 202: Assuming you want `org.mozilla.components`
```suggestion
      - org.mozilla.components:nimbus
```

repositories.yaml
```diff
@@ -243,6 +257,7 @@ applications:
       # does not actually depend on glean-android, including for backwards
       # compat
       - org.mozilla.components:service-glean
+      - nimbus
```
fbertsch, line 260:
```suggestion
      - org.mozilla.components:nimbus
```

fbertsch (approved):
r+wc. The entry in `dependencies` needs to match the `dependency_name`.

repositories.yaml
travis79: This was a bit of a mystery to me, nimbus has moved to the appservices repo now and it soon may only have a shell in android-components, but that is the place we want it to be consumed from.

fbertsch: Honestly it doesn't matter, you could just call it `nimbus` and it would be fine. We just use these strings for dependency lookups downstream from the probe-info-service.

fbertsch: "It doesn't matter" in that _what_ the string is doesn't matter to anyone outside the probe-info-service. It still needs to match between `dependency_name` and `dependencies`.

travis79: So it should probably be `org.mozilla.components:service-nimbus

repositories.yaml
travis79: This is iOS, does it need the android library dependency listed here like Glean does?

repositories.yaml
travis79: > "It doesn't matter" in that _what_ the string is doesn't matter to anyone outside the probe-info-service. It still needs to match between `dependency_name` and `dependencies`.

Gotcha, let's keep it simple and just `nimbus` then, if that's okay with you?

wlach: FWIW the glean dictionary now displays some dependency information (and will possibly soon display more) so it might be good to keep any conventions used consistent: https://dictionary.protosaur.dev/apps/fenix/app_ids/org_mozilla_fenix

fbertsch: @wlach it might be good to list the friendly name of the dependencies rather than the dependency_name string, what do you think?

wlach: I think I'm onboard with that: we should think about this holistically in terms of what we want the UX in the glean dictionary to be. I suspect the key user story is helping people (likely data engineers and application developers) better understand where metrics for a given application are coming from. In that respect I think the metadata for the library itself (rather than the references to it in repositories.yaml) is most important.

PR #304, opened 2021-04-14
Bug 1704878 - Firefox background updater

wlach:
This is a prototype of how we might implement the background updater
as a variant of Firefox (imagine in your mind that we've changed
the schema to say "variants" instead of "channels" for applications).

### DIFF
--- a/probeinfo_api.yaml
+++ b/probeinfo_api.yaml
@@ -524,7 +524,7 @@ components:
 
     V1Name:
       type: string
-      pattern: "^[a-z][a-z0-9-]{0,29}$"
+      pattern: "^[a-z][a-z0-9-]{0,39}$"
       description: Identifier used in v1 URI paths that reference a specific repository
       example: firefox-android-beta
 
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -154,26 +154,34 @@ applications:
     url: https://github.com/mozilla/gecko-dev
     notification_emails:
       - chutten@mozilla.com
+    metrics_files:
+      - toolkit/components/glean/metrics.yaml
+    ping_files:
+      - toolkit/components/glean/pings.yaml
     dependencies:
       - glean-core
+      # does not actually depend on glean-android, including for backwards
+      # compat
+      - org.mozilla.components:service-glean
     channels:
       - v1_name: firefox-desktop
         app_id: firefox-desktop
-        metrics_files:
-          - toolkit/components/glean/metrics.yaml
-        ping_files:
-          - toolkit/components/glean/pings.yaml
-        additional_dependencies:
-          # does not actually depend on glean-android, including for backwards
-          # compat
-          - org.mozilla.components:service-glean
-      - v1_name: firefox-background-updater
-        app_id: firefox-background-updater
-        description: The desktop version of Firefox's background update tasks
-        metrics_files:
-          - toolkit/mozapps/update/update_metrics.yaml
-        ping_files:
-          - toolkit/mozapps/update/update_pings.yaml
+
+  - app_name: firefox_desktop_background_updater
+    canonical_app_name: Firefox for Desktop Background Updater
+    app_description: >-
+      The desktop version of Firefox's background update tasks. This is
+      considered a seperate application because it has its own client
+      id and sends its own version of the standard pings.
+    metrics_files:
+      - toolkit/mozapps/update/update_metrics.yaml
+    ping_files:
+      - toolkit/mozapps/update/update_pings.yaml
+    dependencies:
+      - glean-core
+    channels:
+      - v1_name: firefox-desktop-background-updater
+        app_id: firefox-desktop-background-updater
 
   - app_name: fenix
     app_description: Firefox for Android (Fenix)
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -173,6 +173,9 @@ applications:
       The desktop version of Firefox's background update tasks. This is
       considered a seperate application because it has its own client
       id and sends its own version of the standard pings.
+    url: https://github.com/mozilla/gecko-dev
+    notification_emails:
+      - chutten@mozilla.com
     metrics_files:
       - toolkit/mozapps/update/update_metrics.yaml
     ping_files:
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -184,7 +184,7 @@ applications:
       - glean-core
     channels:
       - v1_name: firefox-desktop-background-updater
-        app_id: firefox-desktop-background-updater
+        app_id: firefox.desktop.backgroundupdate
 
   - app_name: fenix
     app_description: Firefox for Android (Fenix)
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -180,10 +180,10 @@ applications:
       - v1_name: firefox-desktop
         app_id: firefox.desktop
 
-  - app_name: firefox_desktop_background_updater
-    canonical_app_name: Firefox for Desktop Background Updater
+  - app_name: firefox_desktop_background_update
+    canonical_app_name: Firefox for Desktop Background Update Task
     app_description: >-
-      The desktop version of Firefox's background update tasks. This is
+      Firefox Desktop's background update task. This is
       considered a separate application because it has its own client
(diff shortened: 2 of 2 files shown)


repositories.yaml
```diff
@@ -154,18 +154,26 @@ applications:
...
-      - toolkit/components/glean/pings.yaml
     dependencies:
       - glean-core
-      # does not actually depend on glean-android, including for backwards
-      # compat
-      - org.mozilla.components:service-glean
     channels:
       - v1_name: firefox-desktop
         app_id: firefox-desktop
+        metrics_files:
```
wlach, line 162: I am not sure if probe scraper currently supports this concept (metrics files per variant), but it should be straightforward to add if not.

repositories.yaml
```diff
@@ -154,18 +154,26 @@ applications:
...
+          - toolkit/components/glean/pings.yaml
+        additional_dependencies:
+          # does not actually depend on glean-android, including for backwards
+          # compat
+          - org.mozilla.components:service-glean
+      - v1_name: firefox-background-updater
+        app_id: firefox-background-updater
+        description: The desktop version of Firefox's background update tasks
+        metrics_files:
+          - toolkit/mozapps/update/update_metrics.yaml
```
jklukas, line 174: The Glean docs definitely talk about `metrics.yaml` files with that specific name, but I don't see anywhere in the docs that explicitly suggest that the file must be named that. In `probe_scraper/runner.py`, it looks like we do some filtering for file paths that end in `metrics.yaml`, which probably covers this case. We should probably break off an issue to discuss whether arbitrary file names should be supported.

jklukas:
Do we know if this new app_id is going to send any of the built-in Glean pings? Or is it just custom pings?

wlach:
Per discussion in #glean-dev, it sounds like the preference (which I reluctantly agree with) is to create a new application here, so as not to conflict with other work which assumes that application variants are channels. I'll update this PR.

wlach:
> Do we know if this new app_id is going to send any of the built-in Glean pings? Or is it just custom pings?

My understanding is that it would send the built-in pings

repositories.yaml
```diff
@@ -167,6 +167,22 @@ applications:
...
+    app_description: >-
+      The desktop version of Firefox's background update tasks. This is
+      considered a seperate application because it has its own client
+      id and sends its own version of the standard pings.
+    metrics_files:
+      - toolkit/mozapps/update/update_metrics.yaml
+    ping_files:
+      - toolkit/mozapps/update/update_pings.yaml
+    dependencies:
+      - glean-core
```
wlach, line 181: The [original bug suggests](https://bugzilla.mozilla.org/show_bug.cgi?id=1704878#c0) we should specify service-glean as a dependency, but that doesn't make a lot of sense to me for a new application. That functionality is android-specific and we only include it for desktop firefox for backwards compatibility reasons.

wlach:
@fbertsch feel free to take this over if you want, I mostly started this PR for the purposes of discussion

wlach:
Looks like CI's failing because the app id is too long, not sure if that's a bigquery-motivated limit or what. Will leave it to someone else to resolve it.

chutten:
I can confirm that this app may send built-in pings. Certainly the at-init "baseline" ping will be sent, which should be kept separate from Firefox Desktop's "baseline" ping tables to avoid being considered part of DAU.

chutten:
@wlach Anything I can do to help this along?

probeinfo_api.yaml
```diff
@@ -524,7 +524,7 @@ components:
 
     V1Name:
       type: string
-      pattern: "^[a-z][a-z0-9-]{0,29}$"
+      pattern: "^[a-z][a-z0-9-]{0,39}$"
```
wlach, line 527: @jklukas is this going to cause any problems downstream in the pipeline? Do you know where this "29 character limit" came from for this field?

wlach:
I think this should work. It includes:

* Updates to the app id and app name per @jklukas's suggestions
* Updates to the locations of the metric and ping files per https://bugzilla.mozilla.org/show_bug.cgi?id=1704878#c6

probeinfo_api.yaml
jklukas: I'm not aware of any reason this would cause a problem.

It looks like this was originally introduced in https://github.com/mozilla/probe-scraper/commit/6399763b7309f11268cfe418cc1ae80ed53a8831 where the intent was simply to make sure the entries were kebab-case. I assume @mdboom chose the character limit based on the general size of values that existed at the time.

repositories.yaml
jklukas: That sounds good to me.

repositories.yaml
```diff
@@ -180,6 +180,25 @@ applications:
...
+      - chutten@mozilla.com
+    metrics_files:
+      - toolkit/mozapps/update/metrics.yaml
+    ping_files:
+      - toolkit/mozapps/update/pings.yaml
+    dependencies:
+      - glean-core
+    channels:
+      - v1_name: firefox-desktop-background-updater
+        app_id: firefox.desktop.backgroundupdate
```
jklukas, line 200: [I now prefer `firefox.desktop.background.update`](https://mozilla.slack.com/archives/C4D5ZA91B/p1619033249269600?thread_ts=1619029486.253700&cid=C4D5ZA91B) so that we have a separator between `background` and `update` which matches the name of the ping (`background_update`).

repositories.yaml
```diff
@@ -180,6 +180,25 @@ applications:
       - v1_name: firefox-desktop
         app_id: firefox.desktop
 
+  - app_name: firefox_desktop_background_updater
+    canonical_app_name: Firefox for Desktop Background Updater
+    app_description: >-
+      The desktop version of Firefox's background update tasks. This is
```
fbertsch, line 186: Is there a non-desktop version of Firefox's background update task?

fbertsch (approved):
LGTM, thanks for putting this together and discussing @wlach!

repositories.yaml
```diff
@@ -180,6 +180,25 @@ applications:
       - v1_name: firefox-desktop
         app_id: firefox.desktop
 
+  - app_name: firefox_desktop_background_updater
```
jklukas, line 183: My preference would be to drop the trailing `r` everywhere for consistency.

For the canonical_app_name, we could say "Firefox for Desktop Background Update Task".

repositories.yaml
```diff
@@ -167,6 +167,25 @@ applications:
...
+      - chutten@mozilla.com
+    metrics_files:
+      - toolkit/mozapps/update/update_metrics.yaml
+    ping_files:
+      - toolkit/mozapps/update/update_pings.yaml
+    dependencies:
+      - glean-core
+    channels:
+      - v1_name: firefox-desktop-background-updater
+        app_id: firefox-desktop-background-updater
```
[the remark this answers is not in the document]
wlach: @fbertsch it looks like you're commenting on an old version of the PR, and you suggestion conflicts with @jklukas's advice above.

probeinfo_api.yaml
fbertsch: Indeed: https://github.com/mozilla/probe-scraper/pull/113#issuecomment-527902928

repositories.yaml
```diff
@@ -180,6 +180,25 @@ applications:
       - v1_name: firefox-desktop
         app_id: firefox.desktop
 
+  - app_name: firefox_desktop_background_updater
+    canonical_app_name: Firefox for Desktop Background Updater
+    app_description: >-
+      The desktop version of Firefox's background update tasks. This is
+      considered a seperate application because it has its own client
```
jklukas, line 187:
```suggestion
      considered a separate application because it has its own client
```

repositories.yaml
fbertsch: This matches :nalexander's comment in the bug. https://bugzilla.mozilla.org/show_bug.cgi?id=1704878#c4

repositories.yaml
fbertsch: Ah, maybe `Firefox Desktop's background update task.`?

jklukas (approved):

PR #307, opened 2021-04-21
Fix app_id for firefox.desktop

jklukas:
The normalized values will stay the same, so this has no effect on the pipeline.

But the intent is for this value to exactly match what's set by the application which was not the case here.

See https://searchfox.org/mozilla-central/rev/08013752b4638f01d41d2a38ca7bd741bc572c86/toolkit/components/glean/src/lib.rs#117

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -178,7 +178,7 @@ applications:
       - org.mozilla.components:service-glean
     channels:
       - v1_name: firefox-desktop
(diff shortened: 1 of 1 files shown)


jklukas:
I have confirmed in the deploy preview that both `document_namespace` and `bq_dataset_family` are unchanged by this.

wlach (approved):

PR #308, opened 2021-04-22
Bug 1706875 - Enable ingestion for rally.debug

Dexterp37:
This enables a temporary ingestion for the data collected with Glean from the Core Add-on through Glean.js


Dexterp37:
Hey @jklukas !

In [this proposal](https://docs.google.com/document/d/1Za-3Srfy2q3VnAnHoUMeBU0J-cEOTKS6EhnRZJz8q9A/edit#) we agreed on adding the `encryption` section to the metadata in probe-scraper so that we can set `use_jwk` to let the schema generator know about encryption.

Do I only need to change [this schema](https://github.com/mozilla/probe-scraper/blob/main/schemas/repositories.json)?

repositories.yaml
```diff
@@ -546,3 +546,23 @@ applications:
...
+    branch: master
+    metrics_files:
+      - metrics.yaml
+    ping_files:
+      - pings.yaml
+    dependencies:
+      - glean-js
+    retention_days: 180
+    channels:
+      - v1_name: rally
```
jklukas, line 565:
```suggestion
      - v1_name: rally-debug
```

I'm assuming `rally.core` will have `v1_name: rally-core`, so probably should be more specific here. The `v1_name` doesn't affect the pipeline, though. It just has to be unique.

repositories.yaml
```diff
@@ -546,3 +546,23 @@ applications:
...
+        app_id: rally.debug
+    encryption:
+      use_jwk: true
```
jklukas, line 568: Tests are currently failing because this is a new field here that we haven't built support for yet.

We'll have to add this to the JSON schema in `probeinfo_api.yaml` for `ApplicationYaml`. I don't have time before EOD to look back through the proposals and ensure this is the right thing to do. The pipeline is expecting the JSON schema for Rally pings to include `jwe_mappings` in `mozPipelineMetadata`. Did we end up deciding that we'd use probe-scaper metadata to configure this injection?  If so, changes will be needed in mozilla-schema-generator to pick up this field and respond to it, and we'll need to make it available in the probeinfo API.

jklukas: I see now that you gave a link to the relevant bit of the proposal. So yes, this is the right thing to do :) but we'll need to make some additional code changes here to support it. I can prioritize working on this on Monday.

acmiyaguchi (approved):
Looks good to me

jklukas (approved):
LGTM.

I'll leave it to @acmiyaguchi to decide when to merge this. If we merge it without `use_jwk` support in m-s-g, then I believe the tables will get generated correctly, but we'll probably see pipeline errors in the Rally job due to lack of encryption config being propagated there.

acmiyaguchi:
There will only be pipeline errors if documents get sent, which would be a good signal regardless since this pipeline hasn't seen action on the rally side of things yet. 

I'll rerun the probe scraper to pick up these changes, and make changes to the msg to add in the metadata fields appropriately.

PR #309, opened 2021-04-26
Add encryption to Application config

jklukas:
Necessary to support https://github.com/mozilla/probe-scraper/pull/308

This field was specific in the [Glean data encryption proposal](https://docs.google.com/document/d/1Za-3Srfy2q3VnAnHoUMeBU0J-cEOTKS6EhnRZJz8q9A/edit#) but never implemented. Once this is integrated, we'll also need to make a change to mozilla-schema-generator to act on this field and put the appropriate info in `mozPipelineMetadata`.

### DIFF
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -53,6 +53,7 @@ class Repository(object):
         self.dependencies = definition.get("dependencies", [])
         self.prototype = definition.get("prototype", False)
         self.retention_days = definition.get("retention_days", None)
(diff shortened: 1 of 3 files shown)


jklukas:
I tested locally that I could add `use_jwk: true`  to the burnham entry in repositories.yaml, and the Netlify preview showed it correctly appearing in the `/glean/repositories` endpoint output for burnham.

acmiyaguchi (approved):
Looks good to me, thanks for adding this to the application schema.

PR #313, opened 2021-05-17
Bug 1710406 - Add Rally Core Add-on Glean ingestion

Dexterp37:
See bug 1710406 for more context.

@acmiyaguchi can the `rally-debug` be deprecated? Do you expect us needing that for something else?

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -565,6 +565,7 @@ applications:
     channels:
       - v1_name: rally-debug
         app_id: rally.debug
+        deprecated: true
     encryption:
       use_jwk: true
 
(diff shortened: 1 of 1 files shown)


jklukas (approved):
LGTM

acmiyaguchi (approved):
Rally-debug can be deprecated, but I'd rather keep the project around since it can be handy to have a testing environment.

wlach:
> Rally-debug can be deprecated, but I'd rather keep the project around since it can be handy to have a testing environment.

@Dexterp37 Can you marked `rally-debug` as `deprecated: true` before landing? That should hide it by default in the Glean Dictionary.

repositories.yaml
```diff
@@ -567,3 +567,24 @@ applications:
         app_id: rally.debug
     encryption:
       use_jwk: true
+
+  - app_name: rally_core
+    canonical_app_name: Rally Core Add-on
+    app_description: |
+      The Rally Core Add-on orchestrates the installation
+      and the lifecycle of Rally studies.
```
wlach, line 575: Feels like it would be good idea to provide a link to Rally here, in case someone is looking at the Glean Dictionary and wondering what this is.

```suggestion
      and lifecycle of [Rally](https://rally.mozilla.org/) studies.
```

Issue #296, opened 2021-03-24
Publish applications to `v2/glean/applications`


fbertsch:
This would be useful for Looker, which pushes one-namespace per-application. The current app-variants listing is decomposed, and we have to find all the channels for every variant; if we switched to an `/applications` endpoint it would simplify that code.

PR #314, opened 2021-05-21
Consistently use app-listings and library-variants internally

jklukas:
This is a helpful step to enable adding endpoints for "applications"
and "libraries", which are nested views of this info.

It's currently confusing that we use "applications" for the flattened
data internally, but use app-listings for that concept in the API.

### DIFF
--- a/probe_scraper/parsers/repositories.py
+++ b/probe_scraper/parsers/repositories.py
@@ -167,14 +167,14 @@ class RepositoriesParser(object):
                 library_listings.append(listing)
 
         return {
-            "libraries": library_listings,
-            "applications": app_listings,
+            "library-variants": library_listings,
+            "app-listings": app_listings,
         }
 
     def _v2_to_v1(self, filename):
         repos_v2 = self.parse_v2(filename)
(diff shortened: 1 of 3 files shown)


wlach (approved):
LGTM. Note that @fbertsch has proposed a slightly different view on applications here: https://github.com/mozilla/probe-scraper/issues/296

jklukas:
> Note that @fbertsch has proposed a slightly different view on applications

Yes, it's my intention that a next step would be to have `parse_v2` add a new entry to its output dict that matches what's needed for https://github.com/mozilla/probe-scraper/issues/296

jklukas:
I think circleci may be having some issues preventing `build_and_test` from getting scheduled.

fbertsch (approved):
This is definitely the right first step to making things more consistent 👍

jklukas:
As I was waiting for tests to run, I realized that I had missed the the API definition layer. I've pushed another commit to use `AppListing` and `LibraryVariant` consistently within the yaml, and to reference by those names in the code.

PR #317, opened 2021-05-27
Bug 1712899 - Ignore MozillaVPN commits before the 26th of May 2021

Dexterp37:


fbertsch (approved):
r+, but the commit says ignore commits before 5/26, but this actually ignores commits before 5/25. I just want to confirm that you tested this ignores the correct range.

probe_scraper/scrapers/git_scraper.py
```diff
@@ -29,6 +29,7 @@
     "reference-browser": "2019-04-01 00:00:00",
     "firefox-desktop": "2020-07-29 00:00:00",
     "glean-js": "2020-09-21 13:35:00",
+    "mozilla-vpn": "2021-05-25 0:0:00",
```
fbertsch, line 32:
```suggestion
    "mozilla-vpn": "2021-05-25 00:0:00",
```

Dexterp37:
> 
> 
> r+, but the commit says ignore commits before 5/26, while this actually ignores commits before 5/25. I just want to confirm that you tested this ignores the correct range.

Yes, I wanted to make sure to include comments from the 26th onwards. The correct commit range is picked up :)

PR #318, opened 2021-06-07
Bug 1713416 - Add Rally Study 01 to the Glean ingestion

Dexterp37:
**DO NOT MERGE YET. There is a pending discussion: https://bugzilla.mozilla.org/show_bug.cgi?id=1713416**

TODO:

- [x] Check with the Rally team if they are cool with using the probe dictionary.
- [ ] ~Check with the Rally team if they are cool with the recommended app id.~ (Will be handled separately after a decision is made)

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -611,3 +611,27 @@ applications:
     channels:
       - v1_name: mozilla-vpn
         app_id: mozillavpn
+
+  - app_name: rally_study_zero_one
+    canonical_app_name: Rally Study-01
+    app_description: |
+      This is an exploratory study on what sites people
+      use and how long they use them, for the Rally
+      platform. It is implemented as a cross-browser
+      WebExtension.
+    url: https://github.com/mozilla-rally/rally-study-01
+    notification_emails:
+      - than@mozilla.com
+      - hulmer@mozilla.com
+    branch: master
+    metrics_files:
+      - metrics.yaml
+    ping_files:
+      - pings.yaml
+    dependencies:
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -611,3 +611,27 @@ applications:
...
+    branch: master
+    metrics_files:
+      - metrics.yaml
+    ping_files:
+      - pings.yaml
+    dependencies:
+      - glean-js
+    channels:
+      - v1_name: rally-study-zero-one
+        app_id: rally-study-zero-one
```
acmiyaguchi, line 635: I'll just note that `rally_zero_one` and `rally_study_zero_one` are close enough together to be very confusing. Why not `rally_zero_two`; is this the same study?

acmiyaguchi (approved):
This looks good to me, in terms of getting it scheduled into mozilla-pipeline-schemas.

repositories.yaml
Dexterp37: The study itself is called "Rally 01", so calling it "zero two" would be even more confusing :) However, we're apparently hitting a GCP limit with this name and we're [discussing potential solutions here](https://bugzilla.mozilla.org/show_bug.cgi?id=1713416).

repositories.yaml
```diff
@@ -613,3 +613,27 @@ applications:
...
+    encryption:
+      use_jwk: true
```
Dexterp37, line 639:
```suggestion
      use_jwk: true
    skip_documentation: true
```

Issue #301, opened 2021-04-07
Read mozilla-central's metrics_index instead of requiring manual addition of new definitions files


chutten:
[Presently we order](https://firefox-source-docs.mozilla.org/toolkit/components/glean/user/new_definitions_file.html#where-do-i-define-new-metrics-and-pings) Firefox on Glean consumers to manually file bugs whenever they add new `metrics.yaml` or `pings.yaml` files to `metrics_index.py`. 

This isn't the most ergonomic way to deal with this. Could we have probe-scraper pull https://hg.mozilla.org/mozilla-central/file/tip/toolkit/components/glean/metrics_index.py and read the list of yamls from there (like mozilla-central's build does)?

PR #321, opened 2021-06-14
bug 1715554 - Add desktop's `browser/base/content/metrics.yaml`

chutten:


jklukas (approved):
LGTM. I've tagged Frank for review too, in case he has any additional context or thoughts on metrics discovery for FOG.

fbertsch (approved):
We could support wildcards, or meta-metrics_files. @chutten it depends on how many there will be and how often they will change. This seems sufficient for now

chutten:
The bug for meta-metrics files is #301. I'm not sure how many `metrics.yaml` we'll end up with. If it's one per [module](https://wiki.mozilla.org/Modules/All) then we're talking dozens, though, so we should perhaps start thinking about #301 sooner rather than later.

PR #326, opened 2021-08-13
Bug 1725500 - Handle user needinfo blocked for probe expiry alert

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1725500

If needinfo fails during bug creation, retry with cc


scholtzan (approved):

BenWu:
Oops just realized I linked the wrong bug and have the wrong number in the commit.  Should be https://bugzilla.mozilla.org/show_bug.cgi?id=1725415

PR #327, opened 2021-08-25
Set mimetype of index.html correctly (fixes #310)

### LINKED ISSUE #310: `/` content type is `application/json`
We have a default HTML page for the root of `probeinfo.telemetry.mozilla.org`, but the content-type is `application/json`, which renders the HTML unreadable.

wlach:
Haven't tested this, but it should work, based on my reading of
the documentation.

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -424,7 +424,7 @@ def sync_output_and_cache_dirs(
                         with gzip.open(out_filename, "wb") as f2:
                             f2.write(f1.read())
 
-            # Synchronize the json files and index.html seperately,
+            # Synchronize the json files and index.html separately,
             # as they have different mimetypes
             sync_params = (
                 f"--content-encoding 'gzip' "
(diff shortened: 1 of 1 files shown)


probe_scraper/runner.py
```diff
@@ -424,15 +424,24 @@ def sync_output_and_cache_dirs(
...
+            sync_params = (
                 f"--content-encoding 'gzip' "
                 f"--cache-control 'max-age=28800' "
                 f"--acl public-read"
             )
-            os.system(sync_cft_cmd)
+            os.system(
+                f"aws s3 sync {tmpdirname}/ s3://{output_bucket}/ "
+                f"--delete "
+                f"--exclude index.html"
```
jklukas, line 437: There a missing space at the end here, so this probably becomes bleeds into the next flag: `index.html--content-type `

probe_scraper/runner.py
```diff
@@ -424,15 +424,24 @@ def sync_output_and_cache_dirs(
...
+                f"--content-type 'application/json' " + sync_params
```
jklukas, line 438: Adding `sync_params` here will inject it with a wrapper, probably like `("--content-encoding 'gzip'", ...`

wlach: Yes, AFAICT that doesn't cause a problem in my local testing.

PR #329, opened 2021-08-30
Fix marking whether metrics are in source code or not

### LINKED ISSUE #320: Mark metrics that are no longer present in the application source code
At present, we have no way of determining whether a metric is still present in the source code. This information would be useful in some downstream contexts (e.g. Glean Dictionary, Looker).

wlach:
In my previous approach, I was using a commit index (not a timestamp, as
I thought I was),  which is actually *lower* for the latest value. This
would work only for metrics with one history entry: other metrics would
always be marked as not "in source". The right solution is actually
really simple: just see if the metric is present in the latest commit
index ("0") -- if it is, it's in source. If it isn't, it's not. No need
to bother with timestamps at all.

Hopefully fixes #320 for good

### DIFF
--- a/probe_scraper/transform_probes.py
+++ b/probe_scraper/transform_probes.py
@@ -328,7 +328,6 @@ def update_or_add_item(
     item,
     definition,
     commit_timestamps,
-    last_timestamp,
     equal_fn,
     type_ctor,
 ):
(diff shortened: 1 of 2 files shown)


wlach:
Hmm, did some further testing and this doesn't quite seem to do the job for Fenix. Will dig further.

wlach:
> Hmm, did some further testing and this doesn't quite seem to do the job for Fenix. Will dig further.

False alarm, I did my longer test run against an older version of the source. Everything is indeed now working as expected, ready for review.

badboy (approved):

PR #331, opened 2021-08-31
Keep git checkouts across runs by default

badboy:
This also clones git repositories inside the specified cache folder.
Previously any pre-existing checkout was always removed.
That's fine on a nightly run, with a fast network, but is notoriously
slow on local developer machines for large repositories such as
gecko-dev itself.
Instead we can just keep that checkout around and just `pull` the latest
commit instead of doing a full fresh clone.
Then we also don't remove the checkout, so the next invocation will
reuse what's there.

---

Not sure who else currently would have the capacity to review this code.


probe_scraper/scrapers/git_scraper.py
```diff
@@ -65,14 +64,21 @@ def retrieve_files(repo_info, cache_dir):
...
 
-    if os.path.exists(repo_info.name):
-        shutil.rmtree(repo_info.name)
-    repo = git.Repo.clone_from(repo_info.url, repo_info.name)
+    repo = None
+    if os.path.exists(repo_path):
+        print(f"Pulling latest commits into {repo_path}")
+        repo = git.Repo(repo_path)
+        o = repo.remotes.origin
+        o.pull()
```
wlach, line 78:
```suggestion
        repo.remotes.origin.pull()
```

wlach (approved):
CC'ing @whd just as a head's up

I'd like to start running a more comprehensive set of integration tests for this repo (aside from just burnham + glean), things like this will potentially make this more tractable so long as we save the state between runs.

PR #334, opened 2021-09-02
Fix errors in syncing logic and update to use subprocess.check_call instead os.system

wlach:


probe_scraper/runner.py
```diff
@@ -434,7 +434,7 @@ def sync_output_and_cache_dirs(
             os.system(
                 f"aws s3 sync {tmpdirname}/ s3://{output_bucket}/ "
                 f"--delete "
-                f"--exclude index.html"
+                f"--exclude index.html "
                 f"--content-type 'application/json' " + sync_params
```
jklukas, line 438: This is probably also a problem that needs to be fixed. See https://github.com/mozilla/probe-scraper/pull/327/files#r701131309

wlach: Hmm, am pretty sure this is fine:

```
wlach@antwerp glean-dictionary % python3
Python 3.8.2 (default, Jun  8 2021, 11:59:35) 
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.system("echo " + ("foo"))
foo
0
```

badboy: Mayb to just circumvent the whole "manually adding spaces" problem, stuff all arguments into an array, then `" ".join()` that together?
(Doesn't `os.system` alternatively take an array to do this for you?)

jklukas: The `subprocess` module provides more comprehensive facilities, including taking in a list of args instead of a full string. I think we should transition this code to do that.

jklukas: > Hmm, am pretty sure this is fine

That example hits a python gotcha. `("foo")` with a single entry and no trailing comma is not interpreted as a tuple. The example fails if you use `("foo",)`:

```
>>> os.system("echo " + ("foo",))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate str (not "tuple") to str
```

wlach: Yeah, I was just thinking that. I'll this PR to use [check_call](https://docs.python.org/3/library/subprocess.html#subprocess.check_call)

jklukas (approved):

PR #335, opened 2021-09-02
Add mlhackweek_search repo 

gleonard-m:


repositories.yaml
```diff
@@ -682,3 +682,25 @@ applications:
...
+      - searchapp/mysite/searchranking/glean_config/metrics.yaml
+    ping_files:
+      - searchapp/mysite/searchranking/glean_config/pings.yaml
+    dependencies:
+      - glean-core
+    channels:
+      - v1_name: mlhackweek-search
+        app_id: mlhackweek-search
+    encryption:
+      use_jwk: true
```
travis79, line 705: Unless you are using the Glean `JweMetricType`, I don't think that you need this.

travis79 (approved):
Other than the extra `use_jwk` that I don't think is needed unless you are using the JweMetricType, this look good to me. No need to re-review once that is removed.

repositories.yaml
jklukas: These fields are all documented in the generated API docs: https://mozilla.github.io/probe-scraper/#tag/application

This one is nested quite a bit under that structure, but its description is:

> If set to true, the pipeline will treat the entire payload as a JSON Web Encryption (JWE) object in compact serialization, using a private JSON Web Key (JWK) to decrypt it.

So we want to leave off the `encryption` section. It's rally-specific.

gleonard-m: Removed.

travis79 (approved):

PR #338, opened 2021-09-09
[Bug 1729938] Revert "bug 1728642 - Add desktop's `gfx/metrics.yaml`"

scholtzan:
Since `gfx/metrics.yaml` got removed, we should also remove it here to make probe-scraper happy


scholtzan:
This won't work since schemas with the new probes have already been deployed. I added some more info in the bug

PR #339, opened 2021-09-09
Bug 1729938 - Handle deleted metrics.yaml

relud:


probe_scraper/scrapers/git_scraper.py
```diff
@@ -41,15 +41,32 @@
...
     log_format = '--format="%H{}%ct"'.format(sep)
-    change_commits = enumerate(repo.git.log(log_format, filename).split("\n"))
-    most_recent_commit = enumerate(repo.git.log("-n", "1", log_format).split("\n"))
-    commits = set(change_commits) | set(most_recent_commit)
+    # include "--" to prevent error for filename not in current tree
+    change_commits = repo.git.log(log_format, "--", filename).split("\n")
+    commits = set(enumerate(change_commits))
+    if _file_in_repo_head(repo, filename):
+        # include HEAD when it contains filename
+        commits |= set(enumerate(repo.git.log("-n", "1", log_format).split("\n")))
```
relud, line 69: This is needed because `SKIP_COMMITS` doesn't allow `HEAD`.

Also, this can't be used to also filter `change_commits` because for reasons I don't understand `_file_in_repo_head` doesn't work with `subtree = repo.tree(commit)` to check arbitrary commits.

scholtzan (approved):

probe_scraper/check_repositories.py
```diff
@@ -11,6 +11,7 @@
 REPOSITORIES = os.path.join(
     os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "repositories.yaml"
 )
+EXPECTED_MISSING_FILES = {("firefox-desktop", "gfx/metrics.yaml")}
```
Dexterp37, line 14: Should we document this somewhere?

How will this behave when the file will be added again? (it's missing because the related patch was backed out due to performance problems that are being worked on)

relud: it won't break anything, but this part should be removed when the file returns.

PR #343, opened 2021-09-29
bug 1724303 - Reorganize Gecko metrics into its own library.

chutten:
Metrics and Pings collected in the reusable parts of libxul will now (soon?) be
reported from all applications using the reusable parts of libxul.
(See bug 1670261) Specifically Fenix and Firefox Desktop, though I wouldn't be
surprised to see Focus and Klar (Android) included in the near future.


chutten:
This raises a question: how do we deal with adding new metrics.yamls to one variant at a time?

chutten:
@wlach I expect this `gecko` library to add new metrics files rapidly between now and the end of the year. Does this mean I should remove the variants and stick to just a single one that tracks nightly?

repositories.yaml
```diff
@@ -159,21 +159,37 @@ libraries:
...
         dependency_name: nimbus
 
-# See https://mozilla.github.io/probe-scraper/#tag/application
-applications:
-  - app_name: firefox_desktop
-    canonical_app_name: Firefox for Desktop
-    app_description: The desktop version of Firefox
-    url: https://github.com/mozilla/gecko-dev
+  - library_name: gecko
+    description: The browser engine developed by Mozilla
```
wlach, line 163: Can we mention something here about how this compares with the geckoview library? Ideally in the description for both.

repositories.yaml
```diff
@@ -226,14 +242,17 @@ applications:
         app_id: org.mozilla.firefox
         app_channel: release
         additional_dependencies:
+          # TODO: This is using beta and not release because...
```
wlach, line 245: Should we figure this out before landing?

wlach (approved):
Have some small questions, but in general LGTM

repositories.yaml
chutten: There's a geckoview library? Oh, you mean `engine-gecko`.

That's a tricky one to put into words. `engine-gecko` is currently being used to show which applications use [GeckoView Streaming Telemetry (EXTRACT)](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/internals/geckoview-streaming.html) (which is just Fenix). In the not-too-distant future we hope to migrate everyone one of those to be collected in Gecko and reported in all things that use both Glean and Gecko (removing the need for `engine-gecko` to exist in `repositories.yaml` entirely). In fact, Jan-Erik's already started that process.

So, in terms of `repositories.yaml`, `gecko`'s relationship with `engine-gecko` is one of temporary fraternity. They don't really have anything to do with each other, and we can hopefully remove `engine-gecko` from the file within a month.

This leaves me uncertain about how to write these. Do you have any ideas @wlach ?

repositories.yaml
chutten: I admit to hoping you'd know what's going on here, but I think @badboy found the answer in the code that this actually _should_ be the `-release` variant.

repositories.yaml
wlach: I see. If `engine-gecko` is going away soon, maybe we can just leave things as-is.

PR #358, opened 2021-10-27
Bug 1734011 - Scrape and expose Glean tags

wlach:


tests/test_git_scraper.py
```diff
@@ -231,9 +232,12 @@ def test_normal_repo(normal_repo):
     # duration same begin/end commits for first history entry
     assert len(set(metrics[duration][HISTORY_KEY][0][COMMITS_KEY].values())) == 1
 
-    # duration same begin/end commits for first history entry
+    # duration *different* begin/end commits for last history entry
```
wlach, line 235: This comment was wrong

tests/resources/test_repo_files/normal/1/metrics.yaml
```diff
@@ -9,7 +9,8 @@ example:
     send_in_pings:
       - baseline
     bugs:
-      - 1497894, 1519120
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1497894
```
wlach, line 12: Making these changes here so we can still test the same sorts of things.

Dockerfile
```diff
@@ -35,6 +35,9 @@ RUN pip install -r requirements.txt
 COPY test_requirements.txt ${HOME}/
 RUN pip install -r test_requirements.txt
 
+# HACK: install custom version of glean parser
+RUN pip install -U git+https://github.com/mozilla/glean_parser@tags
```
wlach, line 39: Obviously this needs to be removed before we merge. Instead we should version bump glean_parser, release a new version, and have probe_scraper depend on it.

probeinfo_api.yaml
```diff
@@ -618,6 +630,15 @@ components:
         the BigQuery datasets where decoded ping data is written by the pipeline.
       example: org_mozilla_firefox_beta
 
+    TagFiles:
+      type: array
+      description: A list of relative paths to `tags.yaml` files
+      default: []
+      items:
+        type: string
+        pattern: tags\.yaml$
```
badboy, line 639: https://github.com/mozilla/glean/pull/1849/files#diff-01a82f719b00c6c1cf94ad9ae35864e3997965fd9bcee99a77a4f5f50c3c3b1bR19-R20

We say they can be named whatever, but then probe-scraper actually needs them with a specific name. We should make sure our docs are consistent on that
(I guess our docs on metrics.yaml/pings.yaml are also wrong then, I'll file a bug)

badboy (approved):
Preliminary r+.
Blocked by the glean_parser release, but you pointed out all the places that need to be reverted again.

probeinfo_api.yaml
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1738880

PR #367, opened 2021-11-08
Keep git checkouts across runs

relud:
Put bare repo in cache dir, but exclude it from upload to s3

restores #331

### DIFF
--- a/probe_scraper/scrapers/git_scraper.py
+++ b/probe_scraper/scrapers/git_scraper.py
@@ -3,7 +3,6 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import os
-import shutil
 import tempfile
 import traceback
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


probe_scraper/scrapers/git_scraper.py
```diff
@@ -95,51 +94,50 @@ def retrieve_files(repo_info, cache_dir):
     results = defaultdict(list)
     timestamps = dict()
     base_path = os.path.join(cache_dir, repo_info.name)
+    repo_path = f"{base_path}.git"
```
badboy, line 97: oooh, so before we were mixing the git checkout and some file checkouts into the same directory, making it harder to cache one but not the other. That makes sense now!

probe_scraper/scrapers/git_scraper.py
```diff
@@ -95,51 +94,50 @@ def retrieve_files(repo_info, cache_dir):
...
 
-    if os.path.exists(repo_info.name):
-        shutil.rmtree(repo_info.name)
-    repo = git.Repo.clone_from(repo_info.url, repo_info.name)
+    if os.path.exists(repo_path):
+        print(f"Pulling latest commits into {repo_path}")
+        repo = git.Repo(repo_path)
+    else:
+        print(f"Cloning {repo_info.url} into {repo_path}")
+        repo = git.Repo.clone_from(repo_info.url, repo_path, bare=True)
```
badboy, line 110: Also a bare repo now, makes much more sense.

badboy (approved):

PR #368, opened 2021-11-09
Add test that repositories.yaml only contains any references to libraries that exist

### LINKED ISSUE #348: Add test that repositories.yaml only contains any references to libraries that exist

Marlene-M-Hirose:
fixes #348

Add test that repositories.yaml only contains any references to libraries that exist #348

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,4 +106,4 @@ _temp/
 index.html
 
 # Ignore .DS_Store
-**/.DS_Store
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,6 @@ _temp/
 
 # Generated files
 index.html
+
+# Ignore .DS_Store
+**/.DS_Store
--- a/tests/test_library_refs.py
+++ b/tests/test_library_refs.py
@@ -5,7 +5,11 @@ def test_library_refs():
     yaml_file = open("repositories.yaml", "r")
     repositories = yaml.safe_load(yaml_file)
     yaml_file.close()
-    libs = set(library["library_name"] for library in repositories["libraries"])
+    dependency_list = []
+    for library in repositories["libraries"]:
+        for variant in library["variants"]:
+            dependency_list.append(variant["dependency_name"])
+    libs = set(dependency_list)
     for app in repositories["applications"]:
         missing_libs = set(app["dependencies"]) - libs
         if missing_libs:
--- a/tests/test_library_refs.py
+++ b/tests/test_library_refs.py
@@ -1,0 +1,10 @@
+import yaml
+
+def test_library_refs():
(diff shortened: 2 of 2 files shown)


tests/test_library_refs.py
```diff
@@ -0,0 +1,10 @@
+import yaml
+
+def test_library_refs():
+  yaml_file = open("repositories.yaml", "r")
+  repositories = yaml.safe_load(yaml_file)
+  yaml_file.close()
+  libs = set(library["library_name"] for library in repositories["libraries"])
+  for app in repositories["applications"]:
+    if missing_libs := set(app["dependencies"]) - libs:
+      raise KeyError(f'application {app["app_name"]} contains invalid library references: {missing_libs}')
```
relud, line 10: this needs to be formatted by running `venv/bin/black tests/test_library_refs.py`

relud (changes requested):

.gitignore
```diff
@@ -104,3 +104,6 @@ _temp/
 
 # Generated files
 index.html
+
+# Ignore .DS_Store
+**/.DS_Store
```
relud, line 109: the `**/` prefix here isn't necessary:
```suggestion
.DS_Store
```

tests/test_library_refs.py
```diff
@@ -0,0 +1,13 @@
+import yaml
+
+
+def test_library_refs():
+    yaml_file = open("repositories.yaml", "r")
+    repositories = yaml.safe_load(yaml_file)
+    yaml_file.close()
+    libs = set(library["library_name"] for library in repositories["libraries"])
+    for app in repositories["applications"]:
+        if missing_libs := set(app["dependencies"]) - libs:
```
relud, line 10: looks like the installed version of flake8 doesn't like `:=`, so this needs to be split across two statements:
```suggestion
        missing_libs = set(app["dependencies"]) - libs
        if missing_libs:
```

relud (changes requested):

tests/test_library_refs.py
```diff
@@ -0,0 +1,18 @@
...
+
+def test_library_refs():
+    yaml_file = open("repositories.yaml", "r")
+    repositories = yaml.safe_load(yaml_file)
+    yaml_file.close()
+    dependency_list = []
+    for library in repositories["libraries"]:
+        for variant in library["variants"]:
+            dependency_list.append(variant["dependency_name"])
+    libs = set(dependency_list)
```
relud, line 12: this can be a set from the start:
```suggestion
    libs = set()
    for library in repositories["libraries"]:
        for variant in library["variants"]:
            libs.add(variant["dependency_name"])
```

relud (commented):
r+wc

relud (approved):

PR #371, opened 2021-11-16
refactor as per relud's suggestion

Marlene-M-Hirose:


relud:
this is a duplicate of #370

PR #372, opened 2021-11-16
Mark the regrets reporter as a prototype application

wlach:
This will clearly mark its current state in the Glean Dictionary

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -756,6 +756,7 @@ applications:
       - pings.yaml
     dependencies:
       - glean-js
+    prototype: true
     retention_days: 365
     channels:
       - v1_name: regrets-reporter-ucs


Dexterp37:
> 
> 
> This will clearly mark its current state in the Glean Dictionary

Please note that this is not really a prototype, but rather a "pre-release". If we merge this, will we be able to roll back and remove its status with an upcoming PR?

wlach:
> > This will clearly mark its current state in the Glean Dictionary
> 
> Please note that this is not really a prototype, but rather a "pre-release". If we merge this, will we be able to roll back and remove its status with a follow-up PR?

Of course, it's just metadata that doesn't affect the pipeline.

Dexterp37 (approved):

PR #374, opened 2021-11-17
Update vendored parsers

relud:


probe_scraper/parsers/scalars.py
```diff
@@ -17,7 +17,6 @@ def extract_scalar_data(s):
     return {
         "description": s.description,
         "expiry_version": get_major_version(s.expires),
-        "cpp_guard": s.cpp_guard,
```
relud, line 20: vendored parser removed `cpp_guard` from the list optional fields for scalars

tests/test_scalar_parser.py
```diff
@@ -15,7 +19,6 @@ def test_scalar_parser():
 
     # Make sure each of them contains all the required fields and details.
     REQUIRED_FIELDS = [
-        "cpp_guard",
```
relud, line 18: vendored parser removed `cpp_guard` from the list optional fields for scalars

probe_scraper/parsers/histograms.py
```diff
@@ -37,8 +37,8 @@ def extract_histogram_data(histogram, version):
         value = None
         if hasattr(histogram, source_field):
             value = getattr(histogram, source_field)()
-        elif source_field in histogram._definition:
-            value = histogram._definition.get(source_field)
+        elif hasattr(histogram, "_" + source_field):
+            value = getattr(histogram, source_field)
```
relud, line 41: `_definition` is no longer available, but the fields being pulled from it here are now available as separate private properties

probe_scraper/parsers/histograms.py
```diff
@@ -85,7 +85,9 @@ def transform_probe_info(probes, version):
 class HistogramsParser:
     def parse(self, filenames, version=None, channel=None):
         # Call the histogram tools for each file.
-        parsed_probes = list(histogram_tools.from_files(filenames))
+        parsed_probes = list(
+            parse_histograms.from_files(filenames, strict_type_checks=False)
+        )
```
relud, line 90: vendored parser changed to require explicitly ignoring type checks. This fixes the previous warning: `Unable to parse whitelist (%s). Assuming all histograms are acceptable.`

probe_scraper/parsers/events.py
```diff
@@ -26,7 +26,7 @@ def extract_events_data(e):
         "expiry_version": "never",
         "expiry_day": "never",
         "name": e.methods[0],
-        "description": e.description,
+        "description": e._definition.get("description"),
```
relud, line 29: vendored parser no longer exposes `description`, I'm uncertain whether this line should instead be deleted.

probe_scraper/parsers/third_party/parse_histograms.py
```diff
@@ -0,0 +1,1011 @@
...
+        if x == "counted_unknown_properties.py"
+        else None
+    ),
+    (lambda x: from_properties_db if x == "properties-db.js" else None),
+]
+
+# Similarly to the dance above with buildconfig, usecounters may not be
+# available, so handle that gracefully.
+try:
+    from . import usecounters
```
relud, line 928: this is the only line I modified in the vendored parsers, it was originally just `import usecounters`. I'm uncertain whether we should move usecounters to the top level to prevent issues each time this file is updated.

tests/resources/UseCounters.conf
```diff
@@ -38,4 +38,3 @@
 
 method SVGSVGElement.getElementById
 attribute SVGSVGElement.currentScale
-property Fill
```
relud, line 41: `property` was removed from the list of keys allowed here, but I don't actually know what this line was doing before, and tests pass fine without it.

tests/test_event_parser.py
```diff
@@ -40,7 +47,12 @@ def parse(channel, version):
     return parser.parse(["tests/resources/test_events.yaml"], version, channel)
 
 
-def test_channel_version_ignore():
+@patch("os._exit")
+@patch(
+    "probe_scraper.parsers.third_party.shared_telemetry_utils.ParserError.eventual_errors"
```
relud, line 52: `tests/resources/test_events.yaml` intentionally contains eventual errors, this prevents those from raising an exception when pytest exits.

probe_scraper/parsers/scalars.py
chutten: We sure did. It kinda sucked. I sure hope this doesn't trip up any historical parsing (if it did, the tests would've caught it, right?).

probe_scraper/parsers/third_party/parse_histograms.py
chutten: I think we're good for patching it like this. I intend to make no future changes to these parsers.

chutten (approved):

probe_scraper/parsers/scalars.py
relud: unit tests failed when the `cpp_guard` field was present in a scalar.

note to future relud: do a test run of probe scraper to see if this will be an issue before merging.

relud: or I could patch `cpp_guard` back in

chutten: It does seem like that'd be the smaller diff, true

relud: I'll do that then. I don't want to break things while I'm out, though, so converting this to a draft until after US thanksgiving.

relud: i misunderstood what was going on. unit tests pass fine when `cpp_guard` is present in the scalar definition.

they only failed because it was missing from the parser output, which is expected.

PR #376, opened 2021-11-19
bug 1742161 - Add new browser component metrics.yaml

chutten:
And move browser/base's metrics.yaml to Firefox Desktop. It's Firefox Desktop-specific and won't be needed in other gecko-dependent apps.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -165,6 +165,7 @@ libraries:
       - chutten@mozilla.com
     url: https://github.com/mozilla/gecko-dev
     metrics_files:
+      - browser/base/content/metrics.yaml
       - gfx/metrics.yaml
       - toolkit/components/glean/metrics.yaml
       - toolkit/components/processtools/metrics.yaml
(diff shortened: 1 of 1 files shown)


wlach (changes requested):
Unfortunately some of these metrics defined here already made their way into Firefox for Android:

https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/browser_ui_proton_enabled

I'm pretty sure we can't remove them without messing up the generated BigQuery schemas. One possible workaround, since it seems like only Firefox for Android uses "gecko" would be to create some kind of shim library just for that application, though it will still accumulate browser metrics over time. Maybe there's a better alternative that I haven't thought of

/ccing @relud and @jklukas to confirm my assessment

jklukas:
> I'm pretty sure we can't remove them without messing up the generated BigQuery schemas

Yes, if these fields have already made it into Fenix tables, then removing the dependency will cause schema deploy to fail, since removing the fields is an incompatible change.

> create some kind of shim library just for that application, though it will still accumulate browser metrics over time

We could create some static repository that just defines the metrics at this point in time and use that for the shim library, so that it doesn't accumulate new metrics.

chutten:
Dang, that's a bummer. Well, the most important part of this change is getting the new metrics.yaml deployed so I'll file "Do something with browser/base/metrics.yaml" as a follow-up. I've filed #377 for this work and will get this change reduced.

PR #381, opened 2021-12-03
Add CODEOWNERS for repositories.yaml

wlach:
Adding/changing the data in repositories.yaml can have large downstream
effects, so we want to be a little strict on who can sign off on changes
here.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+# Adding/changing the data in repositories.yaml can have large downstream
+# effects, so we're a little stricter on who can sign off on changes here.
+repositories.yaml @wlach @jklukas @relud @whd @jasonthomas


wlach:
Are there are any names that aren't on this list that should be? @scholtzan maybe?

jklukas (approved):

scholtzan:
I haven't done much work on probe-scraper, so probably best to keep me off that list.

whd (approved):
I normally put CODEOWNERS at the top level for visibility but this seems reasonable as well. I will consider moving CODEOWNERS in some of my draft PRs e.g. https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/708/files to match this.

You need to enable "require review from a CODEOWNER" in branch protection rules to actually enforce r+ coming from a CODEOWNER.

I believe that this list gates "client-side DE", who are the other frequent contributors to the repo (I see most of Mike's team in `git blame`), from making changes without "server-side DE" (or SRE)'s approval, so e.g. all glean app requests will now require server-side DE to approve them. This is intentional, but depending on the amount of friction this causes, we should prioritize CI checks for this accordingly, or add some client team folks to this list. In practice, I think adding new repos/glean apps is not expected to be problematic, it's changing existing apps that has issues, and the former happens more often than the latter.

wlach:
> I believe that this list gates "client-side DE", who are the other frequent contributors to the repo (I see most of Mike's team in `git blame`), from making changes without "server-side DE" (or SRE)'s approval, so e.g. all glean app requests will now require server-side DE to approve them. This is intentional, but depending on the amount of friction this causes, we should prioritize CI checks for this accordingly, or add some client team folks to this list. In practice, I think adding new repos/glean apps is not expected to be problematic, it's changing existing apps that has issues, and the former happens more often than the latter.

Yeah, it might well make sense for someone on the client team to be added to this list. I'm planning on writing up some documentation on how to reason through changes here to make that more possible. /cc @mdboom

wlach:
Also enabled the branch protection rule on main.

Issue #276, opened 2021-03-08
create bugs about expiring probes in toolkit/components/glean/metrics.yaml


Archaeopteryx:
In Firefox 89, the probes `fog_validation.legacy_telemetry_client_id/os_version/profile_disk_is_ssd` will expire. Unless they will be renewed or removed, this will cause build bustage. They are defined in [toolkit/components/glean/metrics.yaml](https://searchfox.org/mozilla-central/rev/44695ef057e422a8d6c6056972bdf32766c36187/toolkit/components/glean/metrics.yaml#52).

They should have been reported as expiring [similar to other telemetry probes](https://bugzilla.mozilla.org/buglist.cgi?list_id=15623791&query_format=advanced&email1=telemetry-probes%40mozilla.bugs&chfieldfrom=2021-02-21&classification=Client%20Software&classification=Developer%20Infrastructure&classification=Components&classification=Server%20Software&classification=Other&chfield=%5BBug%20creation%5D&chfieldto=2021-03-08&emailtype1=substring&emailreporter1=1).

See [bug 1696890](https://bugzilla.mozilla.org/show_bug.cgi?id=1696890).

chutten:
We also don't send emails. Which sets our users up to permafail their tests on merge day (a la https://bugzilla.mozilla.org/show_bug.cgi?id=1744667). We should both send emails _and_ file bugs.

PR #382, opened 2021-12-07
WIP #276 Firefox desktop expiry mails and bugs

chutten:
Please check my approach so far. This is the largest change I've made to `probe-scraper` and so I want to make sure I'm not bull-in-a-china-shopping it.

So far this supports email sending, not bug filing (yet). Fixes some nits while I was in the neighbourhood.

For bug filing I'm thinking about importing as much as I can from `probe_expiry_alert` and calling it directly from `runner.py`, but I'm not firm on that yet.

### DIFF
--- a/probe_scraper/glean_checks.py
+++ b/probe_scraper/glean_checks.py
@@ -143,26 +143,28 @@ def check_for_duplicate_metrics(repositories, metrics_by_repo, emails):
 
 
 EXPIRED_METRICS_EMAIL_TEMPLATE = """
-The following metrics in {repo_name} will expire in the next {expire_days} days or have already expired.
+Each metric in the following list from {repo_name} will expire in the next {expire_days} days or has already expired.
 
 {expired_metrics}
 
 What to do about this:
 
-1. If the metric is no longer needed, remove it from the `metrics.yaml` [1] file.
+1. If the metric is no longer needed, remove it from its `metrics.yaml` [1] file.
 2. If the metric is still required, resubmit a data review [2] and extend its expiration date.
 
-If you have any problems, please ask for help on the #glean Slack channel. We'll give you a hand.
+If you have any problems, please ask for help on the #glean Matrix channel[3]. We'll give you a hand.
 
 What happens if you don't fix this:
 
-The metrics listed above will stop collecting data from builds built after this expiration date, and you will continue to get this e-mail as a daily reminder.
+The metrics listed above will stop collecting data from builds built after this expiration date,
+and you will continue to get this e-mail as a reminder.
 
 Your Friendly, Neighborhood Glean Team
 
-[1] One of these files:
+[1] The correct metrics.yaml is in this list:
 {metrics_yaml_url}
 [2] https://wiki.mozilla.org/Firefox/Data_Collection
+[3] https://chat.mozilla.org/#/room/#glean:mozilla.org
 
 This is an automated message sent from probe-scraper.  See https://github.com/mozilla/probe-scraper for details.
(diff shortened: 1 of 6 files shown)


probe_scraper/glean_checks.py
```diff
@@ -189,6 +194,10 @@ def check_for_expired_metrics(
         repo_by_name[repo.name] = repo
 
     for repo_name, commits in repos_metrics.items():
+
+        if repo_name == "firefox-desktop" and latest_nightly_version is None:
```
wlach, line 198: Can you add a comment about the reasoning here?

probe_scraper/parsers/third_party/usecounters.py
```diff
@@ -9,7 +9,7 @@
 
 def read_conf(conf_filename):
     # Can't read/write from a single StringIO, so make a new one for reading.
-    stream = open(conf_filename, "rU")
+    stream = open(conf_filename, "r")
```
wlach, line 12: `r` is the default

```suggestion
    stream = open(conf_filename)
```

probe_scraper/glean_checks.py
```diff
@@ -142,44 +142,49 @@ def check_for_duplicate_metrics(repositories, metrics_by_repo, emails):
     return found_duplicates
 
 
+FIREFOX_DESKTOP_EXPIRY_LANGUAGE = (
+    "expire in the next version (Firefox {latest_nightly_version})"
+)
+DATE_BASED_EXPIRY_LANGUAGE = "expire in the next {expire_days} days"
 EXPIRED_METRICS_EMAIL_TEMPLATE = """
-The following metrics in {repo_name} will expire in the next {expire_days} days or have already expired.
+Each metric in the following list from {repo_name} will {expiry_language} or has already expired.
```
wlach, line 150: "The following metrics in" reads a little better to me, any reason why you changed the language?

Also, I think ideally we would list out each set of metrics (has expired, soon to expire) separately. It might be worth considering using a jinja template here, which would make it easier to express this type of conditional logic.

probe_scraper/glean_checks.py
```diff
@@ -202,15 +211,34 @@ def check_for_expired_metrics(
         for metric_name, metric in metrics.items():
             if metric["expires"] == "never":
                 continue
+
+            # `expires` field supports manual expiry, too.
+            if metric["expires"] == "expired":
```
wlach, line 216: TIL. Is this documented somewhere? Has anyone done this?

probe_scraper/glean_checks.py
```diff
@@ -202,15 +211,34 @@ def check_for_expired_metrics(
...
+                expired_metrics.append(f"- {metric_name} manually expired")
+                addresses.update(metric["notification_emails"])
```
wlach, line 218: This duplicates code that comes later. I wonder if it would be better to define a "has expired" function and then use it here? Might make this easier to follow

wlach (commented):
This looks pretty good to me! Have some minor suggestions

probe_scraper/glean_checks.py
chutten: It'd have to return the custom messages, but it could work. I'll see what I can do.

chutten:
Actually, you know what, If I'll be filing bugs for these expiring metrics then the emails we send should include links to the filed bugs.

I think I need to factor all the FOG-specific stuff to its own module at this point and stop trying to piggy-back on glean checks. Yes, FOG is Glean, but not sufficiently so that we can do expiry detection.

PR #394, opened 2022-01-14
Add gecko tags.yaml, xre metrics.yaml

chutten:
Adding the tags.yaml to firefox_desktop as well since the tags are used in both

xre's metrics are Firefox Desktop only and not Gecko because XRE stands for XUL Runtime Environment,
and on Android there is no XUL.

Part of bug 1746941 and bug 1743333

Should not land until tags.yaml lands in bug 1746941


chutten:
The patch for bug 1746941 is on autoland and should be landing shortly.

relud (approved):

jklukas (approved):

Issue #395, opened 2022-01-18
probe-scraper failure 2022-01-18: firefox_desktop's tags.yaml bounced on landing


chutten:
As seen in [the `gecko-dev` history for the file](https://github.com/mozilla/gecko-dev/commits/master/toolkit/components/glean/tags.yaml) `toolkit/components/glean/tags.yaml` landed, backed out, and then stuck its landing all within the same day.

Gonna hafta do something about that.

PR #397, opened 2022-01-18
Fix firefox-desktop's and background-update's bouncy tags

chutten:
They use the same repository and `tags.yaml` as `gecko`,
meaning they too are affected by that file landing, then being backed out,
then landing again.

Related to #396 and #395


relud (approved):

chutten:
Tested via `python -m probe_scraper.runner --cache-dir .scraper_cache/ --dry-run --glean --glean-repo gecko --glean-repo firefox-desktop-background-update --glean-repo glean-core`

Unfortunately, I couldn't add `firefox-desktop` into the mix without getting errors about `org.mozilla.components:service-glean` which I couldn't get the commandline to load a repo for.

PR #400, opened 2022-02-04
Update bergamot repo

eu9ene:
The repo was moved to another place

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -624,14 +624,14 @@ applications:
   - app_name: bergamot
     canonical_app_name: Bergamot Translator
     app_description: Web extension for on-device machine translation
-    url: https://github.com/mozilla-extensions/bergamot-browser-extension
+    url: https://github.com/mozilla/firefox-translations
     notification_emails:
       - epavlov@mozilla.com
     branch: main
     metrics_files:
-      - src/core/ts/background-scripts/background.js/telemetry/metrics.yaml
+      - extension/model/telemetry/metrics.yaml
     ping_files:
-      - src/core/ts/background-scripts/background.js/telemetry/pings.yaml
+      - extension/model/telemetry/pings.yaml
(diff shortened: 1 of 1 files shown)


jklukas (changes requested):
Probe scraper relies on reading the history of the target git repo to know when different metric definitions are valid at different points in the history. Moving an existing Glean app to point at a different repository that doesn't share the same commit history isn't supported, so this can't be merged as-is.

I defer to @relud for more info about what options we might have here.

The cleanest thing may be do define a new app. Is the bergamot name going away? If so, then creating a new `firefox_translations` app ID could be a good long-term move anyway.

eu9ene:
yes, it makes sense to add a new app to not mess data from the old extension and the new one

eu9ene:
> Probe scraper relies on reading the history of the target git repo to know when different metric definitions are valid at different points in the history. Moving an existing Glean app to point at a different repository that doesn't share the same commit history isn't supported, so this can't be merged as-is.
> 
> I defer to @relud for more info about what options we might have here.
> 
> The cleanest thing may be do define a new app. Is the bergamot name going away? If so, then creating a new `firefox_translations` app ID could be a good long-term move anyway.

@jklukas I added a new app, can we merge this?

jklukas (approved):
New app config looks great

jklukas:
Merged. One follow-up request, though: Can you send a PR to add `deprecated: true` as a top-level key to the `bergamot` app to indicate that it's no longer current?

eu9ene:
> Merged. One follow-up request, though: Can you send a PR to add `deprecated: true` as a top-level key to the `bergamot` app to indicate that it's no longer current? This has no direct effect in the pipeline at this point; it's just informational and may affect generated documentation.

sure https://github.com/mozilla/probe-scraper/pull/402

PR #404, opened 2022-02-15
Update glean_parser to support expiry-by-version

badboy:
Waiting for @chutten to verify this is all we need to do


Dexterp37:
> That update should contain the schema update permitting the `expires` key to have a value of type `int`, yes... but @Dexterp37 , does probe_scraper need to invoke glean_parser any differently to parse either type of expiry?

@chutten I used the same logic that you used for the custom expiry functions: if it wasn't needed for that, it isn't needed for the updated version (AFAICT).

chutten (approved):

PR #407, opened 2022-02-24
Skip expire checks for Glean repositories which use expire-by-version

badboy:
Fenix is starting to use expire-by-version.
This was merged last night: https://github.com/mozilla-mobile/fenix/pull/23805

Without this skip probe-scraper will fail due to:

```
    expires = datetime.datetime.strptime(
TypeError: strptime() argument 1 must be str, not int
```


Dexterp37 (approved):

badboy:
Note: Monday's probe-scraper run will fail, as the expire checks are only done on mondays.

relud (approved):

PR #408, opened 2022-03-03
Set min date for firefox translations

eu9ene:


probe_scraper/scrapers/git_scraper.py
```diff
@@ -30,6 +30,7 @@
     "glean-js": "2020-09-21 13:35:00",
     "mozilla-vpn": "2021-05-25 00:00:00",
     "rally-markup-fb-pixel-hunt": "2021-12-04 00:00:00",
+    "firefox-translations": "2022-02-25 00:00:00",
```
relud, line 33: please instead add to `SKIP_COMMITS` below:
```python
    "firefox-translations": [
       "02dc27b663178746499d092a987ec08c026ee560",  # Invalid extension/model/telemetry/metrics.yaml
    ],
```

relud (changes requested):

probe_scraper/scrapers/git_scraper.py
eu9ene: done. it's what I wanted to do initially, but then I thought anyway we don't need earlier data. thanks!

relud (approved):

relud (approved):

PR #418, opened 2022-04-15
Configure google cloud function to scrape glean repo commits

relud:


.circleci/config.yml
```diff
@@ -56,6 +56,24 @@ jobs:
...
+      - image: gcr.io/google.com/cloudsdktool/cloud-sdk:323.0.0
+    working_directory: ~/mozilla/probe-scraper
+    steps:
+      - checkout
+      - run:
+          name: Configure gcloud
+          command: |
+            echo $GCLOUD_SERVICE_KEY | gcloud auth activate-service-account --key-file=-
+            gcloud --quiet config set project ${GOOGLE_PROJECT_ID}
+            gcloud --quiet config set compute/zone ${GOOGLE_COMPUTE_ZONE}
```
relud, line 70: TODO: verify that these values are all correctly set in the build context

.circleci/config.yml
```diff
@@ -110,11 +128,18 @@ workflows:
...
+      - deploy_docker:
           requires:
             - build_and_test
           filters:
             tags:
               only: /.*/
             branches:
               only: main
+      - deploy_cloud_function:
+          context: data-eng-probe-scraper
```
relud, line 140: TODO: update this with the build context provided in [DSRE-684](https://mozilla-hub.atlassian.net/browse/DSRE-684)

probe_scraper/glean_push.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+
+
+def main(request: Request) -> Response:
+    """Scrape probes from a single glean commit."""
+    output_bucket = os.environ.get(
+        "OUTPUT_BUCKET",
+        # "s3://net-mozaws-prod-us-west-2-data-pitmo",
+        "gs://relud1-probe-scraper/gcf_output",
+    )
```
relud, line 121: TODO: update this with the gcs bucket provided in [DSRE-684](https://mozilla-hub.atlassian.net/browse/DSRE-684)

.github/workflows/glean.yaml
```diff
@@ -0,0 +1,25 @@
+name: GLEAN probe-scraper
+
+on:
+  workflow_call:
```
jklukas, line 4: This line seems strange. Is this defining an empty map?

.github/workflows/glean.yaml
```diff
@@ -0,0 +1,25 @@
+name: GLEAN probe-scraper
```
jklukas, line 1: Any particular reason we're styling Glean in all-caps?

main.py
```diff
@@ -0,0 +1,5 @@
+"""Google cloud function entry points."""
```
jklukas, line 1: Does this need to exist at the top level of the repo? Why can't we have this under `probe_scraper`? The docs here should contain some commentary about that.

jklukas (approved):
r+wc

.github/workflows/glean.yaml
relud: for some reason i incorrectly thought that's how Glean was supposed to be styled. will fix.

.github/workflows/glean.yaml
relud: This is directly copied from [the reusable workflow docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow), and is equivalent to:

```
on:
  workflow_call: null
```

we don't need any inputs or secrets, but if we did it would be like this:

```
on:
  workflow_call:
    inputs:
      username:
        required: true
        type: string
    secrets:
      envPAT:
        required: true
```

main.py
relud: it does, i will expand the docstring to explain

PR #419, opened 2022-04-20
Bug 1762997 - Do not scrape old version of the new Rally study

Dexterp37:
See bug 1762997 for more context, but this boils down to older commits being temporary.


whd (approved):
If this has the effect of deleting old pings, an update also needs to be made to MSG to allow this deletion: https://github.com/mozilla/mozilla-schema-generator/blob/main/incompatibility-allowlist

jklukas (approved):
r+ but agree with @whd about needing to add this to the incompatibility allowlist in MSG

Dexterp37:
> If this has the effect of deleting old pings, an update also needs to be made to MSG to allow this deletion: https://github.com/mozilla/mozilla-schema-generator/blob/main/incompatibility-allowlist

Thanks for the tip. Here's the PR: mozilla/mozilla-schema-generator#224

PR #421, opened 2022-04-29
bug 1765429 - Add pine app to repositories.yaml

chutten:
Might've been cleaner if I could've marked `firefox-desktop` as a dep, or done something refined by extracting a common dep as a "library", but this gets us where we need to be in the most understandable way, I think.


chutten:
Tested this locally with `python -m probe_scraper.runner --glean --glean-repo glean-core --glean-repo glean-android --glean-repo pine --glean-repo gecko --dry-run --cache-dir .scraper_cache/`

relud (approved):

PR #422, opened 2022-05-04
Bug 1767513 - Add the Sync/* metric files to Firefox iOS

Dexterp37:
This complements the implementation work that happened in mozilla-mobile/firefox-ios#10353

Note that the pings are marked `temp-*` to clearly spell out that these are a temporary workaround to get data out since legacy telemetry is not working as expected in Firefox iOS.

They will go away once Firefox iOS starts using the implementation of Sync from A-S that already sends Glean data.


jklukas (approved):
Thank you for including the in-code commentary to make sure we continue to have the right context on this available.

PR #424, opened 2022-05-05
Bug 1763467 Add moz_pipeline_metadata to repository

gleonard-m:
Added moz_pipeline_metadata and moz_pipeline_metadata_default to repository definition

### DIFF
--- a/probeinfo_api.yaml
+++ b/probeinfo_api.yaml
@@ -294,6 +294,10 @@ components:
             $ref: "#/components/schemas/RetentionDays"
           encryption:
             $ref: "#/components/schemas/Encryption"
+          moz_pipeline_metadata_defaults:
+            $ref: "#/components/schemas/MozPipelineMetadata"
+          moz_pipeline_metadata:
+            $ref: "#/components/schemas/MozPipelineMetadataOverride"
           skip_documentation:
             $ref: "#/components/schemas/SkipDocumentationBool"
 
@@ -336,6 +340,10 @@ components:
             $ref: "#/components/schemas/RetentionDays"
           encryption:
             $ref: "#/components/schemas/Encryption"
+          moz_pipeline_metadata_defaults:
+            $ref: "#/components/schemas/MozPipelineMetadata"
+          moz_pipeline_metadata:
+            $ref: "#/components/schemas/MozPipelineMetadataOverride"
           skip_documentation:
             $ref: "#/components/schemas/SkipDocumentationBool"
 
@@ -397,6 +405,10 @@ components:
           $ref: "#/components/schemas/RetentionDays"
         encryption:
           $ref: "#/components/schemas/Encryption"
+        moz_pipeline_metadata_defaults:
+          $ref: "#/components/schemas/MozPipelineMetadata"
+        moz_pipeline_metadata:
+          $ref: "#/components/schemas/MozPipelineMetadataOverride"
         skip_documentation:
           $ref: "#/components/schemas/SkipDocumentationBool"
         channels:
@@ -513,6 +525,10 @@ components:
           $ref: "#/components/schemas/RetentionDays"
         encryption:
           $ref: "#/components/schemas/Encryption"
+        moz_pipeline_metadata_defaults:
+          $ref: "#/components/schemas/MozPipelineMetadata"
+        moz_pipeline_metadata:
+          $ref: "#/components/schemas/MozPipelineMetadataOverride"
         skip_documentation:
           $ref: "#/components/schemas/SkipDocumentationBool"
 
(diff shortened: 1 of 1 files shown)


probeinfo_api.yaml
```diff
@@ -734,6 +750,125 @@ components:
...
+            - minutes
+            - hours
+            - days
+
+    MozPipelineMetadataOverride:
+      type: object
+      description: |
+        Currently not in use.
+        Allows ping level override of metadata values.
+      patternProperties:
```
jklukas, line 849: This isn't showing up in the rendered docs. Can this be expressed with `additionalProperties` instead like https://github.com/mozilla/probe-scraper/blob/58667b28fd0e13e49e89afb527af2483f03ab598/probeinfo_api.yaml#L253-L257 ?

jklukas: And can we add a comment here to indicate that we're expecting the keys here to be ping names?

probeinfo_api.yaml
```diff
@@ -734,6 +750,125 @@ components:
...
+      description: |
+        Currently not in use.  Various options controlling data lifecycle.
+      additionalProperties: false
+      properties:
+        collect_through_date:
+          type: string
+          description: |
+            Currently not in use.  If present, the pipeline will reject new data with submission_timestamp after the
+            given date, sending it to error output.
+          pattern: "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
```
jklukas, line 789:
```suggestion
          pattern: "^[0-9]{4}-[0-9]{2}-[0-9]{2}$"
          example: "2025-12-31"
```

For string types, redoc generates `"string"` as the example field, but you can provide a more specific example this way and it will get rendered in the docs.

jklukas (changes requested):
Well done! This looks nearly complete. It's just the patternProperties piece that I'm concerned about.

probeinfo_api.yaml
```diff
@@ -294,6 +294,10 @@ components:
             $ref: "#/components/schemas/RetentionDays"
           encryption:
             $ref: "#/components/schemas/Encryption"
+          moz_pipeline_metadata_defaults:
+            $ref: "#/components/schemas/MozPipelineMetadata"
+          moz_pipeline_metadata:
+            $ref: "#/components/schemas/MozPipelineMetadataOverride"
```
jklukas, line 300: I notice that you chose the name `MozPipelineMetadataOverride` for this. I do worry a little bit about the naming of `moz_pipeline_metadata` and `moz_pipeline_metadata_defaults`. Do you feel like `moz_pipeline_metadata_overrides` would be a more helpful/explicit name?

probeinfo_api.yaml
jklukas: Yes, that looks great. Is this working with the current code? I rendered these locally but may have been using an outdated version of `redoc-cli`.

probeinfo_api.yaml
gleonard-m: I followed naming that was suggested in the proposal but I do think that `moz_pipeline_metadata/moz_pipeline_metadata_overrides` is a better structure (the more specific naming is lower in the tree) , except when I consider having no `moz_pipeline_metadata` at the top level and only have ping specific pipeline_metadata configured, then the naming `moz_pipeline_metadata_overrides` doesn't make sense.  I'll explore some options and update here.

jklukas: Another option would be to mash the two together like:

```
moz_pipeline_metadata:
  all-pings: # here are the defaults
  specific-ping-name: # here are overrides for specific-ping-name
  other-ping: ...
```

[`all-pings` is a reserved name in Glean internals](https://mozilla.github.io/glean/dev/core/internal/reserved-ping-names.html?highlight=all-pings#reserved-ping-names) so no application can ever define a ping with that name. So we'd be following that same format. Then we'd have the `probe_scraper` code handle `all-pings` specially as the place where defaults are defined. This feels a little bit magical, though.

jklukas: cc @Dexterp37 for gut reaction to the idea of using `all-pings` as a magic key for defining defaults here.

probeinfo_api.yaml
jklukas: We should also specify here that ping name here is specifically referring to the document type (which is `kebab-case`) rather than the BigQuery table name (which is `snake_case`).

probeinfo_api.yaml
jklukas: Also, we don't have to block on this naming discussion. I'm fine with merging this as it exists now so you can confidently work on the next pieces. It's still possible to circle back and modify these names fairly easily up to the point where we start relying on the interface in mozilla-schema-generator.

gleonard-m: What if we don't include `all-pings`?  The documentation would state that moz_pipeline_metadata sets metadata for all pings and describe how to set for specific pings (override or single config).

```
moz_pipeline_metadata:
  expiration_policy:
    delete_after_days: 18
  specific-ping-name: # here are overrides for specific-ping-name
    expiration_policy:
      delete_after_days: 30
  other-ping: ...
```
This case looks readable to me as well:
```
moz_pipeline_metadata:
    specific-ping-name: # here are overrides for specific-ping-name
    expiration_policy:
      delete_after_days: 30
```

jklukas: I think it would be difficult/confusing to mix the ping-specific sections at the same level (in JSON schema terms, we would have both properties and additionalProperties) - particularly imagine that an app defined a ping named `expiration_policy` which would be ambiguous. But I think providing a specific field for per-ping overrides nested under `moz_pipeline_metadata` would be reasonable. So that would be like:

```
moz_pipeline_metadata:
  expiration_policy:
    delete_after_days: 18
  overrides:
    my-specific-ping:
      expiration_policy:
        delete_after_days: 30
```

Again the `overrides` name is up for debate. Could be `ping_specific`, `by_ping`, `per_ping`, `pings`, ... ?

probeinfo_api.yaml
gleonard-m: Update has been pushed.

probeinfo_api.yaml
gleonard-m: We can determine if any changes are needed.  In the meantime I did update the definition name and ref definitions to match (moz_pipeline_metadata_defaults -> MozPipelineMetadataDefaults, moz_pipeline_metadata -> MozPipelineMetadata)

Dexterp37: > cc @Dexterp37 for gut reaction to the idea of using `all-pings` as a magic key for defining defaults here.

While it's true that it's reserved, I'd stay away from this name to minimize confusion.  Maybe something like `default-for-all-pings`? Or `all-pings-defaults`? or `base`?

jklukas (approved):
The changes here look great! Feel free to merge.

probeinfo_api.yaml
Dexterp37: Current plan (or nested field for overrides) look both reasonable to me !

PR #427, opened 2022-05-16
Bug 1763467 Updated repositories.yaml to new format

gleonard-m:
Updated existing apps to duplicate jwe_mappings and expiration_policy info into the new format and added  custom settings for topsites-impression pings.  The values are still not used.


relud:
looks fine, but deferring to @jklukas on having more context before an r+

relud (commented):
looks fine, but deferring to @jklukas on having more context before an r+

jklukas (approved):

PR #431, opened 2022-05-18
Add support for using gcs instead of s3

relud:
co-depends on https://github.com/mozilla/telemetry-airflow/pull/1515

### DIFF
--- a/probe_scraper/remote_storage.py
+++ b/probe_scraper/remote_storage.py
@@ -1,0 +1,174 @@
+import fnmatch
+import gzip
+from pathlib import Path
+from subprocess import check_call
+from tempfile import TemporaryDirectory
+from typing import Optional, Tuple, Union
+
+TEXT_HTML = "text/html"
+APPLICATION_JSON = "application/json"
+INDEX_HTML = "index.html"
+
+
+def _s3_sync(
+    src: Union[str, Path],
+    dst: Union[str, Path],
+    delete: bool = False,
+    exclude: Tuple[str, ...] = (),
+    acl: Optional[str] = None,
+    content_type: Optional[str] = None,
+    content_encoding: Optional[str] = None,
+    cache_control: Optional[str] = None,
+):
+    # must use sync for dirs and cp for files
+    if isinstance(src, Path) and src.is_file():
+        # must upload files with cp
+        s3_cmd = "cp"
+    else:
+        s3_cmd = "sync"
+
+    check_call(
+        ["aws", "s3", s3_cmd, str(src), str(dst)]
+        + (["--delete"] if delete else [])
+        + [
+            arg
+            for key, value in zip(
+                (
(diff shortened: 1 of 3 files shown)


probe_scraper/remote_storage/gcs.py
```diff
@@ -0,0 +1,64 @@
...
+
+from .utils import compile_glob, extract_bucket_and_prefix, path_walk
+
+
+def pull(src: str, dst: Path):
+    bucket_name, prefix = extract_bucket_and_prefix(src)
+    prefix_length = len(prefix)
+    bucket = storage.Client().get_bucket(bucket_name)
+    # TODO verify that this correctly handles prefix being a file
+    for blob in bucket.list_blobs(prefix=prefix):
```
akkomar, line 14: Would it make sense to use `gsutil rsync` to simplify this, similar to `aws s3 sync` in the other backend?

probe_scraper/remote_storage/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def _get_implementation(remote: str):
+    if remote.startswith("s3://"):
+        return s3
+    elif remote.startswith("gs://"):
+        return gcs
+    else:
+        raise ValueError(
+            f"remote path must have scheme like s3:// or gs://, got: {remote!r}"
```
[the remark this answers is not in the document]
akkomar: Also this requires a change in a DAG in `telemetry-airflow`.

akkomar (commented):
What's the plan for this switch?
What about schema generator which is consuming artifacts produced here?
What about Cloudfront?

probe_scraper/remote_storage/gcs.py
relud: Unfortunately `gsutil` doesn't expose flags for setting content-encoding and content-type, but the sdk does.

relud:
> What's the plan for this switch? What about schema generator which is consuming artifacts produced here? What about Cloudfront?

We will continue to use s3 in production while push-mode is being tested. When push mode is ready for production, that's when we'll cut the probe info url over to gcs. schema generator will need to be updated to use gcs at that time if it's not using the probe info url. Cloudfront will presumably be replaced with something in GCP, but I expect that will all be handled by SRE.

probe_scraper/remote_storage/__init__.py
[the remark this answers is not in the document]
relud: https://github.com/mozilla/telemetry-airflow/pull/1515

probe_scraper/remote_storage/gcs.py
relud: turns out it does support those things, via `-h`. I also thought that gsutil wasn't a python package, but it turns out I was also wrong about that.

I'm working on changing this to use `gsutil`

akkomar (approved):

PR #434, opened 2022-05-20
Use python's new pathlib instead of string paths

relud:
and add .dockerignore so that dev environments don't build silly large docker images
and add type hints

This is in preparation for adding gcs support in #431


akkomar (approved):

probe_scraper/runner.py
```diff
@@ -438,30 +455,30 @@ def setup_output_and_cache_dirs(output_bucket, cache_bucket, out_dir, cache_dir)
...
-                    out_filename = os.path.join(tmpdirname, rel_root, filename)
-                    with open(in_filename, "rb") as f1:
-                        with gzip.open(out_filename, "wb") as f2:
-                            f2.write(f1.read())
+            tmp = Path(tmpdirname)
+            for in_filename in out_dir.rglob("*"):
+                if not in_filename.is_dir():
+                    out_filename = tmp / in_filename.relative_to(out_dir)
+                    out_filename.parent.mkdir(parents=True, exist_ok=True)
+                    out_filename.write_bytes(gzip.compress(in_filename.read_bytes()))
```
chutten, line 481: I can't speak to this reconstruction. It appears as though it'll do the same thing (recursively globbing out_dir, and for each file ensuring its parent dirs are mirrored appropriately, then outputting gzipped bytes), but I'm not sufficiently familiar with the quirks of os.path and pathlib to be certain that there aren't dragons

probe_scraper/scrapers/git_scraper.py
```diff
@@ -131,31 +136,34 @@ def retrieve_files(repo_info, cache_dir):
...
 
 
-def scrape(folder=None, repos=None):
+def scrape(
+    folder: Optional[Path] = None, repos: Optional[List[Repository]] = None
+) -> Tuple[
+    Dict[str, Dict[str, Tuple[int, int]]],
+    Dict[str, Dict[str, List[Path]]],
+    Dict[str, Dict[str, List[Union[Dict[str, str], str]]]],
+]:
```
chutten, line 166: Yikes.

chutten (approved):
Type annotations FTW.

jklukas (approved):
LGTM

PR #437, opened 2022-05-23
Support filtering glean repos by url

relud:


Dexterp37:
@relud is this in service of the "triggered pull" ("push model"?)

relud:
> @relud is this in service of the "triggered pull" ("push model"?)

yes, i'm currently working on breaking down breaking that feature down into more manageable pieces that can be merged separately.

chutten (approved):
Do you think this warrants an example in README?

relud:
> Do you think this warrants an example in README?

I don't think so. The [Performing a Dry-Run](https://github.com/mozilla/probe-scraper/blob/main/README.md#performing-a-dry-run) section is already kind of long, and doesn't cover all of the cli options available. I think this one is fine to leave out for now. I'll probably add an example that uses `--glean-url` when I add `--glean-commit` and `--glean-commit-branch` in another PR.

PR #438, opened 2022-05-24
Bug1763467 add metadata to ping

gleonard-m:
- publish moz_pipeline_metadata as a top-level field for each ping, handling default and ping specific values
- publish added bq_dataset_family, bq_metadata_format, bq_table as top-level fields for each ping

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import argparse
+import copy
 import datetime
 import errno
 import gzip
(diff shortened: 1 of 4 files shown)


probe_scraper/runner.py
```diff
@@ -213,6 +214,47 @@ def dedupe_probes(results):
...
+        current_pings = pings_by_repo.get(repo.name)
+        for ping_name, ping in current_pings.items():
+            ping_metadata = repo.moz_pipeline_metadata.get(ping_name, {})
+            pipeline_metadata = copy.deepcopy(metadata_defaults)
+            pipeline_metadata["bq_table"] = ping_name.replace("-", "_") + "_v1"
+
+            apply_ping_specific_metadata(pipeline_metadata, ping_metadata)
+
+            if pipeline_metadata:
+                pings_by_repo.get(repo.name).get(ping_name)[
```
jklukas, line 253:
```suggestion
                ping[
```

I believe `ping` from the `for` statement should still be in scope here and refers to the relevant item within the overall `pings_by_repo` structure.

jklukas (commented):
This looks great. We discussed the problem of `"value": null` in Slack, and we should definitely make sure we retain such entries here before merging.

probe_scraper/runner.py
gleonard-m: Agreed

gleonard-m:
I removed `remove_none()` function call in **probe_scraper/parsers/repositories.py** since the **probeinfo_api.yaml** restricts null values.   `remove_none()` effectively did a deep copy therefore added an explicit `deepcopy()` in its place.  

Example glean/fenix/pings subsection:
```
"topsites-impression": {
  "history": [
  ],
  "in-source": true,
  "moz_pipeline_metadata": {
    "bq_dataset_family": "org_mozilla_fenix",
    "bq_metadata_format": "structured",
    "bq_table": "topsites_impression_v1",
    "expiration_policy": {
      "delete_after_days": 30
    },
    "override_attributes": [
      {
        "name": "geo_city",
        "value": null
      }
    ],
    "submission_timestamp_granularity": "seconds"
  },
  "name": "topsites-impression"
}
```

PR #439, opened 2022-05-25
Revert "Use python's new pathlib instead of string paths (#434)"

akkomar:
This reverts commit b0b39a8d75def123b1d353ff0cf70720ebb6b6d9
which seems to generate output to s3 incompatible with schema-generator
(particularly `repositories` file)


akkomar:
Reverting to unblock glam etl.

PR #442, opened 2022-05-31
Bug 1769579 - omit deleted onboarding ping in rally

relud:


relud:
r? @whd I ran this locally and confirmed that the changes were:
onboarding ping deleted
one commit removed from history for the `rally.id` metric

whd (approved):

PR #443, opened 2022-05-31
Add @akkomar to CODEOWNERS

relud:
These files are under CODEOWNERS restriction because changes to the files in question can result in ignoring history in probe scraper, which can have the unintentional impact of deleting tables and fields from schemas.

Adding @akkomar so that there is more than one DE authorized to review changes made by another DE in the list.

@akkomar this will cause you to be automatically tagged for review on changes to `repositories.yaml` and `probe_scraper/scrapers/git_scraper.py`.


akkomar (approved):
SGTM.

Just so I get more context: do we have history of unintentional table or field deletions by changes made here? (bugs maybe?)

PR #446, opened 2022-06-01
Add options for scraping a single commit from glean repos

relud:
and for validating whether a commit is on the published branch for that repo

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -13,6 +13,7 @@ import tempfile
 import traceback
 from collections import defaultdict
 from pathlib import Path
+from textwrap import dedent
 from typing import Any, Dict, List, Optional
 
 from dateutil.tz import tzlocal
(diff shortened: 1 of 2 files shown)


relud:
@akkomar @chutten or @jklukas any chance one of you will have time to review this on monday?

jklukas:
Can you provide a little more context on where this fits in the overall effort? Will anything be using these new options immediately? How will that be configured?

relud:
> Can you provide a little more context on where this fits in the overall effort? Will anything be using these new options immediately? How will that be configured?

push mode will use these options via a google cloud function writing to a gcs bucket, which will eventually replace the current prod. The PR for that cloud function depends on this and #448. GCP resources are pending [DSRE-853](https://mozilla-hub.atlassian.net/browse/DSRE-853).

the google cloud function will effectively call probe scraper with `--glean --update --glean-url={} --glean-commit={} --glean-branch={}`, and that will write probe info for all repos with that url to `--out-dir` but only upload probe info to `--output-bucket` for repos where that commit is on the correct branch.

in addition to that cloud function, there will be three new daily airflow jobs that write to gcp: probe scraper for moz central, probe scraper for glean apps that can't use push mode (everything in gecko-dev), and probe scraper that writes all of the generic stuff (index.html, repositories and v2 listings, etc).

this all won't be prod until we cut over to it after letting it run for a few weeks alongside the existing probe scraper stuff and verifying that everything is the same.

jklukas (approved):

PR #449, opened 2022-06-13
Publish cloud function via CircleCI

relud:
blocked on [DSRE-684](https://mozilla-hub.atlassian.net/browse/DSRE-684)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ version: 2
 jobs:
   build_and_test:
     machine:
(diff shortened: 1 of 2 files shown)


jklukas (approved):

badboy (approved):
Looks good. I just wonder if we should document it somewhere that every merge to `main` triggers a deploy of the function and thus changes are immediately active?

relud:
every merge to main already triggers a deploy of the docker image, which also makes changes immediately active.

PR #455, opened 2022-07-06
Bug 1778196 - Add the Rally Attention Stream study

Dexterp37:
The dry run seems to run fine, locally.


chutten (approved):
(Data Steward Hat On) I presume since it is opting out of Glean Dictionary documentation, it has a plan for publishing its own data documentation to the users subject to the collections?

PR #458, opened 2022-07-12
Fix Rally Attention Stream url

akkomar:
This is a follow up to https://github.com/mozilla/probe-scraper/pull/455/files. Looks like trailing slash in the git url causes probe scraper to fail here: https://github.com/mozilla/probe-scraper/blob/30c2783d4dd229703c86d028eb6d013f343ec7a1/probe_scraper/scrapers/git_scraper.py#L157 when retrieving files.


cmharlow (approved):
thank you for handling this

Dexterp37:
Thank you @cmharlow and @akkomar !

PR #462, opened 2022-07-19
Update Firefox Desktop metrics_files directories

Iinh:
Recently probeinfo API was not able to pick up some new Firefox Desktop metrics, such as [the new Early Hints type](https://github.com/mozilla/gecko-dev/commit/f15f91dc5652019b375263935ea59ddffc2b084c), because we did not have the most up-to-date `metrics_files` map.

I copied this list over from [toolkit/components/glean/metrics_index.py](https://github.com/mozilla/gecko-dev/blob/master/toolkit/components/glean/metrics_index.py#L14-L33), let me know if we need to leave out anything.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -196,9 +196,18 @@ applications:
       - browser/components/newtab/metrics.yaml
       - browser/components/search/metrics.yaml
       - browser/modules/metrics.yaml
+      - dom/media/metrics.yaml
+      - dom/metrics.yaml
+      - gfx/metrics.yaml
+      - netwerk/protocol/http/metrics.yaml
       - toolkit/components/extensions/metrics.yaml
+      - toolkit/components/glean/metrics.yaml
+      - toolkit/components/glean/tests/test_metrics.yaml
+      - toolkit/components/nimbus/metrics.yaml
+      - toolkit/components/processtools/metrics.yaml
       - toolkit/components/search/metrics.yaml
       - toolkit/components/telemetry/metrics.yaml
+      - toolkit/mozapps/update/metrics.yaml
       - toolkit/xre/metrics.yaml
     ping_files:
       - browser/components/newtab/pings.yaml
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -196,19 +196,11 @@ applications:
       - browser/components/newtab/metrics.yaml
       - browser/components/search/metrics.yaml
       - browser/modules/metrics.yaml
-      - dom/media/metrics.yaml
-      - dom/metrics.yaml
-      - gfx/metrics.yaml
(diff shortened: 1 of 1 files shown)


chutten (changes requested):
A lot of those files are actually part of the reusable `gecko` dependency, not `firefox-desktop` itself.

I think the only new one is `netwerk/protocol/http/metrics.yaml`? Is that something reusable across Desktop and Mobile? If so, it should probably be part of `gecko`.

If some of these should be added to firefox-desktop, please also note the note about pinebuild.

Iinh:
Thanks @chutten! You're right, looks like we only need to add `netwerk/protocol/http/metrics.yaml`.

repositories.yaml
```diff
@@ -200,6 +200,7 @@ applications:
       - toolkit/components/search/metrics.yaml
       - toolkit/components/telemetry/metrics.yaml
       - toolkit/xre/metrics.yaml
+      - netwerk/protocol/http/metrics.yaml
```
chutten, line 203: Please confirm that you've spoken with the component owner and ensured that this component isn't shared to our mobile browsers via geckoview. (If so, it'll just need to be moved up to `gecko`)

Also please confirm that you've established that `pine` doesn't use this code either.

chutten (changes requested):

repositories.yaml
Iinh: Thanks @chutten for your patience. Could you elaborate more on what you mean by

> ensured that this component isn't shared to our mobile browsers via geckoview. (If so, it'll just need to be moved up to gecko)

I asked @MarLeclair and this was the response.

> I thought that Amy network probe would get automatically called from either mobile or desktop. Unless I have to do something different to make it that way?

chutten: Definitely!

If the instrumentor is pretty sure we're using the same networking code in both Desktop and Mobile, then that means this definitions file properly belongs under the `gecko` dependency, not the `firefox-desktop` application. That way all the applications that use `gecko`'s networking will have columns for the networking probes.

MarcLeclair:
@chutten  Does this mean I have to change anything on my end for the probe to show up in Glam?

chutten:
@MarcLeclair Nope! If you confirm that this is in the reusable portion of m-c (e.g. maybe Fenix uses it) then this simply needs to be put in the `gecko` dep so all the apps that record it will have a column for this in their datasets.

@Iinh Do you need any help with this? I have some experience with changes like these and might be able to assist.

Iinh:
> @Iinh Do you need any help with this? I have some experience with changes like these and might be able to assist.

Thanks @chutten, assistance would be greatly appreciated! (I'm actually out sick atm, will be slow to respond)

PR #463, opened 2022-07-19
collect_through_date on rally-markup-fb-pixel-hunt

rhelmer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1779801 - add a collect_through_date to stop data collection on the Rally/Markup FB Pixel Hunt study.


whd (commented):
This looks correct to me but tagging @gleonard-m to verify since this is the first metadata update of its kind

PR #468, opened 2022-08-09
Handle missing emails.txt

relud:


probe_scraper/glean_push.py
```diff
@@ -67,8 +67,11 @@ def main(request: Request) -> Response:
             message = f"update published for {updates}\n"
         else:
             message = "update is valid, but not published\n"
-        emails = email_file.read_text()
-        if emails:
+        try:
+            emails = email_file.read_text()
```
akkomar, line 71: LGTM, but when can this be empty? On errors? Maybe it would make sense to add a note in the `except` block below.

akkomar (approved):

probe_scraper/glean_push.py
relud: this is usually empty, there are normally only emails when something recoverable went wrong or something is expiring.

probe_scraper/glean_push.py
```diff
@@ -67,8 +67,11 @@ def main(request: Request) -> Response:
...
+        except FileNotFoundError:
+            pass
```
relud, line 73:
```suggestion
            pass  # no emails means no warnings or expiring metrics, which is good
```

PR #470, opened 2022-08-11
Adding some missing Desktop components to repositories.yaml

travis79:
Adding the pdfjs and nimbus components to firefox_desktop and also adding nimbus to pine. The pdfjs one does not currently need enabled for Pine, but may need so in the future.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -202,7 +202,7 @@ applications:
       - toolkit/components/telemetry/metrics.yaml
       - toolkit/xre/metrics.yaml
       - toolkit/components/nimbus/metrics.yaml
-+     - toolkit/components/pdfjs/metrics.yaml
+      - toolkit/components/pdfjs/metrics.yaml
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -201,6 +201,8 @@ applications:
       - toolkit/components/search/metrics.yaml
       - toolkit/components/telemetry/metrics.yaml
       - toolkit/xre/metrics.yaml
+      - toolkit/components/nimbus/metrics.yaml
++     - toolkit/components/pdfjs/metrics.yaml
```
badboy, line 205:
```suggestion
      - toolkit/components/pdfjs/metrics.yaml
```

badboy: Mistakenly added a `+`?

travis79: Yeah, oops

badboy (approved):

PR #471, opened 2022-08-11
Properly accumulate parsing errors

### LINKED ISSUE #354: Prober scraper ignores glean ping errors
The way probe scraper is currently written, it will ignore any ping errors because the value of any accumulated errors get overwritten by a call to get the metric information:

https://github.com/mozilla/probe-scraper/blob/a05308b55a7c6e94251feec04b8a1b820c6cc88f/probe_scraper/runner.py#L325

relud:
fixes #354


fbertsch (approved):
LGTM. Doesn't look like theres an easy way to add a test for this, but since it's mostly informational I don't think it matters

relud:
waiting to merge this until Tuesday when i return from PTO

PR #474, opened 2022-08-16
Fix how git scraper handles a single commit

relud:
`git log --max-count=1 <ref> -- path/to/file.yaml` returns the most recent commit in the history of `<ref>` to modify the file, but we actually just want to return `<ref>` when `<ref>` is a commit. `--max-count` only produced the correct result when the repo was shallow, and the repo may be unshallowed by checking that a commit is in a given branch when the commit is not the current head of that branch.

In order to achieve the desired output, this PR removes the ineffective `max_count` argument in favor of a new `only_ref` argument.

### DIFF
--- a/probe_scraper/scrapers/git_scraper.py
+++ b/probe_scraper/scrapers/git_scraper.py
@@ -105,19 +105,19 @@ def _file_in_commit(repo: git.Repo, filename: Path, ref: str) -> bool:
 
 
 def get_commits(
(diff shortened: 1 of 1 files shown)


badboy (changes requested):
Same as with the other PR: please put the PR description into the commit message.
Otherwise this looks good.

badboy (approved):

PR #476, opened 2022-08-16
Fix glean definition updates

relud:
push mode can deliver commits out of order, so `make_item_defn()` should only update `last` entries with newer than existing values and `update_or_add_item()` needs to check for equality over all of history.

separately, `ping_equal()` needs to ignore `source_url`, because all commits will have a different value, and in push mode that causes an entry to appear for every commit.

### DIFF
--- a/probe_scraper/transform_probes.py
+++ b/probe_scraper/transform_probes.py
@@ -252,17 +252,15 @@ def get_minimum_date(probe_data, revision_data, revision_dates):
     return min_dates
 
 
-def pretty_ts(ts):
-    return datetime.utcfromtimestamp(ts).isoformat(" ")
-
-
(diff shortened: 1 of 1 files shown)


badboy (changes requested):
> push mode can deliver commits out of order, so `make_item_defn()` should only update `last` entries with newer than existing values and `update_or_add_item()` needs to check for equality over all of history.
> 
> separately, `ping_equal()` needs to ignore `source_url`, because all commits will have a different value, and in push mode that causes an entry to appear for every commit.

Can you put all this information in the commit message?
The change together with this description makes sense to me, the commit message as it stands currently on its own however does not explain why things are changed.

badboy (approved):

PR #480, opened 2022-08-24
Allow missing yaml files if a repo is deprecated

relud:


probe_scraper/exc.py
```diff
@@ -0,0 +1,4 @@
+class ProbeScraperInvalidRequest(Exception):
+    """Exception type for returning HTTP 4XX in push mode."""
```
chutten, line 2: It's specifically returning only `400`. Is this docstring future-proofing?

chutten (approved):

probe_scraper/exc.py
relud: yes

PR #483, opened 2022-08-25
Extend timeout so fenix has enough time

relud:
540s is the maximum timeout https://cloud.google.com/functions/docs/configuring/timeout


chutten (commented):
Seems fine to me, but I don't have the context to understand why this is important. Leaving it to Jan-Erik for that part.

relud:
context: fenix usually takes too long for the default 60s timeout, so their CI has been failing.

badboy (approved):

PR #484, opened 2022-08-25
Expose error message for cli failures in glean_push

relud:
Two failures in push mode earlier today were due to gsutil throwing `Caught non-retryable exception while listing gs://probe-scraper-prod-artifacts/glean/...: Could not reach metadata service: Internal Server Error`, and it would be useful if that message were visible in ci, but right now we just get a generic 500 error message, e.g. https://github.com/mozilla-mobile/mozilla-vpn-client/runs/8021419260?check_suite_focus=true

### DIFF
--- a/probe_scraper/exc.py
+++ b/probe_scraper/exc.py
@@ -1,4 +1,20 @@
-class ProbeScraperInvalidRequest(Exception):
+class ProbeScraperError(Exception):
+    """Exception type for returning errors in push mode."""
+
+    def __init__(self, message, status_code):
+        self.status_code = status_code
+        self.message = message
+
+
+class ProbeScraperInvalidRequest(ProbeScraperError):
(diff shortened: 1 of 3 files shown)


probe_scraper/exc.py
```diff
@@ -1,4 +1,20 @@
-class ProbeScraperInvalidRequest(Exception):
+class ProbeScraperError(Exception):
+    """Exception type for returning errors in push mode."""
+
+    def __init__(self, message, status_code):
+        self.status_code = status_code
+        self.message = message
```
chutten, line 6: Up to you, but I like having these in the same order as the args list.

chutten (approved):

probe_scraper/exc.py
relud:
```suggestion
        self.message = message
        self.status_code = status_code
```

PR #485, opened 2022-08-25
Use new exception type for invalid glean url and commits

relud:


chutten (dismissed):

relud:
re-requesting review, change the exception type in an additional spot

chutten (approved):

PR #492, opened 2022-09-15
Bug 1790046 - Add a nightly task to update FOG

badboy:
This requires changes in m-c before we can merge that here.
This is an import of https://github.com/mozilla/fog-update-bot, because it turns out keeping the action within this repository is actually easier.

### DIFF
--- a/.github/workflows/update-fog.yml
+++ b/.github/workflows/update-fog.yml
@@ -1,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+# This workflow periodically calls the fog-update-bot action to update
+# the list of FOG metrics.yaml and ping.yaml files from its metrics_index.py
+
+
+name: "Update FOG"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  schedule:
+    # 04:20 UTC - every morning
+    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
+    - cron: '20 4 * * *'
+
+jobs:
+  main:
+    name: "Update FOG"
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: "Update FOG"
+        uses: ./fog-updater
+        id: fog-updater
+        if: github.repository == 'mozilla/probe-scraper'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/fog-updater/README.md
+++ b/fog-updater/README.md
@@ -1,0 +1,63 @@
+# fog-update-bot
+
+Automation to update `repositories.yaml` of `probe-scraper` with the latest `metrics_index.py` list.
+
+Fetches and parses the `metrics_index.py` from `gecko-dev`, extracts the relevant list of YAML files
+and creates a new Pull Request against `probe-scraper` if it contains any changes.
+
+## Environment variables
+
+| Name | Description |
+| ---- | ----------- |
+| `DEBUG` | If set enables debug logging |
+| `DRY_RUN` | If set to `True` will not create a PR |
+| `GITHUB_REPOSITORY_OWNER` | The owner of the `probe-scraper` repository |
+| `AUTHOR_NAME` | The name to use for the commit |
+| `AUTHOR_EMAIL` | The email to use for the commit |
+
+## Running with Docker
+
+```
+$ docker build -t fog-update .
+$ docker run -it --rm fog-update
+```
+
+## Development
+
+```
+$ python3 -m venv env
+$ pip install -r requirements.txt
+$ pip install pytest
+```
+
+## Testing
+
+You can run the tests:
+
+```
+pytest
+```
+
+Manual runs of the updater requires a `GITHUB_TOKEN`.
+Go to <https://github.com/settings/tokens> and create a new token (no additional scopes necessary).
+Set it in your shell:
+
+```
+export GITHUB_TOKEN=<the generated token>
+```
+
+## Code of Conduct
+
+This repository is governed by Mozilla's code of conduct and etiquette guidelines.
+For more details, please read the
+[Mozilla Community Participation Guidelines](https://www.mozilla.org/about/governance/policies/participation/).
+
+See [CODE_OF_CONDUCT.md](../CODE_OF_CONDUCT.md)
+
+## License
+
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+See [LICENSE](../LICENSE).
--- a/fog-updater/src/fog_update.py
+++ b/fog-updater/src/fog_update.py
@@ -1,0 +1,253 @@
+#!/usr/bin/env python3
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+from github import Github, GithubException, InputGitAuthor, enable_console_debug_logging
+import datetime
+import io
+import os
+import requests
+import sys
+
+DEFAULT_ORGANIZATION = "mozilla"
+DEFAULT_AUTHOR_NAME = "data-updater"
+DEFAULT_AUTHOR_EMAIL = "telemetry-alerts@mozilla.com"
+USAGE = "usage: fog-update"
+REVIEWERS = ["@mozilla/glean"]
+INDEX_URL = "https://raw.githubusercontent.com/mozilla/gecko-dev/master/toolkit/components/glean/metrics_index.py"  # noqa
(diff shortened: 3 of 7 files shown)


.github/workflows/update-fog.yml
```diff
@@ -0,0 +1,32 @@
...
+
+name: "Update FOG"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  schedule:
+    # 04:00 UTC - every morning
```
chutten, line 17: In what way are we guaranteed that this is interpreted wrt UTC?

fog-updater/README.md
```diff
@@ -0,0 +1,63 @@
...
+and creates a new Pull Request against `probe-scraper` if it contains any changes.
+
+## Environment variables
+
+| Name | Description |
+| ---- | ----------- |
+| `DEBUG` | If set enables debug logging |
+| `DRY_RUN` | If set to `True` will not create a PR |
+| `GITHUB_REPOSITORY_OWNER` | The owner of the `probe-scraper` repository |
+| `AUTHOR_NAME` | The name name to use for the commit |
```
chutten, line 15:
```suggestion
| `AUTHOR_NAME` | The name to use for the commit |
```
The Department of Redundancy Department would like a word.

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+from github import Github, GithubException, InputGitAuthor, enable_console_debug_logging
+import datetime
+import io
+import os
+import requests
+import sys
+
+DEFAULT_ORGANIZATION = "badboy"
```
chutten, line 14: Should this be `mozilla` now that we're here?

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+DEFAULT_AUTHOR_NAME = "data-updater"
+DEFAULT_AUTHOR_EMAIL = "jrediger@mozilla.com"
```
chutten, line 16: Could use `telemetry-alerts@` if we wanted to grant that weird pseudo-address its own account, I guess.

Depends on how much you'd like your email to not be in code.

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+USAGE = "usage: fog-update"
+REVIEWERS = ["@mozilla/glean"]
+INDEX_URL = "https://raw.githubusercontent.com/mozilla/gecko-dev/master/toolkit/components/glean/metrics_index.py"  # noqa
+BODY_TEMPLATE = """
+This (automated) patch updates the list from metrics_index.py.
+
+For reviewers:
+
+* Canonical source for the index: <https://searchfox.org/mozilla-central/source/toolkit/components/glean/metrics_index.py>
```
chutten, line 25: Didn't I have a comment about making sure this and `INDEX_URL` match?

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+* Please double-check that the changes here are valid and that the referenced files exist.
```
chutten, line 26: It's good to include consequences. "If the referenced files do not exist, schema deploys will fail" (though I suppose the CI run on the PR will fail first?

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+* Please double-check that none of the files are _deleted_ from the list.
```
chutten, line 27: "Deleted files result in incompatible schema changes that might not be immediately caught."

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+* Delete this branch after merging or closing the PR.
+
+---
+
+The source code of this automation bot lives in [badboy/fog-update-bot](https://github.com/badboy/fog-update-bot).
```
chutten, line 32: Not any more it doesn't!

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def eval_extract(code):
+    """
+    Eval `code` and return a map of variables and their values.
+
+    `code` should be valid Python code.
+    Only the builtins `list` and `set` are provided.
+
+    Note: this execute arbitrary Python code.
```
chutten, line 47:
```suggestion
    Note: this executes arbitrary Python code.
```

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+
+    new_content = content
+    for item in data:
+        name, metrics_or_pings, library, files = item
+        new_content = swap_file_list(
+            new_content, name, files, metrics_or_pings, library
+        )
+
+    if content == new_content:
+        raise Exception(
```
chutten, line 128: The Software Engineer in me objects to using exceptions to mark unexceptional (common/normal) situations.

The pragmatic programmer in me doesn't see that it makes any difference, though.

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+    except Exception as e:
+        print(f"{ts()} {e}")
+        return
+
+    if dry_run:
+        print(f"{ts()} Dry-run so not continuing.")
+        return
+
+    # Create a non unique PR branch name for work on this ac release branch.
+    pr_branch_name = f"fog-update/update-metrics-index-{short_version}"
```
chutten, line 189: TIL branch names can contain `/`

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+
+    try:
+        pr_branch = repo.get_branch(pr_branch_name)
+        if pr_branch:
+            print(f"{ts()} The PR branch {pr_branch_name} already exists. Exiting.")
```
chutten, line 194: So this means if we aren't quick enough to review these each day, the work will pile up?

fog-updater/src/fog_update.py
```diff
@@ -0,0 +1,252 @@
...
+            return
+    except GithubException:
+        # TODO Only ignore a 404 here, fail on others
```
chutten, line 197: 👀

fog-updater/src/test_util.py
```diff
@@ -0,0 +1,82 @@
...
+    "F"
+]
+"""
+
+
+def test_eval_metrics_index():
+    data = eval_extract(METRICS_INDEX)
+    assert data["first_yamls"] == ["A", "B"]
+    assert data["second_yamls"] == ["B", "C"]
+    assert data["metrics_yamls"] == ["A", "B", "C"]
```
chutten, line 45: Should probably also assert something about pings_yamls

fog-updater/src/test_util.py
```diff
@@ -0,0 +1,82 @@
...
+
+from fog_update import eval_extract, swap_file_list
+
+
+REPOSITORIES_YAML = """
+---
+version: "2"
+applications:
+  - app_name: firefox_desktop
+    metrics_files: # When adding here, consider if you should also add to pine.
```
chutten, line 14: Are you keeping this comment because it's hilarious?

fog-updater/src/test_util.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # New files added.
+    assert "- METRICS_FILES" in output
+    assert "- A" not in output
+    # ping files untouched.
+    assert "- PING_FILES" in output
+
+    # Other app untouched
+    assert "- OTHER_METRICS_FILES" in output
+    assert "- OTHER_PING_FILES" in output
```
chutten, line 82: Wouldn't mind a test covering `ping_files`

chutten (changes requested):
This seems like a thing that should be linked to from documentation. Maybe in FOG's docs we should link directly to this dir? Otherwise it might become some of that semi-hidden magic that mostly works and only like two people know exists.

.github/workflows/update-fog.yml
badboy: GitHub Actions docs [say so](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) (hm, might want to shift that by a bit just to not run into the called out "high load" periods.

fog-updater/src/fog_update.py
badboy: Whaaaaat? yes of course! 
You just can't have `fog-update` as a branch name then, at least in checkouts, as this is really just files in `.git/logs/refs/heads/`

fog-updater/src/fog_update.py
badboy: ... and we wouldn't even hear about it until we merge & delete and the next run comes around!

I'm ok with that for now.
We could switch to adding additional commits to the branch if it exists or always creating a fresh branch or something.

fog-updater/src/fog_update.py
badboy: aha copy'pasta from relbot.

fog-updater/src/test_util.py
badboy: depends on who you ask! (Nah, I copied that from the original a while back in the before times)

badboy:
Gecko work has been merged and https://firefox-source-docs.mozilla.org/toolkit/components/glean/user/new_definitions_file.html shows the current docs. We can land this now!

chutten (approved):

badboy:
Let's merge and see! (Nothing will break because it's just a PR)

PR #501, opened 2022-09-28
Fix KeyError due to incorrect type in fog_checks.get_current_metrics

relud:
partially fixes [Bug 1792701](https://bugzilla.mozilla.org/show_bug.cgi?id=1792701)


chutten (approved):

chutten:
Right, though of course you'll need to update the test first before landing

PR #502, opened 2022-10-05
Bug 1792701 - Do not ni? inactive accounts

Dexterp37:
Accounts reported in the metrics can get out of date and get disabled. We should not fail the probe-scraper because of that.


akkomar (approved):
👍

PR #503, opened 2022-10-05
Bug 1789931 - Update repositories.yaml to support the firefox-android monorepo

gabrielluong:
👋 We're looking to move the https://github.com/mozilla-mobile/android-components repo into a monorepo at https://github.com/mozilla-mobile/firefox-android. You can see an initial prep branch at https://github.com/mozilla-mobile/firefox-android/tree/ac-prep.

We'll want to coordinate how we want to land this while moving to the monorepo at some point. If there is someone from the data team that can take a look at the proposed change and provide feedback, and if there are folks on the team that should be in sync of this monorepo migration to continue supporting the [glean-probe-scraper.yml](https://github.com/mozilla-mobile/firefox-android/blob/ac-prep/android-components/.github/workflows/glean-probe-scraper.yml) workflow, please let me know. Thanks

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -44,10 +44,10 @@ libraries:
       - dthorn@mozilla.com
       - aplacitelli@mozilla.com
       - glean-team@mozilla.com
-      - android-components-team@mozilla.com
-    url: https://github.com/mozilla-mobile/android-components
+      - android-probes@mozilla.com
+    url: https://github.com/mozilla-mobile/firefox-android
     metrics_files:
-      - components/lib/crash/metrics.yaml
+      - android-components/components/lib/crash/metrics.yaml
     variants:
       - v1_name: lib-crash
         dependency_name: org.mozilla.components:lib-crash
@@ -56,13 +56,12 @@ libraries:
     description: Sync telemetry helper functionality
     notification_emails:
       - dthorn@mozilla.com
-      - lina@mozilla.com
-      - grisha@mozilla.com
-    url: https://github.com/mozilla-mobile/android-components
+      - android-probes@mozilla.com
+    url: https://github.com/mozilla-mobile/firefox-android
     metrics_files:
-      - components/support/sync-telemetry/metrics.yaml
+      - android-components/components/support/sync-telemetry/metrics.yaml
     ping_files:
-      - components/support/sync-telemetry/pings.yaml
+      - android-components/components/support/sync-telemetry/pings.yaml
     variants:
       - v1_name: sync
         dependency_name: org.mozilla.components:support-sync-telemetry


gabrielluong:
@badboy Any chance we could get a review for this? Also, we would like to proposal that we generate versioned releases of the probe-scraper actions so that we could land this and generate it's own release for it. Otherwise, we are currently consuming this action by pointing to the main branch https://searchfox.org/mozilla-mobile/search?q=probe-scraper&path=

badboy:
> @badboy Any chance we could get a review for this? Also, we would like to proposal that we generate versioned releases of the probe-scraper actions so that we could land this and generate it's own release for it. Otherwise, we are currently consuming this action by pointing to the main branch [searchfox.org/mozilla-mobile/search?q=probe-scraper&path=](https://searchfox.org/mozilla-mobile/search?q=probe-scraper&path=)

Sure. I'll take a look this week and see if there's anything we need to look out for to not break stuff.

badboy:
Running that, with the addition of specifying the branch, yields only the expected changes (different commits, different urls). Full diff here: https://gist.github.com/badboy/b1c8ad1821bddfaadf63ed46f26e015a
That means it will not break the generated schemas. 

For the action I loop in @relud. The data from it is not yet used in production, but soon will. I'm not 100% sure what the best way of landing this is.

relud:
> if there are folks on the team that should be in sync of this monorepo migration to continue supporting the [glean-probe-scraper.yml](https://github.com/mozilla-mobile/firefox-android/blob/ac-prep/android-components/.github/workflows/glean-probe-scraper.yml) workflow, please let me know.

> Running that, with the addition of specifying the branch, yields only the expected changes (different commits, different urls). Full diff here: https://gist.github.com/badboy/b1c8ad1821bddfaadf63ed46f26e015a That means it will not break the generated schemas.
> 
> For the action I loop in @relud. The data from it is not yet used in production, but soon will. I'm not 100% sure what the best way of landing this is.

merging this PR should fix that workflow, no other migration is needed.

relud:
I think CI might work if you re-push the branch

badboy:
I can take care of this today once the main branch has the right content (note that nothing should break even if we don't merge this today after the mode as the old repo still exists and contains all commits)

csadilek:
@badboy the `main` branch is ready. Can we land this now?

badboy:
```
python -m probe_scraper.runner --out-dir tmp/out --cache-dir tmp/cache --glean --glean-repo lib-crash --glean-repo sync --dry-run
```

produces nearly the same results across `main` and this PR. The only differences are the commit ids, the URLs and the timestamps, as expected. Rebasing and then I merge.

badboy (approved):

PR #505, opened 2022-10-19
Bug 1796067 - Consider needinfo a best-effort attempt

Dexterp37:
Instead of looking for specific needinfo errors, always try one more time without needinfos and hope for the best. The priorites for the probe-scraper are (a) to keep on scraping and (b) to do its best to file a bug about problems.


badboy (approved):
I'm ok with this, but lets have @chutten have another look

Dexterp37:
> I've been playing with the BMO API over in https://bugzilla.mozilla.org/show_bug.cgi?id=1792701 and think the "proper" fix would be to, on failure, check the [get user endpoint](https://bmo.readthedocs.io/en/latest/api/core/v1/user.html#get-user) for users who can log in (`can_login: true`) and retry with only those users needinfo'd. And _then_ file without any ni?s.

Thanks for the expedite review. I filed #506 for the proper/better approach you outlined!

Issue #506, opened 2022-10-19
Check if emails are Bugzilla users when needinfo-ing them


Dexterp37:
From #505 :

> I've been playing with the BMO API over in https://bugzilla.mozilla.org/show_bug.cgi?id=1792701 and think the "proper" fix would be to, on failure, check the [get user endpoint](https://bmo.readthedocs.io/en/latest/api/core/v1/user.html#get-user) for users who can log in (`can_login: true`) and retry with only those users needinfo'd. And _then_ file without any ni?s.

chutten:
We'll also want to tackle re-adding #508 at this point to Cc emails that are valid and active BMO users but have needinfos blocked (e.g. for PTO).

PR #508, opened 2022-10-26
Bug 1797446 - Don't cc folks as it might break

Dexterp37:
Bugzilla doesn't allow cc'ing non bugzilla accounts. cc'ing non bugzilla accounts will make the REST API call fail. I think we can safely skip cc'ing people as cc will be done automatically when adding a needinfo and the expectations is that bugs will be handled with triage once filed, so no cc is needed if we are not able to needinfo.


chutten (approved):
Yeah, fair enough. This is here to catch folks who are blocking needinfo, but are valid and active users (ie, those on PTO)... but this won't work properly until #506 reworks this logic.

I'll add a note to #506

relud:
looks like CI wasn't running on your fork, so I pushed the branch to the mozilla repo

PR #509, opened 2022-10-26
Remove jklukas from the codeowners

Dexterp37:
Otherwise it will mark the codeowners file as invalid.


badboy:
> This seems like a flaw - I'm able to approve changes to the codeowners, which is how we lock-down changes to the `git_scraper` and `repositories.yaml` files. Hmm...

Only listed files in `CODEOWNERS` get codeowner enforcement. Everything else is effectively `*`, so everyone with write access here.

PR #510, opened 2022-10-27
 Bug 1797446 - Don't cc folks as it might break

Dexterp37:
Bugzilla doesn't allow cc'ing non bugzilla accounts. cc'ing non bugzilla accounts will make the REST API call fail. I think we can safely skip cc'ing people as cc will be done automatically when adding a needinfo and the expectations is that bugs will be handled with triage once filed, so no cc is needed if we are not able to needinfo.


Dexterp37:
Duplicates #508 .

badboy (approved):

PR #511, opened 2022-10-27
Allow running expiry checks as a separate job

relud:
needs test fixes and telemetry-airflow PR


chutten (approved):
I'm not sure of the "why" of this change, but the architecture makes more sense than "let's run every day and then ask the calendar what day of the week it is to see if we should run more stuff" so I'm broadly in favour (Thus the r+)

relud:
> I'm not sure of the "why" of this change, but the architecture makes more sense

the main reason is that push mode can't send emails, so we need to make emails run as a separate airflow job that will continue to exist. the other reason is so that email and bug-filing failures don't interrupt schema updates.

PR #520, opened 2022-11-29
Bug 1802982 - ffx-ios: Add missing Storage/metrics.yaml file

badboy:
Added in https://github.com/mozilla-mobile/firefox-ios/pull/11634 and recently updated in https://github.com/mozilla-mobile/firefox-ios/pull/12306

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -359,6 +359,7 @@ applications:
     url: https://github.com/mozilla-mobile/firefox-ios
     metrics_files:
       - Client/metrics.yaml
+      - Storage/metrics.yaml
       # The Sync/* files are meant to be a stop-gap measure, see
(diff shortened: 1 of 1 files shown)


Dexterp37 (approved):
I wonder if there's any way for us, given that we now have a cloud function, to catch these.

badboy:
> I wonder if there's any way for us, given that we now have a cloud function, to catch these.

In the current way this works probe-scraper still only relies on the list.
There's 2 ways this could work:
1. probe-scraper searches for all metrics.yaml/pings.yaml files in the repository and compare that to the list and complains (or not and just takes all searched files)
2. the push action tells us about changed files and if those are metrics.yaml/pings.yaml probe-scraper reads them (or compares them to the known list and complains)

I'll see to get this filed as a bug for further discussion

PR #526, opened 2022-12-07
Bug 1804033 - Update to glean-parser 6.3.0

badboy:
That comes with a higher limit for allowed event extras, which is already used in FOG.


chutten (approved):
If CI's happy, I'm happy

PR #530, opened 2022-12-12
Adding collect_through_date to Rally studies

jepstein:
Rally studies are winding down.

All of the remaining Rally glean-based studies will end 3/27/2023.
I added a 12/31/2022 date for rally-study-zero-one just to be safe, though there shouldn't be any data being collected at this point.


whd (approved):
@relud I believe these will be subject to https://mozilla-hub.atlassian.net/browse/DENG-565, though there isn't any real urgency in landing the change as the collect through date is 3+ months out

PR #535, opened 2022-12-19
DENG-565 - Update pipeline metadata in nightly job

relud:
fixes [DENG-565](https://mozilla-hub.atlassian.net/browse/DENG-565)

the `generate_metadata` condition here matches the airflow task `probe_scraper.probe_scraper_glean_repositories`.

write per repo metadata files in that job to ensure changes in `repositories.yaml` are published nightly, instead of just when a repo's commits are next scraped (maybe never).

this can wait to merge until after the winter code freeze is lifted

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -136,32 +136,14 @@ def write_general_data(out_dir: Path) -> List[Path]:
     return [general_path, index_path]
 
 
-def write_glean_metric_data(
-    metrics: Dict[str, Any], dependencies: Dict[str, Any], out_dir: Path
-):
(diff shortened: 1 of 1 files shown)


probe_scraper/runner.py
```diff
@@ -518,21 +492,56 @@ def load_glean_metrics(
             if glean_repos is None and glean_urls is None:
                 raise
 
+    # currently always true, but left in for clarity
+    if scrape_commits or generate_metadata:
```
relud, line 496: and future bug resistance, should there become a case where this overlap isn't always true

badboy (approved):

PR #549, opened 2023-01-26
Bug 1804304 - Add new app: firefox.desktop.background.tasks

badboy:
https://phabricator.services.mozilla.com/D156941 has not yet landed, so the files don't actually exist yet.
We need to wait for this to land before merging.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1149,3 +1149,27 @@ applications:
     channels:
       - v1_name: treeherder
         app_id: treeherder
+
+  - app_name: firefox_desktop_background_tasks
+    canonical_app_name: Firefox Desktop background tasks
+    app_description: |
+      Background tasks of Firefox Desktop.
+      Currently used by `BackgroundTask_removeDirectory`.
+    url: https://github.com/mozilla/gecko-dev
+    notification_emails:
+      - necko@mozilla.com
+      - krosylight@mozilla.com
+      - vgosu@mozilla.com
+    branch: master
+    metrics_files:
+      - toolkit/components/backgroundtasks/metrics.yaml
+    ping_files:
+      - toolkit/components/backgroundtasks/pings.yaml
+    dependencies:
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -1149,3 +1149,25 @@ applications:
...
+    ping_files:
+      - toolkit/components/backgroundtasks/pings.yaml
+    dependencies:
+      - glean-core
+    channels:
+      - v1_name: firefox_desktop_background_tasks
+        app_id: firefox.desktop.background.tasks
+    moz_pipeline_metadata_defaults:
+      expiration_policy:
+        delete_after_days: 180
```
badboy, line 1173: Bug specifies 180 days.
I copied this from another app, but the docs on how to specify are not 100% clear.

To the reviewer: Is this the correct way to specify it?

relud: yeah, pretty sure this is right

relud (approved):

repositories.yaml
```diff
@@ -1149,3 +1149,25 @@ applications:
...
         app_id: treeherder
+
+  - app_name: firefox_desktop_background_tasks
+    canonical_app_name: Firefox Desktop background tasks
+    app_description: |
+      Background tasks of Firefox Desktop.
+      Currently used by `BackgroundTask_removeDirectory`.
+    url: https://github.com/mozilla/gecko-dev
+    notification_emails:
+      - krosylight@mozilla.com
```
chutten, line 1160: Maybe ask :nalexander if there are other emails/lists to include?

chutten (approved):

repositories.yaml
badboy: asking krosylight in the bug right now

badboy:
[Bug 1788986](https://bugzilla.mozilla.org/show_bug.cgi?id=1788986) landed.

PR #550, opened 2023-01-27
Revert "Bug 1804256 - Add Mozilla VPN (Android) as a separate channel of Mozilla VPN"

scholtzan:
Reverts mozilla/probe-scraper#541

discussion in https://mozilla.slack.com/archives/CKFKC7Y1Y/p1674839919027819?thread_ts=1674837813.590049&cid=CKFKC7Y1Y

PR #469, opened 2022-08-09
Move VPN Android app to be channel of main VPN app

sean-rose:
Both are from the same [mozilla-mobile/mozilla-vpn-client](https://github.com/mozilla-mobile/mozilla-vpn-client) repo, both use the same metrics and ping files (though I understand the different Glean SDKs used may cause minor differences), and the VPN team would like unified views of all the VPN app Glean data in BigQuery & Looker ([DENG-210](https://mozilla-hub.atlassian.net/browse/DENG-210)).

Based on a dry run, the only consequential result of this change is in the v2 API the Android app's `app_name` would change from `mozilla_vpn_android` to `mozilla_vpn` (its `app_id` would remain the same).

PR #552, opened 2023-02-06
Revert "Revert "Bug 1804256 - Add Mozilla VPN (Android) as a separate channel of Mozilla VPN""

scholtzan:
Reverts mozilla/probe-scraper#550

This will essentially apply the changes from https://github.com/mozilla/probe-scraper/pull/541 which we had to revert.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -829,6 +829,7 @@ applications:
     notification_emails:
       - vpn@mozilla.com
       - amarchesini@mozilla.com
+      - brizental@mozilla.com
     branch: main
     metrics_files:
       - glean/metrics.yaml
@@ -841,10 +842,13 @@ applications:
     channels:
       - v1_name: mozilla-vpn
         app_id: mozillavpn
+        app_channel: release
         additional_dependencies:
           - glean.js
       - v1_name: mozilla-vpn-android
         app_id: org.mozilla.firefox.vpn
+        app_channel: release
+        description: Mozilla VPN (Android)
         additional_dependencies:
           - glean-core
           - org.mozilla.components:service-glean
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -834,40 +834,20 @@ applications:
       - glean/metrics.yaml
     ping_files:
       - glean/pings.yaml
-    dependencies:
-      - glean-js
+    dependencies: []  # Empty. Dependencies are set per channel.
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 180
     channels:
       - v1_name: mozilla-vpn
         app_id: mozillavpn
-        app_channel: release
-
-  - app_name: mozilla_vpn_android
-    canonical_app_name: Mozilla VPN (Android)
-    app_description: |
-      Mozilla VPN is a VPN client application. The first
-      Mozilla premium service.
-    url: https://github.com/mozilla-mobile/mozilla-vpn-client
-    notification_emails:
-      - vpn@mozilla.com
-      - amarchesini@mozilla.com
-      - brizental@mozilla.com
-    branch: main
-    metrics_files:
-      - glean/metrics.yaml
-    ping_files:
-      - glean/pings.yaml
-    dependencies:
-      - glean-core
-    moz_pipeline_metadata_defaults:
-      expiration_policy:
-        delete_after_days: 180
-    channels:
+        additional_dependencies:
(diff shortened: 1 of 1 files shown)


relud (approved):

badboy (approved):
The only question I have is:
How can one differentiate the different apps in the unioned view?
We don't set a "channel" in the `channels` list (which is limited to nightly/beta/release right now anyway).
Is the app id or v1-name available in the dataset?

scholtzan:
Usually we differentiate them based on their channel and OS. But I think you are right, it would be handy to add the app IDs to the generated views. I'll open a PR.

sean-rose:
I also noticed that the VPN apps are sending the `client_info.channel` value "production", so it seems like clients can send any value there, and the VPN apps could send a differentiated value if they want (though then there'd be discontinuity with the historical pings).

repositories.yaml
```diff
@@ -834,40 +834,20 @@ applications:
...
-    dependencies:
-      - glean-js
+    dependencies: []  # Empty. Dependencies are set per channel.
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 180
     channels:
       - v1_name: mozilla-vpn
         app_id: mozillavpn
-        app_channel: release
```
sean-rose, line 845: IMO it'd be better to leave both channels with `app_channel: release`, which will help clarify that these are both actually release channels (not beta/nightly).

scholtzan: Good point. Was there a reason why you removed this in your initial PR, @badboy?

repositories.yaml
```diff
@@ -834,40 +834,20 @@ applications:
...
-        delete_after_days: 180
-    channels:
+        additional_dependencies:
+          - glean.js
       - v1_name: mozilla-vpn-android
         app_id: org.mozilla.firefox.vpn
-        app_channel: release
+        additional_dependencies:
+          - glean-core
+          - org.mozilla.components:service-glean
```
sean-rose, line 850: Why the extra dependency on `org.mozilla.components:service-glean`, which wasn't there before?

repositories.yaml
```diff
@@ -834,40 +834,20 @@ applications:
...
-  - app_name: mozilla_vpn_android
-    canonical_app_name: Mozilla VPN (Android)
-    app_description: |
-      Mozilla VPN is a VPN client application. The first
-      Mozilla premium service.
-    url: https://github.com/mozilla-mobile/mozilla-vpn-client
-    notification_emails:
-      - vpn@mozilla.com
-      - amarchesini@mozilla.com
-      - brizental@mozilla.com
```
sean-rose, line 856: Bea's email should be added to `notification_emails` for the newly consolidated app.

repositories.yaml
```diff
@@ -834,40 +834,20 @@ applications:
...
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 180
     channels:
       - v1_name: mozilla-vpn
         app_id: mozillavpn
-        app_channel: release
-
-  - app_name: mozilla_vpn_android
-    canonical_app_name: Mozilla VPN (Android)
```
sean-rose, line 848: I think to keep using this "Mozilla VPN (Android)" name for the Android app channel you can add a `description` to the channel to be used in place of the `canonical_app_name` (this is based on [me doing that in my abandoned PR doing this same thing](https://github.com/mozilla/probe-scraper/pull/469/commits/1f84e960d3faef15a3e6d2d128ae48c701a710bb#diff-a039c9b63812e8c69d4ad7a53ccb7c5b15f3721aadca48e869ca5653fa7b4229R827), where I tried to minimize the resulting changes).

repositories.yaml
badboy: I tried giving them _different_ channel names, that were not "release", "beta" or "nightly" and that failed, so I just removed it.
Having them both called "release" is the right choice though.

repositories.yaml
badboy: Oh, this might actually be correct. `service-glean` is the dependency that has `glean-core/android/metrics.yaml`, which has the `android_sdk_version` metric (though that's a `client_info` metric and so special anyway.

repositories.yaml
scholtzan: I'll add release back then

scholtzan:
Let's try this again

PR #555, opened 2023-02-15
Bug 1816937 - Restore fenix dependency on support-migration

relud:


repositories.yaml
```diff
@@ -288,6 +304,7 @@ applications:
       - org.mozilla.components:lib-crash
       - org.mozilla.components:support-sync-telemetry
       - org.mozilla.appservices:logins
+      - org.mozilla.components:support-migration
```
relud, line 307: dependencies need to be inclusive of history, or pings from dependencies will be deleted from schemas, which should only be done if we want the ping table deleted.

travis79: Thanks, this is good to know.

travis79 (approved):
I didn't realize the consequence of removing the dependency, I'll keep an eye out for this in the future. Is this documented somewhere?

chutten (approved):
LGTM as well.

relud:
> I didn't realize the consequence of removing the dependency, I'll keep an eye out for this in the future. Is this documented somewhere?

I didn't either, but this bug made me notice. I don't even know where this would documented if it were.

travis79:
> > I didn't realize the consequence of removing the dependency, I'll keep an eye out for this in the future. Is this documented somewhere?
> 
> I didn't either, but this bug made me notice. I don't even know where this would documented if it were.

@Dexterp37 was putting together a "run-book" for adding applications for ingestion, perhaps that would be a good place to make note of this?

Dexterp37:
@relud @travis79 I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1817246

PR #556, opened 2023-02-17
Add `glean-core` dependency to `mozillavpn` app

sean-rose:
It turns out the `mozillavpn` app also uses the Glean Rust SDK in addition to Glean.js.

CC: @brizental, @badboy, @scholtzan


repositories.yaml
```diff
@@ -845,12 +845,15 @@ applications:
...
           - glean-js
+          # It also uses the Rust SDK.
+          - glean-core
       - v1_name: mozilla-vpn-android
         app_id: org.mozilla.firefox.vpn
         app_channel: release
         description: Mozilla VPN (Android)
         additional_dependencies:
           - glean-core
+          # Extra dependency to have `android_sdk_version` recorded.
```
sean-rose, line 856: This isn't directly related, but I figured since I was adding a comment for the other dependency I might as well add a clarifying comment here after the [discussion in probe-scraper#552](https://github.com/mozilla/probe-scraper/pull/552/files#r1099860766).

repositories.yaml
```diff
@@ -845,12 +845,15 @@ applications:
         app_channel: release
         additional_dependencies:
           - glean-js
+          # It also uses the Rust SDK.
+          - glean-core
```
sean-rose, line 849: I'm curious how officially supported using multiple SDKs like this is.  This isn't just a different set of language bindings for the same core SDK code, [Glean.js](https://github.com/mozilla/glean.js) is a whole separate implementation, right?

With the pings from Glean.js and the Rust SDK going into the same tables it seems like there's a risk of incompatible schema changes/requirements between the two causing problems.

chutten:
Lemme just add @travis79 in here as someone who might better understand VPN's special sauce .

repositories.yaml
brizental: Yes, the SDKs are different implementation. However the ping schema is exactly the same. Unless there is some special casing based on telemetry agent that I am not aware about this should not be an issue.

brizental: FWIW on android specifically we have been using glean-core since last year and there were no issues AFAICT.

travis79: I don't think we have any other places where we've even tried to do something like this before. There's likely going to be some duplication between metrics in the metrics.yaml files found in those dependencies.

travis79: I'd be curious how the dry-run goes and how it sorts out the duplication of glean internal metrics between these two dependencies.

travis79 (commented):
I commented inline, just clearing the requested review.

Summary: Seems like there may be conflicts between the Glean internal metrics definitions between Glean.js and Glean-core.

repositories.yaml
brizental: If the dry run turns out to be a mess, I would suggest not merging this now. We have duplication at the moment because we are migrating all platforms to glean-core. As of right now the glean-core data is still to be validated. We can simply wait for validation final bugs to be fixed and Glean.js to be removed and then instaed of having both of them listed here we have only glean-core. 

The internal metrics are all the same, so this should not be a problem. Right?

sean-rose: > I'd be curious how the dry-run goes and how it sorts out the duplication of glean internal metrics between these two dependencies.

Trying a dry-run shows this would fail as-is due to duplicate metrics between glean-js and glean-core:
```
Glean has detected duplicated metric identifiers coming from the product 'mozilla-vpn'.

- 'app_build' defined more than once in glean-core, glean-js
- 'app_channel' defined more than once in glean-core, glean-js
- 'app_display_version' defined more than once in glean-core, glean-js
- 'architecture' defined more than once in glean-core, glean-js
- 'client_id' defined more than once in glean-core, glean-js
- 'device_manufacturer' defined more than once in glean-core, glean-js
- 'device_model' defined more than once in glean-core, glean-js
- 'end_time' defined more than once in glean-core, glean-js
- 'first_run_date' defined more than once in glean-core, glean-js
- 'locale' defined more than once in glean-core, glean-js
- 'os' defined more than once in glean-core, glean-js
- 'os_version' defined more than once in glean-core, glean-js
- 'seq' defined more than once in glean-core, glean-js
- 'start_time' defined more than once in glean-core, glean-js
- 'glean.error.invalid_label' defined more than once in glean-core, glean-js
- 'glean.error.invalid_overflow' defined more than once in glean-core, glean-js
- 'glean.error.invalid_state' defined more than once in glean-core, glean-js
- 'glean.error.invalid_value' defined more than once in glean-core, glean-js
- 'telemetry_sdk_build' defined more than once in glean-core, glean-js
- 'glean.restarted' defined more than once in glean-core, glean-js
- 'build_date' defined more than once in glean-core, glean-js
```

sean-rose: > @travis79 are we suggesting to do not merge this PR? If yes, I need to disable glean-scraper github action because currently it fails with out this PR. Are we OK with it?

@bakulf can you clarify what's failing?  The [VPN client's Glean probe-scraper action](https://github.com/mozilla-mobile/mozilla-vpn-client/actions/workflows/glean-probe-scraper.yml) doesn't seem to be failing.

sean-rose:
> @brizental 
> If the dry run turns out to be a mess, I would suggest not merging this now.

> @sean-rose 
> Trying a dry-run shows this would fail as-is due to duplicate metrics between glean-js and glean-core

Closing without merging.

PR #563, opened 2023-03-02
Add crash pings back to firefox-desktop-background-update

relud:
blocks [Bug 1815432](https://bugzilla.mozilla.org/show_bug.cgi?id=1815432)

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -242,8 +242,10 @@ applications:
       - chutten@mozilla.com
     metrics_files:
       - toolkit/mozapps/update/metrics.yaml
+      - toolkit/components/crashes/metrics.yaml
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -242,8 +242,10 @@ applications:
       - chutten@mozilla.com
     metrics_files:
       - toolkit/mozapps/update/metrics.yaml
+      - toolkit/components/crashes/metrics.yaml
     ping_files:
       - toolkit/mozapps/update/pings.yaml
+      - toolkit/components/crashes/pings.yaml
```
chutten, line 248: Oh, actually, don't these need to be lexicographically sorted?

PR #569, opened 2023-03-03
Skip backout commits for ipc/ipdl/metrics.yaml

relud:


chutten (approved):
Wow, up to five already?

Thank goodness it's in gecko so we only need to list it once.

Issue #364, opened 2021-11-08
Handle metrics.yaml files with different filenames


mdboom:
Though they are referred to as `metrics.yaml` and `pings.yaml` files in the Glean docs, they technically can have any filename, and there are legitimate (though currently unused) use cases for putting more than one of these files in the same directory.

It seems like the restriction that they are all called `metrics.yaml` is just perhaps a [check that is too strict](https://github.com/mozilla/probe-scraper/blob/6e4d48f1b60a19d14cc11d909da0fd2edcb400e2/schemas/repositories.json#L41) (but maybe there's something beyond that that makes this hard.

If possible, let's remove this restriction to be inline with what the Glean docs say.  

See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1738880

PR #576, opened 2023-03-31
Update file paths for the Mozilla-VPN repo

bakulf:


repositories.yaml
```diff
@@ -840,9 +840,14 @@ applications:
       - brizental@mozilla.com
     branch: main
     metrics_files:
-      - glean/metrics.yaml
+      - src/shared/telemetry/metrics.yaml
+      - src/shared/telemetry/metrics_deprecated.yaml
```
badboy, line 844: Unfortunately we currently enforce that files are named `metrics.yaml`, see also #364 
Let me figure out if we can now just lift that check.

badboy:
Using my admin powers.

badboy:
Oh, forgot to formally r+ this PR, but as I also merged that it should be clear.

PR #577, opened 2023-03-31
Bug 1738880 - Alllow any filename ending in `.yaml` for metrics and ping files

### LINKED ISSUE #364: Handle metrics.yaml files with different filenames

badboy:
Fixes #364

This will make #576 work

### DIFF
--- a/schemas/repositories.json
+++ b/schemas/repositories.json
@@ -38,14 +38,14 @@
         "type": "array",
         "items": {
           "type": "string",
-          "pattern": "metrics\\.yaml$"
+          "pattern": ".+\\.yaml$"
         }
       },
       "ping_files": {
         "type": "array",
         "items": {
           "type": "string",
-          "pattern": "pings\\.yaml$"
+          "pattern": ".+\\.yaml$"
         }
       },
       "dependencies": {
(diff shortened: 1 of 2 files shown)


schemas/repositories.json
```diff
@@ -38,14 +38,14 @@
         "type": "array",
         "items": {
           "type": "string",
-          "pattern": "metrics\\.yaml$"
+          "pattern": ".+\\.yaml$"
```
badboy, line 41: I don't think this file is actually used anymore?

relud (approved):

PR #580, opened 2023-04-03
Deprecate viu-politica and ignore commit that removed data collection

badboy:


badboy:
Failing on the VPN stuff, will be fixed in #576

Dexterp37:
Before merging this let me double check this is actually deprecated. Do we delete stuff if we mark things as deprecated?

travis79 (approved):
Looks good to me... provided @Dexterp37's concern is 👍

badboy:
> Looks good to me... provided @Dexterp37's concern is 👍

I already removed the deprecation, we can still do that afterwards.

PR #585, opened 2023-04-12
Update repositories.yaml to replace notification email for logins-store

tarikeshaq:
Noticed when looking at the repositories - 

rfkelly left Mozilla a while ago, and Lina (although is back 🪃, is probably not the best email) putting my own email instead.


chutten (approved):
Excellent. Love me a contact info update PR.

PR #589, opened 2023-05-03
Bug 1830504 - Add 2 new apps: accounts_frontend and accounts_backend

badboy:


travis79 (approved):
Looks good to me (once the app-ids are confirmed)

probe_scraper/runner.py
```diff
@@ -431,10 +428,24 @@ def load_glean_metrics(
...
+                metrics_files = [
+                    p
+                    for p in paths
+                    if any([str(p).endswith(fp) for fp in this_repo.metrics_file_paths])
+                ]
+                pings_files = [
+                    p
+                    for p in paths
+                    if any([str(p).endswith(fp) for fp in this_repo.ping_file_paths])
+                ]
```
badboy, line 448: We recently (#577) started allowing any file name, but we still had this hardcoded here.
We have the file names configured, so IMO we should use that knowledge to look for the right files.
This turned out to be a bit messy because of the double-iteration.

probe_scraper/runner.py
```diff
@@ -431,10 +428,24 @@ def load_glean_metrics(
         metrics = defaultdict(dict)
         pings = defaultdict(dict)
         for repo_name, commits in commits_by_repo.items():
+            this_repo = [r for r in all_repos if r.name == repo_name][0]
```
relud, line 431: this can retrun early if we use `next` instead:
```suggestion
            this_repo = next(r for r in all_repos if r.name == repo_name)
```

relud (approved):
r+wc

looks good, just suggesting a couple simple efficiency improvements

PR #593, opened 2023-05-26
Update where probe-scraper looks for sync metrics

skhamis:
The sync team is trying to consolidate metrics for both iOS and Android and thus requiring to move the metrics.yaml to application-services repo so that they can share the same file. This updates where the probe-scraper looks for the `metrics.yaml`

app-services PR (merged): https://github.com/mozilla/application-services/pull/5624
firefox-android PR: https://github.com/mozilla-mobile/firefox-android/pull/2117

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -57,17 +57,15 @@ libraries:
   - library_name: sync
     description: Sync telemetry helper functionality
     notification_emails:
-      - dthorn@mozilla.com
-      - android-probes@mozilla.com
-    url: https://github.com/mozilla-mobile/firefox-android
+      - sync-team@mozilla.com
+    url: https://github.com/mozilla/application-services
     metrics_files:
-      - android-components/components/support/sync-telemetry/metrics.yaml
+      - components/sync_manager/metrics.yaml
     ping_files:
-      - android-components/components/support/sync-telemetry/pings.yaml
+      - components/sync_manager/pings.yaml
     variants:
       - v1_name: sync
-        dependency_name: org.mozilla.components:support-sync-telemetry
-
+        dependency_name: org.mozilla.appservices:syncmanager
   - library_name: engine-gecko
     description: GeckoView metrics
     notification_emails:
@@ -313,7 +311,7 @@ applications:
       - glean-core
       - org.mozilla.components:service-glean
       - org.mozilla.components:lib-crash
-      - org.mozilla.components:support-sync-telemetry
+      - org.mozilla.appservices:syncmanager
       - org.mozilla.appservices:logins
       - org.mozilla.components:support-migration
       - org.mozilla.components:places
@@ -479,7 +477,7 @@ applications:
     dependencies:
       - glean-core
       - org.mozilla.components:service-glean
-      - org.mozilla.components:support-sync-telemetry
+      - org.mozilla.appservices:syncmanager
     channels:
       - v1_name: firefox-reality
         app_id: org.mozilla.vrbrowser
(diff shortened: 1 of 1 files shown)


Dexterp37:
@badboy isn't this related to some bug you "recently" looked at?

badboy:
Yup: https://bugzilla.mozilla.org/show_bug.cgi?id=1827949
Short of renaming the metric the metrics that clash we cannot support it.

repositories.yaml
```diff
@@ -57,17 +57,15 @@ libraries:
...
-      - android-components/components/support/sync-telemetry/metrics.yaml
+      - components/sync_manager/metrics.yaml
     ping_files:
-      - android-components/components/support/sync-telemetry/pings.yaml
+      - components/sync_manager/pings.yaml
     variants:
       - v1_name: sync
-        dependency_name: org.mozilla.components:support-sync-telemetry
-
+        dependency_name: org.mozilla.appservices:syncmanager
```
relud, line 68: CI is failing because this name also needs to be changed on fenix and firefox_reality

relud (dismissed):
r+wc

skhamis:
No worries! This won't land until after the main two land, appreciate the reviews! I wanted to ensure I get this up early to ensure I can course correct easier than fumbling around -- 

> Yup: https://bugzilla.mozilla.org/show_bug.cgi?id=1827949 Short of renaming the metric the metrics that clash we cannot support it.

yes indeed, once we make this move we'll rename the metrics with the `_v2` guidance and it'll be easier for us since we'll have both android and iOS sharing the same metrics.yaml so the renaming only needs to happen once

skhamis:
@badboy now that https://github.com/mozilla/application-services/pull/5624 landed, I imagine this needs to land before: https://github.com/mozilla-mobile/firefox-android/pull/2117 given that it's failing on that side due to the removal of the `sync-telemetry` module. Though I wonder if it's better to leave the sync-telemetry module in android for now and just switch the calls to app-services? Curious the best way from a glean perspective.

badboy:
> @badboy now that [mozilla/application-services#5624](https://github.com/mozilla/application-services/pull/5624) landed, I imagine this needs to land before: [mozilla-mobile/firefox-android#2117](https://github.com/mozilla-mobile/firefox-android/pull/2117) given that it's failing on that side due to the removal of the `sync-telemetry` module. Though I wonder if it's better to leave the sync-telemetry module in android for now and just switch the calls to app-services? Curious the best way from a glean perspective.
> 
> Edit: and remove the sync-telemetry module in a separate patch after everything lands.

A dry-run locally with this patch works. 
Switching over and removing the files in one go is fine, the data will also be fine, we never "forget" old information, so the system will continue to know about the old metrics just fine.

Is it ok if we land this only on Monday? While I'm nearly confident this all works out, I'd rather not have that merge before the weekend just in case.

badboy:
> > > @badboy now that [mozilla/application-services#5624](https://github.com/mozilla/application-services/pull/5624) landed, I imagine this needs to land before: [mozilla-mobile/firefox-android#2117](https://github.com/mozilla-mobile/firefox-android/pull/2117) given that it's failing on that side due to the removal of the `sync-telemetry` module. Though I wonder if it's better to leave the sync-telemetry module in android for now and just switch the calls to app-services? Curious the best way from a glean perspective.
> > > Edit: and remove the sync-telemetry module in a separate patch after everything lands.
> > 
> > 
> > A dry-run locally with this patch works. Switching over and removing the files in one go is fine, the data will also be fine, we never "forget" old information, so the system will continue to know about the old metrics just fine.
> > Is it ok if we land this only on Monday? While I'm nearly confident this all works out, I'd rather not have that merge before the weekend just in case.
> 
> No problem! Happy to land whenever the team is comfortable, [mozilla-mobile/firefox-android#2117](https://github.com/mozilla-mobile/firefox-android/pull/2117) got approved on that side -- so it's all ready to start landing!

Thanks, this got lost in the past 2 days because of other fires. Landing it now!

badboy (approved):

PR #597, opened 2023-05-31
Update Glean dependency for mozilla_vpn

brizental:
Ok, so. I thought it would be fine to just keep glean.js in there as a dependency until we validate the glean-core data and yank glean.js out (https://github.com/mozilla/probe-scraper/pull/556#event-9352957919).

Turns out that was a bad assesment, because now we are sending data with glean-core using the `metrics` ping, but there is no `mozilla_vpn.metrics` table generated.

If I am not mistaken -- calling on @badboy and @travis79 for a double check on this -- glean-js is just a subset of glean-core. So if we have both, we can just have glean-core and there will be no problems anymore. Is that correct?

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -857,7 +857,7 @@ applications:
         app_id: mozillavpn
         app_channel: release
         additional_dependencies:
-          - glean-js
+          - glean-core
       - v1_name: mozilla-vpn-android
         app_id: org.mozilla.firefox.vpn
         app_channel: release


relud (commented):
cc @sean-rose

badboy:
> Ok, so. I thought it would be fine to just keep glean.js in there as a dependency until we validate the glean-core data and yank glean.js out ([#556 (comment)](https://github.com/mozilla/probe-scraper/pull/556#event-9352957919)).
> 
> Turns out that was a bad assesment, because now we are sending data with glean-core using the `metrics` ping, but there is no `mozilla_vpn.metrics` table generated.
> 
> If I am not mistaken -- calling on @badboy and @travis79 for a double check on this -- glean-js is just a subset of glean-core. So if we have both, we can just have glean-core and there will be no problems anymore. Is that correct?

That's correct. This should work out and a dry-run runs through on this.

sean-rose (approved):
LGTM

sean-rose:
@brizental this is blocked waiting for a code owner to approve, and is a blocker for [DENG-996](https://mozilla-hub.atlassian.net/browse/DENG-996) "Recover mozilla_vpn metrics ping data".

relud (approved):

PR #606, opened 2023-07-05
Update VPN paths

brizental:
This change will be applied in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/7367, so I guess we need to block merging this on that.


badboy (approved):
Looks good. Indeed we should wait for the VPN PR to be merged, then we can merge here and afterwards all CI will be green again.

PR #610, opened 2023-07-20
Update PyYAML to avoid install issue due to cython

badboy:
Additionall update:
* awscli so it plays nice with the PyYAML update
* boto3 so it plays nice with the awscli update

---

Should fix the CI issues from e.g. #609 
See upstream issues:
* https://github.com/yaml/pyyaml/issues/724
* https://github.com/yaml/pyyaml/issues/728


badboy:
muhaha, now CI fails because of what #609 will fix.

travis79 (approved):

PR #617, opened 2023-08-28
Bug 1848664 - New app: monitor.cirrus

badboy:


travis79 (approved):

repositories.yaml
```diff
@@ -1252,3 +1252,23 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 760
+
+  - app_name: monitor_cirrus
```
jaredlockhart, line 1256: This should actually just say `monitor`

badboy: Yup, as said in the bug this is its own app in Glean/probe-scraper terms with its own app ID. These need to be unique and the way probe-scraper currently works is that metric files are per app.
Monitor's frontend metrics will be its own app with its own app id and a unique set of metrics files.

There's always the way to build derived datasets to merge data from different datasets/tables into one. If that's needed please file a bug with data engineering to take this on.

PR #618, opened 2023-08-31
Fix version detection for bug template modified in #591

relud:
fixes [bug 1851022](https://bugzilla.mozilla.org/show_bug.cgi?id=1851022)

### DIFF
--- a/probe_scraper/probe_expiry_alert.py
+++ b/probe_scraper/probe_expiry_alert.py
@@ -128,7 +128,10 @@ def find_existing_bugs(
 
     probes_with_bugs = {}
     for bug in found_bugs:
-        if re.search(r"release: version (\d+)", bug["description"]).group(1) != version:
+        if (
+            re.search(r"release: \[?version (\d+)", bug["description"]).group(1)
+            != version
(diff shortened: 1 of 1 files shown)


probe_scraper/probe_expiry_alert.py
```diff
@@ -128,7 +128,10 @@ def find_existing_bugs(
 
     probes_with_bugs = {}
     for bug in found_bugs:
-        if re.search(r"release: version (\d+)", bug["description"]).group(1) != version:
+        if (
+            re.search(r"release: \[?version (\d+)", bug["description"]).group(1)
```
relud, line 132: optionally match the bracket that now appears here https://github.com/mozilla/probe-scraper/blob/ade1d7c958e8a95af663f2992314beda50584fa0/probe_scraper/fog_checks.py#L79
but still does not appear here https://github.com/mozilla/probe-scraper/blob/ade1d7c958e8a95af663f2992314beda50584fa0/probe_scraper/probe_expiry_alert.py#L45

chutten (approved):
D'oh

PR #632, opened 2023-10-10
Bump the FF background task retention to 400 days

Dexterp37:
400 days matches better with keeping autoland build for 1 year.


saschanaz (approved):
Thanks!

badboy (approved):

PR #638, opened 2023-10-23
Mark deprecated apps as such

fbertsch:
Includes all Rally apps.


fbertsch:
See also: https://github.com/mozilla/probe-scraper/pull/639

PR #654, opened 2023-11-06
[Bug 1862755] Add moso-mastodon-backend app

scholtzan:
Adding `moso-mastodo-backend` app per https://bugzilla.mozilla.org/show_bug.cgi?id=1862755

A few questions below.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1381,3 +1381,23 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 760
+
+  - app_name: moso_mastodon_backend
+    canonical_app_name: Mozilla.Social Mastodon Backend
+    app_description: |
+      Mastodon instance for Mozilla.Social
+    url: https://github.com/MozillaSocial/mastodon
+    notification_emails:
+      - breyes@mozilla.com
+    branch: main
+    metrics_files:
+      - .glean/metrics.yaml
+    ping_files: []
+    dependencies:
+      - glean-core
+    channels:
+      - v1_name: moso-mastodon-backend
+        app_id: moso.mastodon.backend
+    moz_pipeline_metadata_defaults:
+      expiration_policy:
+        delete_after_days: 400
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1393,8 +1393,7 @@ applications:
     metrics_files:
       - .glean/metrics.yaml
     ping_files: []
-    dependencies:
-      - glean-core
+    dependencies: []
     channels:
       - v1_name: moso-mastodon-backend
         app_id: moso.mastodon.backend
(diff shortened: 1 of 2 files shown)


repositories.yaml
```diff
@@ -1381,3 +1381,23 @@ applications:
...
+    canonical_app_name: Mozilla.Social Mastodon Backend
+    app_description: |
+      Mastodon instance for Mozilla.Social
+    url: https://github.com/MozillaSocial/mastodon
+    notification_emails:
+      - breyes@mozilla.com
+    branch: main
+    metrics_files:
+      - .glean/metrics.yaml
+    ping_files: []
```
scholtzan, line 1395: Should there be a ping file in the repo?

repositories.yaml
```diff
@@ -1381,3 +1381,23 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 760
+
+  - app_name: moso_mastodon_backend
```
scholtzan, line 1385: I'm assuming from the Bugzilla ticket that this is for collecting backend data using Glean core? Which is why I changed the app ID to `moso.mastodon.backend`. I'm guessing we'll also want to collect frontend data using glean.js at some point for which we'd add a new app with the ID `moso.mastodon.frontend`.

Does this sound correct, @braunreyes?

braunreyes: Mastodon is a server-side html Rails app I just went with moso.mastodon since we are probably never going to instrument the front end aspects of the mastodon app.  That being said I am perfectly fine with being explicit here and can create a PR to update the mastodon code base as needed.  `moso.mastodon.backend` seems fine to me.

repositories.yaml
braunreyes: We are leveraging the 'events' ping which does not require a `pings.yaml` file per my understanding.  @Dexterp37 and @badboy can probably validate that for us.

repositories.yaml
scholtzan: I see, being consistent and adding a `_backend` is I think preferable here. We have done the same for FxA and some other applications as well.

repositories.yaml
Dexterp37: @scholtzan @braunreyes yes, that is correct (we do the same for the Glean Debug Ping Viewer for the same reason).

repositories.yaml
```diff
@@ -1381,3 +1381,23 @@ applications:
...
+      Mastodon instance for Mozilla.Social
+    url: https://github.com/MozillaSocial/mastodon
+    notification_emails:
+      - breyes@mozilla.com
+    branch: main
+    metrics_files:
+      - .glean/metrics.yaml
+    ping_files: []
+    dependencies:
+      - glean-core
```
akkomar, line 1397:
```suggestion
    dependencies: []
```
This is server side collection that does not use Glean SDK.

badboy: Woops, thanks for noticing! I should have paid more attention to that.

scholtzan:
It looks like the data review hasn't been completed yet. Should we wait with merging this until that is done?

scholtzan:
Data review has been completed. Merging this now

PR #664, opened 2023-11-28
Skip sending expiry emails for deprecated repositories

### LINKED ISSUE #401: Don't send expiry emails for deprecated applications
Deprecated applications might contain expired metrics.
Given they are deprecated it's unlikely these will get updates or metric removals. We should just skip sending emails for them all together.

badboy:
Fixes #401

---

I think that's really all that's needed.


Dexterp37 (changes requested):
@badboy any chance you could kindly add test coverage for this? 🙈 Seems like test_glean_checks.py might be a good spot for it.

badboy:
> @badboy any chance you could kindly add test coverage for this? 🙈 Seems like test_glean_checks.py might be a good spot for it.

Actually we had another check in test_git_scraper, which was easier to duplicate. Otherwise I would have to set up new fixtures and ... meh.

Dexterp37 (approved):

PR #670, opened 2023-12-12
change mozillavpn_cirrus to mozillavpn_backend_cirrus

Marlene-M-Hirose:
change app_name, app_id, and v1_name in repositories.yaml

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -965,7 +965,7 @@ applications:
         additional_dependencies:
           - glean-core
 
-  - app_name: mozillavpn_cirrus
+  - app_name: mozillavpn_backend_cirrus
     canonical_app_name: Mozilla VPN Cirrus Sidecar
     app_description: |
       The sidecar Cirrus container for the Mozilla VPN server
@@ -983,8 +983,8 @@ applications:
       expiration_policy:
         delete_after_days: 180
     channels:
-      - v1_name: mozillavpn-cirrus
-        app_id: mozillavpn_cirrus
+      - v1_name: mozillavpn-backend-cirrus
+        app_id: mozillavpn_backend_cirrus
 
   - app_name: rally_study_zero_one
     canonical_app_name: Rally Study-01


sean-rose (approved):

repositories.yaml
```diff
@@ -983,8 +983,8 @@ applications:
       expiration_policy:
         delete_after_days: 180
     channels:
-      - v1_name: mozillavpn-cirrus
-        app_id: mozillavpn_cirrus
+      - v1_name: mozillavpn-backend-cirrus
+        app_id: mozillavpn_backend_cirrus
```
yashikakhurana, line 987: 👍

repositories.yaml
```diff
@@ -965,7 +965,7 @@ applications:
         additional_dependencies:
           - glean-core
 
-  - app_name: mozillavpn_cirrus
+  - app_name: mozillavpn_backend_cirrus
```
yashikakhurana, line 968: 👍

yashikakhurana (approved):

badboy:
This will create new tables, won't it? But they are empty, so we can just drop them and I assume data, if any, has already been sent with the updated app ID anyway?
(Only other thing then is that we're now running into the change freeze today)

sean-rose:
> This will create new tables, won't it? But they are empty, so we can just drop them and I assume data, if any, has already been sent with the updated app ID anyway? (Only other thing then is that we're now running into the change freeze today)

Yes, the live & stable tables for the `mozillavpn_cirrus` app ID are still empty.

The [DE change freeze](https://docs.google.com/document/d/1bK9-nx_mAAb_-vhZ0NImmLZ2Nk1gM7nhsqHJ87kguDI/edit) does specifically call out "no new Glean apps", which this app rename would technically qualify as, so I guess this will have to wait until Jan 3 (sorry @brizental).

yashikakhurana:
@sean-rose just wondering can we merge this now?

sean-rose:
> @sean-rose just wondering can we merge this now?

In terms of the change freeze we're clear to merge, but we still need a code owner to approve this PR.  I've [asked in Slack](https://mozilla.slack.com/archives/C01E8GDG80N/p1704403354838329).

travis79 (approved):

PR #676, opened 2024-01-09
[no-bug] Adding some commits to skip from gecko

travis79:
Some gecko metrics.yaml files seemed to be missing from certain commits in my local dry-runs, not sure exactly why but these are the commits that need to be skipped.

### DIFF
--- a/probe_scraper/scrapers/git_scraper.py
+++ b/probe_scraper/scrapers/git_scraper.py
@@ -70,6 +70,16 @@ SKIP_COMMITS = {
         "0e55b6d34c8fac3144f10f9aa450e33e4b55d520",  # Missing ipc/ipdl/metrics.yaml
         "0b1543e85d13c30a13c57e959ce9815a3f0fa1d3",  # Missing ipc/ipdl/metrics.yaml
         "9bc20993bc6960762ed281201e9cff437a88ca6c",  # Missing ipc/ipdl/metrics.yaml
+        "3f6ba0d4adbdf9d3e81b7047ff4c21384abbd234",  # Missing dom/base/use_counter_metrics.yaml
+        "1a7724cfd6b3cce2c599e323afb14f31430e5acd",  # Missing dom/base/use_counter_metrics.yaml
+        "b16c6e1f04e563c916fb43b62661fdc0d354a925",  # Missing security/manager/ssl/metrics.yaml
+        "02731904bba2c2f4e1c043e45a492bb21b33a930",  # Missing security/manager/ssl/metrics.yaml
+        # Missing toolkit/components/reportbrokensite/metrics.yaml
+        "42acdc9cd5ae89222bdceeeaed7bacac755be48f",
+        # Missing toolkit/components/reportbrokensite/metrics.yaml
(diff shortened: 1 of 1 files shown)


travis79:
cc @badboy what are your thoughts on this? Is this normal due to some uplifts or something? Is there something we can do to catch and automate this sort of thing?

badboy:
A full dry-run is LOOOOONG, which is why we don't do it in CI.
It's also not even necessarily exactly what runs in airflow, so it might still not be perfect anyway.

As said in channel:
adding so many skip commits feels wrong, so maybe we should look into how to handle that better; but if that clears it up for the time being I won't stand in the way of course
but we should involve other DEs for visibility

badboy (changes requested):
The ACTUAL dry-run you want to be running is this:

```
python3 -m probe_scraper.runner --dry-run --cache-dir tmp/cache --out-dir tmp/out --glean --glean-url=https://github.com/mozilla/gecko-dev --output-bucket=gs://probe-scraper-prod-artifacts/ --glean-limit-date=2024-01-09 --update
```

because that closely resembles what is run nightly in airflow.
Notice the `glean-limit-date` (and update). That teaches probe-scraper to go back at most to that date. We incrementally add new metrics and don't have to re-parse old commits.
Yes, that means a dry-run over everything in gecko-dev won't work as is.

I don't think fixing that dry-run-over-everything case is worth it here.
It will happen again in the future and we won't notice.

So really the next steps should be:
* Document that one should dry-run with a limit-date
* Require from everyone to include the exact command they ran when a dry-run failed.

travis79:
> The ACTUAL dry-run you want to be running is this:
> 
> ```
> python3 -m probe_scraper.runner --dry-run --cache-dir tmp/cache --out-dir tmp/out --glean --glean-url=https://github.com/mozilla/gecko-dev --output-bucket=gs://probe-scraper-prod-artifacts/ --glean-limit-date=2024-01-09 --update
> ```
> 
> because that closely resembles what is run nightly in airflow. Notice the `glean-limit-date` (and update). That teaches probe-scraper to go back at most to that date. We incrementally add new metrics and don't have to re-parse old commits. Yes, that means a dry-run over everything in gecko-dev won't work as is.
> 
> I don't think fixing that dry-run-over-everything case is worth it here. It will happen again in the future and we won't notice.
> 
> So really the next steps should be:
> 
>     * Document that one should dry-run with a limit-date
> 
>     * Require from everyone to include the exact command they ran when a dry-run failed.

Oh, I ran that plenty of times....  but I've had unusual things (like commits that need skipped) show up after a PR has landed that breaks probe-scraper so I've gotten into the habit of doing full dry-runs locally to give myself some piece of mind.

badboy:
glean-limit-date is set for the nightly run: https://github.com/mozilla/telemetry-airflow/blob/03274fdad2ade74de6b51a36e4e4fb7f15985240/dags/probe_scraper.py#L169-L173

travis79:
Based on your assurances that this isn't something that could break in production thanks to the `--glean-limit-date`, I'm going to close this PR as unnecessary.

PR #683, opened 2024-02-06
Bug 1878674 - Exempt metrics moving out of geckoview-streaming from duplication checks

perrymcmanis144:
This is a fix as we migrate metrics out of Geckoview Streaming and is intended to be removed once that effort is complete.

As per https://github.com/mozilla/probe-scraper/blob/main/repositories.yaml#L69 engine-gecko is the library for Geckoview.

### DIFF
--- a/probe_scraper/glean_checks.py
+++ b/probe_scraper/glean_checks.py
@@ -138,7 +138,11 @@ def check_for_duplicate_metrics(repositories, metrics_by_repo, emails):
                     metric_sources.setdefault(metric_name, []).append(dependency)
 
         duplicate_sources = dict(
-            (k, v) for (k, v) in metric_sources.items() if len(v) > 1
+            # Exempt cases when one of the sources is Geckoview Streaming to
+            # avoid false positive duplication accross app channels.
+            (k, v)
+            for (k, v) in metric_sources.items()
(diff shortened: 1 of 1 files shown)


badboy:
What's the command you ran locally that breaks on `main` and runs with this?

perrymcmanis144:
> What's the command you ran locally that breaks on `main` and runs with this?

`python -m probe_scraper.runner --glean --glean-repo glean-core --glean-repo glean-android --glean-repo engine-gecko --glean-repo fenix-nightly --glean-repo gecko --cache-dir ./.scraper_cache/ --dry-run`

badboy (approved):
The given probe-scraper command indeed does work with this now.
I'm not super happy about the special-casing, but once GVST is gone, `engine-gecko` is gone.

I also tested the output of this against MSG:

```
mozilla-schema-generator generate-glean-pings --out-dir out --repo org-mozilla-fenix-nightly --pretty --mps-branch main
```

(needs a patch in `mozilla_schema_generator/generic_ping.py` to change the `probe_info_base_url` to localhost).
The generated schema looks good.
Differences are only the changed description for the migrated GVST metrics (note: because the probe-scraper run left out some dependencies, those will be missing in the generated schema, e.g. fxa metrics. This is not an issue for the full run, which has all dependencies)

PR #693, opened 2024-02-19
pine: Remove xre/metrics.yaml to avoid new dupes

badboy:
Seems that does the trick.

### DIFF
--- a/probe_scraper/glean_checks.py
+++ b/probe_scraper/glean_checks.py
@@ -16,6 +16,14 @@ from probe_scraper import probe_expiry_alert
 
 from .scrapers.git_scraper import Commit
 
+# Ugly hack to skip certain metrics which we know aren't duplicated,
+# but show up duplicated due to them being moved from one
+# to the other application/library.
+SKIP_METRICS = {
+    "gecko.version": ["gecko", "pine", "firefox-desktop"],
+    "gecko.build_id": ["gecko", "pine", "firefox-desktop"],
+}
+
 
 def _metric_sort_key(metric: Dict[str, Any]):
     return (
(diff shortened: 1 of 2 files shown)


Dexterp37 (approved):
@badboy isn't Pine deprecated? If so, should we even be doing this?

badboy:
> @badboy isn't Pine deprecated? If so, should we even be doing this?

It fails right now and this should help it getting unstuck.
Right now the dupe checks do not check the deprecated flag. I _think_ we can add a check there, but this PR here was the quicker way to ensure we can land that today without further help.

akkomar (approved):

badboy:
gnaaaah, it's actually a bit more complicated. We moved `toolkit/xre/metrics.yaml` from `firefox-desktop` to `gecko`

badboy:
I think I know what's going on (and this is a hassle to clean up now):

* On 2024-02-10 [`gecko.build_id` and `gecko.version` landed in `toolkit/xre/metrics.yaml`](https://hg.mozilla.org/mozilla-central/rev/68e91a8da283)
* This was scraped shortly after, because `toolkit/xre/metrics.yaml` was listed for `firefox-desktop` (and `pine`). It was thus considered a new metric for those apps
* On [2024-02-14 I merged the move of that file to gecko](https://github.com/mozilla/probe-scraper/pull/689). Now that causes the dupe check to consider both metrics "alive" and thus clash. 

We know that the metric is not actually duplicated, so everything should be fine. We just need to teach probe-scraper that.
We could "hack" it by just skipping that specific metric for those specific applications. It's an ugly hack.
I'm checking if there's a better reliable way to do this now.

badboy:
```
python -u -m probe_scraper.runner --dry-run --cache-dir tmp/cache --out-dir tmp/out --glean --update --glean-repo gecko --glean-repo pine --glean-repo firefox-desktop --output-bucket=gs://probe-scraper-prod-artifacts/ --glean-limit-date 2024-02-18 
```

works now with those 2 commits.
I dislike everything about the second commit, but I have no better way to cleanly tackle this today.

PR #695, opened 2024-02-20
Update test dependencies

badboy:


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
 # See http://pep8.readthedocs.io/en/latest/intro.html#configuration
-max-line-length = 99
+max-line-length = 100
 filename = *.py, +.lint
-exclude = probe_scraper/parsers/third_party/*
+exclude = probe_scraper/parsers/third_party/* .venv
```
BenWu, line 5: We're using the non-hidden `venv/` in our .ignores and readme in this repo.  Although I don't mind adding .venv as well
```suggestion
exclude = probe_scraper/parsers/third_party/* venv
```

BenWu (approved):
One minor comment

PR #708, opened 2024-03-12
Switch `mozillavpn_backend_cirrus` to use `nimbus-cirrus` dependency (bug 1867770)

sean-rose:
## [Bug 1867770](https://bugzilla.mozilla.org/show_bug.cgi?id=1867770): Enable new Glean App `mozillavpn_cirrus`

There should be no functional change, this just makes it clearer what's happening and avoids duplicating the Cirrus metrics/ping files config for VPN.

### DIFF
--- a/probe_scraper/scrapers/git_scraper.py
+++ b/probe_scraper/scrapers/git_scraper.py
@@ -399,6 +399,17 @@ def scrape(
     upload_repos = []
 
     for repo_info in repos:
+        if not (
+            repo_info.metrics_file_paths
+            or repo_info.ping_file_paths
+            or repo_info.tag_file_paths
+        ):
+            print(
+                f"Skipping commits for repository {repo_info.name}"
+                " because it has no metrics/ping/tag files."
+            )
+            continue
+
         print("Getting commits for repository " + repo_info.name)
 
         commits_by_repo[repo_info.name] = {}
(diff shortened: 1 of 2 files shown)


Dexterp37 (approved):

repositories.yaml
```diff
@@ -968,16 +968,15 @@ applications:
     canonical_app_name: Mozilla VPN Cirrus Sidecar
     app_description: |
       The sidecar Cirrus container for the Mozilla VPN server
-    url: https://github.com/mozilla/experimenter
+    url: https://github.com/mozilla-services/guardian-website
```
badboy, line 971: That repo doesn't exist (or is not public). probe-scraper will fail on that. Should this be just https://github.com/mozilla-mobile/mozilla-vpn-client ?

probe_scraper/scrapers/git_scraper.py
```diff
@@ -399,6 +399,17 @@ def scrape(
...
+        if not (
+            repo_info.metrics_file_paths
+            or repo_info.ping_file_paths
+            or repo_info.tag_file_paths
+        ):
+            print(
+                f"Skipping commits for repository {repo_info.name}"
+                " because it has no metrics/ping/tag files."
+            )
+            continue
```
sean-rose, line 411: This would currently skip the following 6 repos: `engine-gecko`, `engine-gecko-beta`, `engine-gecko-nightly`, `lockwise-ios`, `mozillavpn-backend-cirrus`, `monitor-cirrus`

badboy: I guess that should be fine indeed.

badboy (approved):

PR #712, opened 2024-03-14
Set commits_by_repo with repo name before skipping commits

BenWu:
Follow-up to #708.  This currently fails for repos with no ping/metrics/tag files because it fails the schema validation in `glean_checks.check_glean_metric_structure` due to no items in `commits_by_repo`.  This initializes `commits_by_repo` to `{"repo_name": {}}` instead of an empty dict.  Alternative is to allow an empty dict but I think it makes more sense to list the repos explicitly with no commits to scrape.  Output is the same either way

### DIFF
--- a/probe_scraper/scrapers/git_scraper.py
+++ b/probe_scraper/scrapers/git_scraper.py
@@ -399,6 +399,14 @@ def scrape(
     upload_repos = []
 
     for repo_info in repos:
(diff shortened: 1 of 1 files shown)


sean-rose (approved):

BenWu:
@badboy Do you mind reviewing as a code owner?

PR #717, opened 2024-03-21
Bug 1874935 - Add new glean-server library to define pings for server-side Glean

badboy:
Right now this just re-uses the Glean SDK repo and its defintions of all builtin pings.
Technically server-side Glean would only send the `events` ping.

Should we:

1. ~~Create a `server-side-pings.yaml` in the Glean SDK to reference here?~~
2. Create a `server-side-pings.yaml` in the `glean_parser` repo to reference here (as that's where all the code for server-side lives)
3. ~~Create a `mozilla/glean-server` repository that only exists for that file?~~

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -21,6 +21,19 @@ libraries:
         branch: main
         dependency_name: glean-core
 
+  - library_name: glean-server
+    description: Modern cross-platform telemetry (server usage)
+    notification_emails:
+      - glean-team@mozilla.com
+    url: https://github.com/mozilla/glean_parser
+    metrics_files: []
+    ping_files:
+      - server-side-pings.yaml
+    variants:
+      - v1_name: glean-server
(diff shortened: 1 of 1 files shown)


akkomar:
I think either of the alternative options is preferable over reusing `glean-core/pings.yaml` to avoid showing unused pings in Glean Dictionary. Since the server component does not use the SDK I'd lean towards option 2, but they all seem fine and are easy to change later.

akkomar (approved):

badboy:
```
python3 -m probe_scraper.runner --dry-run --cache-dir tmp/cache --out-dir tmp/out --glean --glean-limit-date 2024-04-01 --output-bucket=gs://probe-scraper-prod-artifacts/ --glean-repo glean-server --glean-repo accounts-backend --glean-repo moso-mastodon-backend
```

looking good.

PR #723, opened 2024-04-08
Upgrade gsutil dependency

badboy:
This makes probe-scraper run on Python 3.12
Previous versions of gsutil load in `boto`, which is not compatible with Python 3.12


BenWu (approved):
Note that the image version that CI is using has gsutil 5.12 ([release notes](https://cloud.google.com/sdk/docs/release-notes#39800_2022-08-16)).  This shouldn't matter, especially since CI doesn't use gsutil, but possibly something to keep in mind for the future

PR #728, opened 2024-04-16
Bug 1891420 - Add the abbrevation WDBA so we know what it is

badboy:


chutten (approved):
LGTM. @ncalexan, this is the right one?

chutten:
> LGTM. @ncalexan, this is the right one?

Ah, Jan-Erik points me to https://bugzilla.mozilla.org/show_bug.cgi?id=1891420#c3 which confirms this is indeed the correct one. Unping.

PR #731, opened 2024-04-22
feat(bug-1891646): add firefox_bridge app to repositories.yaml

kik-kik:
# feat(bug-1891646): add firefox_bridge app to repositories.yaml

Bug link: https://bugzilla.mozilla.org/show_bug.cgi?id=1891646

[ ] Make sure the project repository has [Validate and publish metrics](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/enable-data-ingestion.html#validate-and-publish-metrics) Github Action set up.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1568,3 +1568,25 @@ applications:
     channels:
       - v1_name: tiktokreporter-android
         app_id: org.mozilla.tiktokreporter
+
+  - app_name: firefox_bridge
+    canonical_app_name: Firefox Bridge
+    app_description: |
+      This is a chromium & firefox extension that allows the user to easily swap
+      between browsers on their device.
+    url: https://github.com/mozilla-extensions/firefox-bridge
+    notification_emails:
+      - ahabibi@mozilla.com
+    branch: main
+    metrics_files:
+      - "./metrics.yaml"
+    ping_files:
+      - "./pings.yaml"
+    dependencies:
+      - glean-js
+    moz_pipeline_metadata_defaults:
+      expiration_policy:
+        delete_after_days: 400
+    channels:
+      - v1_name: firefox-bridge
+        app_id: org.mozilla.firefox-bridge
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1572,16 +1572,16 @@ applications:
   - app_name: firefox_bridge
     canonical_app_name: Firefox Bridge
     app_description: |
-      This is a chromium & firefox extension that allows the user to easily swap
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -1568,3 +1568,25 @@ applications:
...
+    metrics_files:
+      - "./metrics.yaml"
+    ping_files:
+      - "./pings.yaml"
+    dependencies:
+      - glean-js
+    moz_pipeline_metadata_defaults:
+      expiration_policy:
+        delete_after_days: 400
+    channels:
```
kik-kik, line 1590: I'm not sure if this is correct.

repositories.yaml
```diff
@@ -1568,3 +1568,25 @@ applications:
     channels:
       - v1_name: tiktokreporter-android
         app_id: org.mozilla.tiktokreporter
+
+  - app_name: firefox_bridge
+    canonical_app_name: Firefox Bridge
+    app_description: |
+      This is a chromium & firefox extension that allows the user to easily swap
+      between browsers on their device.
```
badboy, line 1576: we should slightly edit this, it's what shows up in the dictionary:

```suggestion
      A Chromium & Firefox extension that allows the user to easily swap
      between browsers on their device.
```

repositories.yaml
```diff
@@ -1568,3 +1568,25 @@ applications:
...
+    url: https://github.com/mozilla-extensions/firefox-bridge
+    notification_emails:
+      - ahabibi@mozilla.com
+    branch: main
+    metrics_files:
+      - "./metrics.yaml"
```
badboy, line 1582: Unusual but not wrong. Seems so far we haven't wrapped these in quotes or used a leading `./`

repositories.yaml
badboy: app id they use is `firefox.bridge`

badboy (changes requested):
Couple of things. There's also no data review for the extension at all and the metrics.yaml has dummy values. I'll reach out and clear that up. Hold off merging this for now.

repositories.yaml
kik-kik: I just copied the entry above. I removed the `./` and the `"`.

kik-kik:
According to the most recent response in the bug the project was put on hold, but should resume in the future.

kik-kik:
@badboy do you know if this is still needed or if I can just close this?

badboy:
this is on halt. We can close this for now.

PR #736, opened 2024-04-30
Bug 1894172 - Add geoip skipping to MDN

Dexterp37:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1894172 for more info.

@akkomar how can we make sure this will produce the right results?

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1159,6 +1159,7 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 1130
+      geoip_skip_entries: 1
     channels:
       - v1_name: mdn-yari
         app_id: mdn-yari


akkomar:
> @akkomar how can we make sure this will produce the right results?

Easiest would be to check the distribution of `distinct metadata.geo.country, metadata.geo.city` in the `events` ping after this change is picked up by the pipeline (tomorrow by default).

One caveat is that the parameter value might depend on the proxy configuration used by MDN. IIRC when we were rolling this out for accounts_frontend, we were expecting `2` entries need to be skipped, but eventually had to switch to `1`. I think the reason was some difference between FxA's load balancer configuration and the one we used while testing this feature earlier. Since in this case geo-decoded data is already broken I wouldn't worry about this and just go with the value proposed in this PR.

akkomar (approved):

PR #746, opened 2024-05-06
Bug 1874935 - Roll back dependency updates to backend apps

akkomar:
This should fix `probe_scraper.mozilla_schema_generator` Airflow task that is currently failing.

The reason for failure is metrics added to FxA in https://github.com/mozilla/fxa/pull/16864 not being picked up by probe scraper. This is caused by the wrong schema identifier in the metrics file. This will get fixed in https://github.com/mozilla/fxa/pull/16872/files.

Nevertheless, these metrics still will not go through probe scraper because their names start with `glean` and are treated as internal by glean_parser. I'd rather revert here to the last known working configuration and think about a fix tomorrow.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1389,7 +1389,6 @@ applications:
     branch: main
     metrics_files:
       - packages/fxa-shared/metrics/glean/fxa-backend-metrics.yaml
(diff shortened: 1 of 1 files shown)


badboy (approved):

akkomar:
Superseded by https://github.com/mozilla/probe-scraper/pull/747 and other linked to the same bug.

PR #755, opened 2024-05-20
Set spoc retention to the same value across glean apps

whd:
https://mozilla-hub.atlassian.net/browse/DENG-3842


Dexterp37 (approved):
Confirmed by @gleonard-m here https://mozilla-hub.atlassian.net/browse/DENG-3842?focusedCommentId=884221

PR #760, opened 2024-05-29
Write overrides file to support app-level pipeline metadata overrides

BenWu:
Goes with https://github.com/mozilla/mozilla-schema-generator/pull/266 to fix https://github.com/mozilla/mozilla-schema-generator/issues/264

This creates a new file called `overrides` that would look like:
```json
{
  "moz_pipeline_metadata_overrides": {
    "pageload": {
      "expiration_policy": {
        "delete_after_days": 120
      }
    }
  }
}
```
to be used by the schema generator to apply `moz_pipeline_metadata` set at the app level to pings defined in a dependency.

I went with a new file instead of `pings` because having mostly empty ping entries might have side effects in consumers like the glean dictionary.  Also pings duplicated with the dependency.  It's not the cleanest solution though so I'm wondering if there's something smarter

https://github.com/mozilla/mozilla-schema-generator/pull/266 is how it's used

### DIFF
--- a/probe_scraper/runner.py
+++ b/probe_scraper/runner.py
@@ -251,6 +251,7 @@ def add_pipeline_metadata_defaults(repositories):
 
 
 def add_pipeline_metadata(pings_by_repo, repositories):
+    """Add pipeline metadata to pings, applying metadata defaults from the repo."""
     for repo in repositories:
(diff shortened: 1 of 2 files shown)


BenWu:
Not needed, this can be done just in the schema generator

PR #763, opened 2024-06-06
mozsocial-android removed Glean integration

badboy:
Removal done here: https://github.com/MozillaSocial/mozilla-social-android/pull/570

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1553,6 +1553,7 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 1110  # 37 months
+    deprecated: true
 
   - app_name: tiktokreporter_ios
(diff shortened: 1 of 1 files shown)


Dexterp37 (approved):
Why do we need to remove the file given that it's deprecated?

badboy:
> Why do we need to remove the file given that it's deprecated?

because `probe_scraper.check_repositories` doesn't skip deprecated repos (I'll file a bug).

badboy:
Oh wait, it does, we just need the deprecated in the right place: https://github.com/mozilla/probe-scraper/blob/51e50ed7beb9ccb5751dc4e68e53531edcad9916/probe_scraper/check_repositories.py#L35

badboy:
D'oh! it was still deprecated, #727 did only half the job of undoing that.

PR #771, opened 2024-06-19
Add explicit retention to all glean apps

whd:
https://mozilla-hub.atlassian.net/browse/SRELEAD-182?focusedCommentId=897713

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -359,6 +359,24 @@ applications:
         expiration_policy:
           delete_after_days: 180
 
+      baseline:
+        expiration_policy:
+          delete_after_days: 10000
+      deletion-request:
+        expiration_policy:
+          delete_after_days: 10000
+      events:
+        expiration_policy:
+          delete_after_days: 10000
+      fog-validation:
+        expiration_policy:
+          delete_after_days: 10000
+      metrics:
+        expiration_policy:
+          delete_after_days: 10000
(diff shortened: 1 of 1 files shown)


whd:
Deployment plan in https://mozilla-hub.atlassian.net/browse/SRELEAD-182?focusedCommentId=901771

badboy (approved):
Should this become a required field in the schema? (`probeinfo_api.yaml`)

akkomar (approved):

akkomar:
> Should this become a required field in the schema? (`probeinfo_api.yaml`)

Seems like we should unless we want to repeat this process at some point.

whd:
>  Should this become a required field in the schema? (probeinfo_api.yaml)

It probably should. If I set that to required will CI automatically fail if it's not specified?

At any rate, I want to land this as-is before making it required or adding a CI check since I deliberately skipped some apps that should be removed in the near future (rally).

badboy:
> > Should this become a required field in the schema? (probeinfo_api.yaml)
> 
> It probably should. If I set that to required will CI automatically fail if it's not specified?
> 
> At any rate, I want to land this as-is before making it required or adding a CI check since I deliberately skipped some apps that should be removed in the near future (rally).

It should! And yeah, should have been explicit: landing this as a followup sounds good to me!

PR #778, opened 2024-07-09
Expire `user-characteristics` ping after 90 days (fixes #759)

### LINKED ISSUE #759: Add `delete_after_days` for the user-charecteristics ping
The [user characteristics ping](https://searchfox.org/mozilla-central/source/toolkit/components/resistfingerprinting/pings.yaml#8-21) is supposed to have an extra-limited retention window of 90 days (and in the future we might try to trim that to 30) - but I don't know if this is codified in the repositories.yaml file correctly.
```[tasklist]
### Tasks
```

mozfreddyb:
This adds a section for the `user-characteristics` ping in Firefox Desktop in repositories.yaml to ensure that data is retained for no longer than 90 days.

@sean-rose Please review :)

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -358,6 +358,9 @@ applications:
       spoc:
         expiration_policy:
           delete_after_days: 180
+      user-characteristics:
+        expiration_policy:
+          delete_after_days: 90
 
   - app_name: firefox_desktop_background_update
     canonical_app_name: Firefox for Desktop Background Update Task


mozfreddyb:
Aside: This is desktop only.

I'm trying to read between the lines what is left to be done, but if someone reviewing in here could help me to clearly identify if the `user-characteristics` ping is a also available mobile, we should probably add an expiration here too. I just didn't really get where I should be looking.. 😕

chutten:
> Aside: This is desktop only.
> 
> I'm trying to read between the lines what is left to be done, but if someone reviewing in here could help me to clearly identify if the `user-characteristics` ping is a also available mobile, we should probably add an expiration here too. I just didn't really get where I should be looking.. 😕

The Glean Dictionary'd give you what you wanted by looking at [e.g. the list of Firefox for Android pings](https://dictionary.telemetry.mozilla.org/apps/fenix?itemType=pings), but we can also look for where the ping is defined (`toolkit/components/resistfingerprinting/pings.yaml`) and see where that file is included in the dependency tree in `repositories.yaml`. It's in `firefox_desktop`'s `ping_files` list, and that's an `application` so it's the root of its dependency tree, meaning that it's only ingested from applications using the `firefox_desktop` application id (which, you guessed it, is only Firefox Desktop).

So it's desktop-only.

The rule to follow would be to include the metadata about the ping's dataset's retention in the same application or library definition that has the file the ping's defined in in its `ping_files` list. Which you did!

sean-rose (approved):

whd (approved):
Apparently Sean's not a CODEOWNER but this LGTM

PR #784, opened 2024-07-16
Revert "Update glean-parser requirement from ~=14.1.0 to ~=14.2.0"

badboy:
Reverts mozilla/probe-scraper#782


whd:
Looking at https://bugzilla.mozilla.org/show_bug.cgi?id=1657947#c7 I don't think reverting this will fix the issue (the error is [coming from MSG](https://github.com/mozilla/mozilla-schema-generator/blob/main/mozilla_schema_generator/__main__.py#L187-L212)), but the work @chutten is doing today should.

PR #790, opened 2024-08-08
Bug 1910995 - Enable TB desktop

Dexterp37:
Permissions will be adjusted as part of [this ticket](https://mozilla-hub.atlassian.net/browse/DO-1787).

The dry-run runs locally without any problem.

@badboy , this is a similar configuration to Firefox Desktop, so we don't have a way to run the GH workflow ([docs](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/enable-data-ingestion.html#validate-and-publish-metrics)). How does it currently work in Firefox? Do you remember?

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1990,3 +1990,33 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 400
+
+  - app_name: thunderbird_desktop
+    canonical_app_name: Thunderbird
+    app_description: Thunderbird email client
+    url: https://github.com/mozilla/releases-comm-central
+    notification_emails:
+      - alessandro@thunderbird.net
+      - sancus@thunderbird.net
+    metrics_files:
+      - mail/metrics.yaml
+      - mail/components/compose/metrics.yaml
+      - mail/components/addrbook/metrics.yaml
+      - mail/components/cloudfile/metrics.yaml
+      - calendar/metrics.yaml
+    # While bug 1910995 asks to use comm/mail/pings.yaml, this
+    # file is currently empty in the repo and we won't add it
+    # to prevent failures, as probe-scraper doesn't like empty
+    # files.
+    ping_files: []
+    tag_files:
+      - mail/tags.yaml
+    dependencies:
+      - gecko
+      - glean-core
+    channels:
+      - v1_name: thunderbird-desktop
+        app_id: thunderbird.desktop
+    moz_pipeline_metadata_defaults:
+      expiration_policy:
+        delete_after_days: 420


repositories.yaml
```diff
@@ -1990,3 +1990,33 @@ applications:
...
+      - calendar/metrics.yaml
+    # While bug 1910995 asks to use comm/mail/pings.yaml, this
+    # file is currently empty in the repo and we won't add it
+    # to prevent failures, as probe-scraper doesn't like empty
+    # files.
+    ping_files: []
+    tag_files:
+      - mail/tags.yaml
+    dependencies:
+      - gecko
```
Dexterp37, line 2015: @chutten the bug only asks for glean-core to be added. Do we need to add "gecko" as well?

badboy:
> @badboy , this is a similar configuration to Firefox Desktop, so we don't have a way to run the GH workflow ([docs](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/enable-data-ingestion.html#validate-and-publish-metrics)). How does it currently work in Firefox? Do you remember?

We have a nightly task that runs against gecko-dev.
It should be good enough to add a new entry here:
https://github.com/mozilla/telemetry-airflow/blob/cebb534a136a09e25f7c422e43f597faf105370f/dags/probe_scraper.py#L207-L210

repositories.yaml
chutten: Yes. Either we include `gecko` as a dep or the `additional_properties` column will be quite large as all of the reusable components of gecko will record data into builtin pings regardless of whether they're in Thunderbird or Firefox

Dexterp37: I will leave the gecko data points then!

repositories.yaml
```diff
@@ -1990,3 +1990,33 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 400
+
+  - app_name: thunderbird_desktop
+    canonical_app_name: Thunderbird
+    app_description: Thunderbird email client
+    url: https://github.com/mozilla/releases-comm-central
```
chutten, line 1997: The github repo description says this repo is EXPERIMENTAL. Have we done everything we need to do to notify folks that we're placing a dependency on an their experimental repo? And what will happen should that repo go away or fall out of date?

repositories.yaml
```diff
@@ -1990,3 +1990,33 @@ applications:
...
+    metrics_files:
+      - mail/metrics.yaml
+      - mail/components/compose/metrics.yaml
+      - mail/components/addrbook/metrics.yaml
+      - mail/components/cloudfile/metrics.yaml
+      - calendar/metrics.yaml
+    # While bug 1910995 asks to use comm/mail/pings.yaml, this
+    # file is currently empty in the repo and we won't add it
+    # to prevent failures, as probe-scraper doesn't like empty
+    # files.
```
chutten, line 2010: Boo. It should at least have a schema decl. File a follow-up to remove the file and remove this comment.

chutten (approved):
r+wc

repositories.yaml
Dexterp37: Yes, I asked for confirmation over slack that that's indeed the official mirror. I now asked for confirmation that, though marked as experimental, we can rely on it.

repositories.yaml
Dexterp37: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1912270 !

PR #800, opened 2024-09-12
fog-update: Fix current CI failings

badboy:


badboy:
> Had to dig into the commit messages to understand what was going on, but now that I know why these are important I can unhesitatingly r+

GRRRRRR github, better show commit messages, damn!

PR #803, opened 2024-09-16
feat(bug-1918744): Add firefox_crashreporter app to repositories.yaml

kik-kik:
# feat(bug-1918744): Add firefox_crashreporter app to repositories.yaml

This is in order to resolve: https://bugzilla.mozilla.org/show_bug.cgi?id=1915732

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -410,7 +410,7 @@ applications:
       The Firefox desktop crash reporter client.
     url: https://github.com/mozilla/gecko-dev
     notification_emails:
-      - chutten@mozilla.com
+      - afranchuk@mozilla.com
     metrics_files:
       - toolkit/components/crashes/metrics.yaml
     ping_files:
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -428,7 +428,7 @@ applications:
     moz_pipeline_metadata:
       crash:
         expiration_policy:
-          delete_after_days: 10000
+          delete_after_days: 1130
 
   - app_name: firefox_desktop_background_update
     canonical_app_name: Firefox for Desktop Background Update Task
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -403,6 +403,32 @@ applications:
         expiration_policy:
           delete_after_days: 1130
 
+  - app_name: firefox_crashreporter
+    canonical_app_name: Firefox Crash Reporter
+    app_description: >-
+      The Firefox desktop crash reporter client.
+    url: https://github.com/mozilla/gecko-dev
+    notification_emails:
+      - chutten@mozilla.com
```
badboy, line 412: This should be Alex' email, he's the one responsible in the end.

repositories.yaml
```diff
@@ -403,6 +403,32 @@ applications:
...
+    channels:
+      - v1_name: firefox-crashreporter
+        app_id: firefox.crashreporter
+    moz_pipeline_metadata_defaults:
+      expiration_policy:
+        delete_after_days: 400
+    moz_pipeline_metadata:
+      crash:
+        expiration_policy:
+          delete_after_days: 1130
```
badboy, line 430: Why the difference here?
(Impl-wise: this crashreporter won't actually send internal pings, so the only existing ping _is_ the crash ping)

badboy (changes requested):

repositories.yaml
afranchuk: The crashreporter may send some other pings in the future (though perhaps unlikely on a long-term basis; probably only for experiments).

badboy: fine with keeping this then.

badboy (approved):
```
python3 -m probe_scraper.runner --dry-run --cache-dir tmp/cache --out-dir tmp/out --glean --output-bucket=gs://probe-scraper-prod-artifacts/ --glean-repo glean-core --glean-repo firefox-crashreporter --glean-limit-date 2024-09-17
```

runs clean.

PR #826, opened 2024-11-01
Bug 2064428 - Add Backstage app

quiiver:
Adding our new backstage app

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -2309,6 +2309,3 @@ applications:
       - v1_name: mozcloud-backstage-prod
         app_id: moz_backstage
         app_channel: production
-    moz_pipeline_metadata_defaults:
-      expiration_policy:
-        delete_after_days: 400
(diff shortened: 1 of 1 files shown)


repositories.yaml
```diff
@@ -2091,3 +2091,15 @@ applications:
...
+    canonical_app_name: Mozilla Service Catalog
+    app_description: Mozilla Service Catalog (Backstage)
+    url: https://github.com/mozilla-services/moz-backstage-app
+    notification_emails:
+      - wstuckey@mozilla.com
+    metrics_files:
+      - plugins/glean/metrics.yaml
+    channels:
+      - v1_name: moz-backstage
+        app_id: moz_backstage
```
whd, line 2105: We don't yet require this value in CI/make it required in the schema but you need to set retention for this app e.g.:
```yaml
      expiration_policy:
        delete_after_days: 400
```

badboy:
Fixing the treeherder issue in https://github.com/mozilla/probe-scraper/pull/827

badboy (changes requested):
Isn't this a Glean.js-powered app?
If so this should have

```
    dependencies:
      - glean-js
```

badboy (approved):

PR #829, opened 2024-11-07
fog-updater: Fix some issues.

badboy:
Every commit documents the changes on its own


chutten (approved):
/me reads the pygithub and gh rest API docs for what `requested_reviewers` does.

Ah, not needed because we auto-apply all the reviewers on PR creation? Very good.

chutten:
(which is indeed what your commit message says, but github makes it so hard to see commit _messages_ (gosh, even commit _subjects_ take effort to locate) that I didn't think to look until after combing through the docs)

PR #840, opened 2024-11-27
Reduce max table expiration days from 10,000 to 9,985

sean-rose:
Tables can have a maximum of 10,000 partitions, but when partitions expire they apparently continue to count toward the table's partition quota for up to 14 days afterwards (7 days of time travel and 7 days of fail-safe), per the [BigQuery docs](https://cloud.google.com/bigquery/docs/managing-partitioned-tables#partition-expiration):
> When a partition expires, BigQuery deletes that partition. The partition data is retained in accordance with [time travel](https://cloud.google.com/bigquery/docs/time-travel) and [fail-safe](https://cloud.google.com/bigquery/docs/time-travel#fail-safe) policies, and can be charged for, depending on your billing model. Until then, the partition counts for purposes of [table quotas](https://cloud.google.com/bigquery/quotas#partitioned_tables).

And I subtracted one additional day just to make sure we don't hit the table partition limit.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -413,28 +413,28 @@ applications:
           delete_after_days: 90
       baseline:
         expiration_policy:
-          delete_after_days: 10000
+          delete_after_days: 9985
       crash:
         expiration_policy:
           delete_after_days: 1130
       deletion-request:
         expiration_policy:
-          delete_after_days: 10000
+          delete_after_days: 9985
       events:
         expiration_policy:
-          delete_after_days: 10000
+          delete_after_days: 9985
       first-startup:
         expiration_policy:
           delete_after_days: 1130
       fog-validation:
         expiration_policy:
-          delete_after_days: 10000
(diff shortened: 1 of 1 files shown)


kwindau (approved):
Don't see the harm in this, seems like a good idea

sean-rose:
> Academically this makes sense but as a practical matter I'm not sure it's worth landing this.

IMO if we're going to set an explicit upper limit it should be something that would actually work in practice, or we should not set the theoretical upper limit at all, but if the consensus is to leave it as is that's fine.

> If we do we should update https://github.com/mozilla-services/mozilla-pipeline-schemas/ as well to match.

Thanks for catching this. I did file https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/833 in case we opt to proceed.

whd:
Per slack discussion deferring decisions on this PR until meetings next week

PR #846, opened 2024-12-19
Change ping expiry Jira link to DENG instead of DO

BenWu:
Changing the link Jira to DENG instead of data org because the pings are a data infra concern so we can triage them.


BenWu:
Confirmed with Rob that we should do this

scholtzan (approved):

PR #849, opened 2025-01-06
Bug 1940030 - Ignore broken_site_report.* reports for dupe-detection

badboy:
These are initially defined in gecko (though this will change), but the existing code was only ever sending them from Firefox Desktop. With recent changes they are redefined in Fenix and submitted from there too.

Because the tables were already generated based on the old schema we can receive them just fine.
We can thus skip them in the dupe-detection check.

This hack (along with the gecko.* metrics also in that list) can hopefully go away once m-c changed to define the metrics for firefox-desktop instead.

---

This works in a dry-run.

### DIFF
--- a/probe_scraper/glean_checks.py
+++ b/probe_scraper/glean_checks.py
@@ -22,6 +22,140 @@ from .scrapers.git_scraper import Commit
 SKIP_METRICS = {
     "gecko.version": ["gecko", "pine", "firefox-desktop"],
     "gecko.build_id": ["gecko", "pine", "firefox-desktop"],
+    "broken_site_report.breakage_category": ["gecko", "fenix", "firefox-desktop"],
+    "broken_site_report.description": ["gecko", "fenix", "firefox-desktop"],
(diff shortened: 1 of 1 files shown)


chutten (approved):
: (

badboy:
I did the following test on `main`:

* Manually move `toolkit/components/reportbrokensite/metrics.yaml` and `toolkit/components/reportbrokensite/pings.yaml`. That's essentially what https://phabricator.services.mozilla.com/D233238 will do
* Run: `python3 -m probe_scraper.runner --dry-run --cache-dir tmp/cache --out-dir tmp/out --glean --output-bucket=gs://probe-scraper-prod-artifacts/ --glean-limit-date 2025-01-04 --glean-repo fenix --glean-repo gecko --glean-repo firefox_desktop`

This succeeds.
Without the move it will fail with the dupe errors.
With this PR and no modified `repositories.yaml` it will also succeed (as it ignores the dupes).

Merging this and the m-c patch, afterwards I should be able to clean it up again.

PR #861, opened 2025-01-30
Add the user characteristics deletion policy to fenix also

tomrittervg:


chutten:
FWIW Fenix isn't the only application that _can_ upload these pings. Since the pings are included in Gecko, they can be sent anywhere that Gecko is (Focus, Klar, Thunderbird, etc). 

( See https://mozilla-hub.atlassian.net/browse/DENG-7765 for what we're doing about this )

@tomrittervg Would you like to set these retention periods defensively right now, or wait for DENG-7765 to provide clearer guidance about what to do about pings coming from shared libraries?

tomrittervg:
Better safe than sorry, I added it to everything with gecko except the background-updater service...

chutten:
Alrighty, I've pulled down this PR and will give it a dry-run like so: `python -m probe_scraper.runner --glean --glean-repo glean-core --glean-repo glean-android --glean-repo gecko --glean-repo firefox-desktop-background-update --glean-repo firefox-desktop --glean-repo fenix --dry-run --cache-dir .scraper_cache/` 

(it may take a little while as it updates the latest git cache for all these very popular repos)

chutten (approved):

chutten:
Dry-run passes! (admittedly with some interesting errors (missing `description`! How does that even _happen_?) on pdfjs metrics on old commits, which aren't a problem. And some additional commits to skip because of bouncing metrics.yaml files.)

PR #870, opened 2025-02-25
Update metadata

whd:
https://mozilla-hub.atlassian.net/browse/DSRE-1928

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -433,37 +433,37 @@ applications:
           delete_after_days: 90
       baseline:
         expiration_policy:
-          delete_after_days: 10000
+          delete_after_days: 775
       crash:
         expiration_policy:
-          delete_after_days: 1130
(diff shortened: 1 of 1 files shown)


chutten (approved):
With immense relish I do `r+`.

repositories.yaml
```diff
@@ -1309,7 +1184,7 @@ applications:
       - nimbus
     moz_pipeline_metadata_defaults:
       expiration_policy:
-        delete_after_days: 720
+        delete_after_days: 755
```
BenWu, line 1187: Is this an intentional 755 instead of 775?

Increasing the retention will (should) cause the expiry alerts to start sending again for the already expiring pings so worth keeping in mind in case anyone asks

BenWu (approved):

kik-kik (approved):

kwindau (approved):

PR #885, opened 2025-04-07
Enable SubPlat backend

edugfilho:


repositories.yaml
```diff
@@ -2028,3 +2028,25 @@ applications:
...
+    branch: main
+    metrics_files:
+      - libs/shared/metrics/glean/src/registry/subplat-backend-metrics.yaml
+    ping_files: []
+    dependencies:
+      - glean-server
+      - glean-js
+    moz_pipeline_metadata_defaults:
+      expiration_policy:
+        delete_after_days: 10000
```
edugfilho, line 2049: not sure if this is the right thing. It seems like the usual is to put a smaller default and then 10k for specific pings, but the bug did not specify. Should I ask?

repositories.yaml
```diff
@@ -2028,3 +2028,25 @@ applications:
...
+      Mozilla Subscription Platform is Mozilla's subscription management
+      solution for account-based end-user services and features.
+    url: https://github.com/mozilla/fxa
+    notification_emails:
+      - dalvarez@mozilla.com
+    branch: main
+    metrics_files:
+      - libs/shared/metrics/glean/src/registry/subplat-backend-metrics.yaml
+    ping_files: []
+    dependencies:
```
edugfilho, line 2044: They're not sure in the bug. I added glean-server because @Dexterp37 mentioned that it seems like it. But how can I tell?
I also added glean-js because the bug mentions glean on NPM.

edugfilho:
CI is failing because Mercurial is returning 406 while downloading certain revisions. The same thing happens to me when I do a dry-run locally without my changes. It is not consistent which revision gets the error. Is this a common thing that we just have to wait and it goes away?

repositories.yaml
Dexterp37: @edugfilho this is a good question for the product. I agree with you, 10k (27 years!?) seems like a weirdly high number. Maybe a typo?

repositories.yaml
Dexterp37: @edugfilho I don't think it can be both glean-server and glean-js. I was assuming backend because I though subplat was backend, but might be wise to ask the requester.

Dexterp37:
> CI is failing because Mercurial is returning 406 while downloading certain revisions. The same thing happens to me when I do a dry-run locally without my changes. It is not consistent which revision gets the error. Is this a common thing that we just have to wait and it goes away?

This is news to me. @badboy have you ever seen this?

badboy:
> CI is failing because Mercurial is returning 406 while downloading certain revisions. The same thing happens to me when I do a dry-run locally without my changes. It is not consistent which revision gets the error. Is this a common thing that we just have to wait and it goes away?

That's a new one and seems unrelated to this PR. [It worked last week](https://app.circleci.com/pipelines/github/mozilla/probe-scraper/2373/workflows/d8b4c036-cb82-48cd-bb3f-b07d316d8b2c/jobs/5752)

repositories.yaml
akkomar: 10k is probably a copy from some other app. These are default values that were put in when we started implementing finite retention and effectively serve as placeholders for indefinite retention. I'd recommend [760 days to keep it consistent with fxa](https://github.com/mozilla/probe-scraper/blob/4373ec1137e17e03337d78edcec42c28a150b6c6/repositories.yaml#L1631).

badboy:
> > CI is failing because Mercurial is returning 406 while downloading certain revisions. The same thing happens to me when I do a dry-run locally without my changes. It is not consistent which revision gets the error. Is this a common thing that we just have to wait and it goes away?
> 
> That's a new one and seems unrelated to this PR. [It worked last week](https://app.circleci.com/pipelines/github/mozilla/probe-scraper/2373/workflows/d8b4c036-cb82-48cd-bb3f-b07d316d8b2c/jobs/5752)

It works for me locally, at least in the last run. It keeps failing on CI.
406 is _supposed_ to be a client-error, but I have the feeling we get locked out for some reason, probably by a stricter WAF or something. I'll look into it.

badboy:
The error looks like something bigger now, I can't access hg.mozilla.org here at all

badboy:
Ok, so that specific error we're getting is unfortunate. Let's ignore it for this PR (I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1959103)

edugfilho:
Thanks for filing the bug @badboy 
I made the changes to dependencies and retention. It looks like they've landed a fix for hg. I'll re-try CI

badboy:
yup, if you rebase it should pass on CI now (#886 )

badboy (approved):

PR #893, opened 2025-04-14
Bug 1959916 - Remove dependencies from Focus Android/iOS

badboy:
These apps do not send general telemetry anymore, only 2 custom pings.

**Question**: Will anything break by us completely removing the glean-core dependency?


akkomar (approved):
> **Question**: Will anything break by us completely removing the glean-core dependency?

I think this is equivalent to removing metrics so nothing should break downstream. To be sure we can re-run probe_scraper DAG after this is merged.

PR #901, opened 2025-05-01
Use current Nightly version in metric expiry bug details

DonalMe:
This fixes the version used in the wording of bugs

### DIFF
--- a/probe_scraper/fog_checks.py
+++ b/probe_scraper/fog_checks.py
@@ -209,7 +209,7 @@ def file_bugs(
 
     metrics_to_bug_numbers = probe_expiry_alert.file_bugs(
         expiring_probes,
-        next_version,
+        str(latest_nightly_version),
         bugzilla_api_key,
         dry_run,
         BUG_WHITEBOARD_TAG,


probe_scraper/fog_checks.py
```diff
@@ -277,7 +277,10 @@ def file_bugs_and_get_emails_for_expiring_metrics(
         addresses = list(metrics_addresses)
 
         filed_bugs: Dict[str, List[str]] = file_bugs(
-            expiring_metrics, latest_nightly_version, bugzilla_api_key, dry_run
+            expiring_metrics,
+            str(int(latest_nightly_version)),
```
chutten, line 281: I'm confused. This won't really change the value. I figured the changes'd be in `fog_checks.file_bugs` to no longer pass `next_version` to `probe_expiry_alert.file_bugs` which, as of #897, assumes it's being called with the latest nightly version, not the one after it.

chutten (changes requested):

chutten (approved):

chutten:
This may refile a couple of bugs, but there's only been three of them so I'll deal with it when it comes to it.

DonalMe:
> This may refile a couple of bugs, but there's only been three of them so I'll deal with it when it comes to it.

Thanks for that!

PR #909, opened 2025-05-30
fog-updater: Also check firefox-ios for new metric files

badboy:
cc @ih-codes

Does not yet work until firefox-ios lands https://github.com/mozilla-mobile/firefox-ios/pull/26910

### DIFF
--- a/fog-updater/src/fog_update.py
+++ b/fog-updater/src/fog_update.py
@@ -11,6 +11,7 @@ import io
 import os
 import requests
 import sys
+import yaml
 
 DEFAULT_ORGANIZATION = "mozilla"
 DEFAULT_AUTHOR_NAME = "data-updater"
@@ -20,6 +21,7 @@ HTTP_HEADERS = {
     "user-agent": "probe-scraper/1.0",
 }
 INDEX_URL = "https://raw.githubusercontent.com/mozilla-firefox/firefox/main/toolkit/components/glean/metrics_index.py"  # noqa
+FFX_IOS_INDEX_URL = "https://raw.githubusercontent.com/mozilla-mobile/firefox-ios/main/firefox-ios/Client/Glean/glean_index.yaml"  # noqa
 BODY_TEMPLATE = f"""This (automated) patch updates the list from metrics_index.py.
 
 For reviewers:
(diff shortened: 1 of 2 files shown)


fog-updater/src/fog_update.py
```diff
@@ -180,6 +188,11 @@ def main(argv, repo, author, debug=False, dry_run=False):
     background_tasks_metrics = sorted(data["background_tasks_metrics"])
     background_tasks_pings = sorted(data["background_tasks_pings"])
 
+    ios_metrics_index = get_latest_ios_metrics_index()
+    data = yaml.safe_load(ios_metrics_index)
+    firefox_ios_pings = ["firefox-ios/" + f for f in sorted(data["ping_files"])]
+    firefox_ios_metrics = ["firefox_ios/" + f for f in sorted(data["metrics_files"])]
```
ih-codes, line 194: Ohh, would it make more sense for me to prepend the root directory to the path so you don't have to?

ih-codes (commented):
Looks great from the iOS side, thanks for the quick work on this @badboy ! 

I'm thinking maybe we should also include checking [the `tags.yaml` path as well](https://github.com/mozilla/probe-scraper/blob/dba2c450debf005fd6ebe31428cbe0d7a3b3dd2d/repositories.yaml#L820) via the `glean_index.yaml`'s `tags_files` key. (That has also changed in my PR... probably won't change much from our side, but just in case!)

fog-updater/src/fog_update.py
badboy: Then we wouldn't need to hard-code it here, yeah.

ih-codes:
I'm just waiting for a ✅ on my PR on the iOS side. You should be able to remove the `firefox_ios` from your path now [after my update](firefox-ios/Client/Glean/glean_index.yaml).

Once we're both good, would you like me to merge my iOS PR first? After yours is merged, then I believe the probe-scraper scripts on the iOS side will start passing again. 🤞 Hopefully we can sync up on this tomorrow or Wednesday.

Question for you. So your GitHub action triggers this bot to run each UTC morning. If we add a new metrics file, that will open a PR here. Do you know if the iOS probe-scraper action (which runs https://github.com/mozilla/probe-scraper/blob/main/.github/workflows/glean.yaml) will fail until that is merged? Or did this action [only fail in the past](https://github.com/mozilla-mobile/firefox-ios/actions/runs/15400175342/job/43330634212?pr=26910) because I was juggling around folders and the preexisting file paths became incorrect?

badboy:
> Just thought of one more question @badboy -- to confirm, if we delete a metrics YAML file (thus causing the probe-scraper to fail), will things self-correct the next time the `fog_update.py` runs or will we be required to manually open a PR?

They will self-correct.
I expect deleting files will only be an issue in the early stages as you split up files. It will be much more rare afterwards I assume. So that's a little bit of a hiccup in the beginngin, but I guess that's ok.

PR #926, opened 2025-06-25
Bug 1974239 - Enable new Glean app: mdn-fred

caugner:
This is mostly the same as mdn-yari, except that it only uses automatic instrumentation (page views, and clicks).

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -1511,7 +1511,7 @@ applications:
         app_id: glean-dictionary
 
   - app_name: mdn_fred
-    canonical_app_name: Mozilla Developer Network
+    canonical_app_name: MDN
     app_description: |
       Fred is the 2025 web frontend of MDN, an open-source, collaborative
       project documenting Web platform technologies, including CSS, HTML,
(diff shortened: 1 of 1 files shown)


badboy:
Is there a bug/ticket associated with this?

badboy:
> > Is there a bug/ticket associated with this?
> 
> There is not, but I can create one. Do you have any preference (Bugzilla, GitHub or JIRA)?

Best to file a DE bug: https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/enable-data-ingestion.html#add-your-product-to-probe-scraper

You can skip most of the details and call out that what you said here: that it's essentially the same as mdn-yari.
Just so we have it on file.

repositories.yaml
```diff
@@ -1507,6 +1507,30 @@ applications:
       - v1_name: glean-dictionary
         app_id: glean-dictionary
 
+  - app_name: mdn_fred
+    canonical_app_name: Mozilla Developer Network
+    app_description: |
+      MDN Web Docs is an open-source, collaborative project documenting Web
+      platform technologies, including CSS, HTML, JavaScript, and Web APIs.
+      MDN also provides an extensive set of learning resources for beginning
+      developers and students.
```
badboy, line 1516: This description shows up on the dictionary (see https://dictionary.telemetry.mozilla.org/)
Can we do add something to better differentiate the two apps? They otherwise have the exact same display name and description.

caugner: Fixed in 421aa72.

badboy (approved):
dry-run:

```
python3 -m probe_scraper.runner --dry-run --cache-dir tmp/cache --out-dir tmp/out --glean --glean-limit-date 2025-07-08 --output-bucket=gs://probe-scraper-prod-artifacts/ --glean-repo mdn-fred
```

looks clean. :shipit:

PR #958, opened 2025-10-31
[SVCSE-3019] Remove regrets reporter and tiktok reporter apps

scholtzan:
https://mozilla-hub.atlassian.net/browse/SVCSE-3019

We are stopping data ingestion for regrets reporter apps.


chutten:
The TikTok reporters aren't yet deprecated. Is skipping that state deliberate?

What does Glean Dictionary do when apps are removed? What does the probe info service do? Will they remain frozen in their current state, or disappear?

chutten (approved):

PR #959, opened 2025-11-03
Add entries for error-support

bendk:
This is my attempt at addressing https://bugzilla.mozilla.org/show_bug.cgi?id=1997931.  Am I added the entry correctly?

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -94,6 +94,22 @@ libraries:
     variants:
       - v1_name: sync
         dependency_name: org.mozilla.appservices:syncmanager
+
+  - library_name: error-support
+    description: >-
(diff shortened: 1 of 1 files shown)


chutten:
This is a necessary but not sufficient change to tell the pipeline about error-support. To get its tables and columns to be generated for its pings and metrics, you'll also need to specify which libraries and/or applications depend on error-support. You can see how this works for other a-s components by grepping for the other `org.mozilla.appservices:` dependencies and finding them in e.g. `fenix`, `firefox-ios`, ...

bendk:
Thanks.  I added this to the fenix dependencies since as of now it's only used by Fenix.

badboy (approved):

PR #961, opened 2025-11-03
Change experimenter from glean-core to glean-server

relud:
experimenter is switching to server-side glean in https://github.com/mozilla/experimenter/pull/13859


badboy (approved):

relud:
filed https://github.com/mozilla/mozilla-schema-generator/pull/299 to allow the deletes

PR #963, opened 2025-11-06
Add subscription-platform-backend-cirrus to repositories.yaml

relud:
fixes [EXP-6110](https://mozilla-hub.atlassian.net/browse/EXP-6110)

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -2122,7 +2122,7 @@ applications:
       expiration_policy:
         delete_after_days: 760
 
-  - app_name: subplat_cirrus
+  - app_name: subscription_platform_backend_cirrus
     canonical_app_name: "Subscription Platform (Cirrus)"
     app_description: |
       Mozilla Subscription Platform is Mozilla's subscription management
@@ -2137,8 +2137,8 @@ applications:
       - glean-server
       - nimbus-cirrus
     channels:
-      - v1_name: subplat-cirrus
-        app_id: subplat.cirrus
+      - v1_name: subscription-platform-backend-cirrus
+        app_id: subscription.platform.backend.cirrus
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 760


repositories.yaml
```diff
@@ -2121,3 +2121,24 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 760
+
+  - app_name: subplat_cirrus
+    canonical_app_name: "Subscript Platform (Cirrus)"
```
sean-rose: **typo:**
```suggestion
    canonical_app_name: "Subscription Platform (Cirrus)"
```

relud: oops, thank you

repositories.yaml
```diff
@@ -2121,3 +2121,24 @@ applications:
     moz_pipeline_metadata_defaults:
       expiration_policy:
         delete_after_days: 760
+
+  - app_name: subplat_cirrus
```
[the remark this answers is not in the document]
relud: tl;dr this isn't backend specific

More precisely, while we expect only the backend to use it, we expect any experiments in the frontend to receive feature configuration from the backend, not directly from cirrus. This matches with `accounts_cirrus` which is also only used in the backend.

integrating with cirrus on the frontend would be for applications that don't have a suitable backend, e.g. static sites that only have client-side code.

relud:
@akkomar cirrus is moving to server telemetry as soon as you release https://github.com/mozilla/glean_parser/pull/820, and also based on @badboy's feedback in slack it only sends custom pings anyways, because it uses the python sdk, doesn't explicitly interact with any glean-core pings, and doesn't use any events-ping events.

I think this should be the recommended approach for newly created cirrus apps, so that doc should be updated from saying glean-core to glean-server.

I had not seen that about naming of apps with glean-server, I'll change this to `subscription_platform_backend_cirrus` as per sean's suggestion, but I think existing cirrus applications will have to settle for their existing names.

mikewilli:
Re: @akkomar 's comment:
>We have some guidelines in https://mozilla-hub.atlassian.net/wiki/spaces/IP/pages/142213141/Enable+ingestion+for+new+Glean+apps that cover this case. They're contradictory (dependencies for _cirrus apps and naming of apps with glean-server) and we probably should update them.

I think we should update the docs to say that the `_cirrus` takes precedence over `_backend` as the suffix. I'll bring this up for discussion to confirm and then can update the docs.

PR #969, opened 2025-11-19
Update repositories.yaml with the crashping library for bug 1989439.

afranchuk:
I think I added it to all the correct applications, but check my work :)

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -75,11 +75,28 @@ libraries:
     url: https://github.com/mozilla-firefox/firefox
     metrics_files:
       - mobile/android/android-components/components/lib/crash/metrics.yaml
-    ping_files:
-      - mobile/android/android-components/components/lib/crash/pings.yaml
     variants:
       - v1_name: lib-crash
         dependency_name: org.mozilla.components:lib-crash
+        branch: main
+
+  - library_name: crashping
+    description: >-
+      The crashping crate, which is used on all platforms to send crash pings.
+    notification_emails:
+      - afranchuk@mozilla.com
+      - crash-reporting-wg@mozilla.org
+      - stability@mozilla.org
+    url: https://github.com/mozilla-firefox/firefox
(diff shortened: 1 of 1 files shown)


badboy (changes requested):
Wait, I thought this will send things under a different app ID? 
Why is this now a library?
There already _is_ a ping named `crash` sent by e.g. `firefox_desktop`. 
Having that re-defined in another library now that's also going under `firefox_desktop` will lead to complications.

afranchuk:
> The application ID isn't stored in the database. Re-using the same database directory with a different application ID would thus work and send the same data. However using the same database simultaneously from multiple applications is dangerous. When and how does this client run? Is a Firefox instance active while it runs? There's also other things we need to consider in this case (e.g. we need to delete/cleanup that crash data with client IDs on opt-out from the user).

The database will not be used from multiple applications at the same time: Firefox desktop doesn't use the `crashping` crate at all, it just invokes the crash reporter client in a mode where it sends a ping, and there is a dedicated, unique glean store. As it is now, in both this case and the main-process-crash case (where the crash reporter client is launched), the same application id is used. I was wondering whether, if I changed this to use different application ids between the two cases, it would use the same client id (and it sounds like it would). But maybe it's not worth changing the application ID here: there are other more informative fields if we needed to separate the source of the ping (and, in theory, we shouldn't need to).

> By derived you mean it gets that ".crashping" suffix? That means our pipeline will need to know about these new application IDs, or we won't ingest data from it. Let me talk with my team to figure out what the best way is on this

Yes, it gets the suffix and isn't a fixed value. But perhaps it would be clearer and easier to track if it were created with some fixed strings rather than using `MOZ_APP_NAME`.

afranchuk:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1989439#c16, once this is merged, we should manually trigger schema generation to confirm this doesn't break anything.

PR #1036, opened 2026-06-19
Bug 2042327 - New library: ios-places

badboy:


chutten (approved):
LGTM -- Do we need focus-ios as well?

badboy:
> LGTM -- Do we need focus-ios as well?

Oh actually no, focus doesn't use the places component.

PR #1037, opened 2026-06-25
Use new Thunderbird git repo

freaktechnik:
We've now been migrated to git as well (https://thunderbird.topicbox.com/groups/developers/Ta0c99fe7d36282a2/git-migration-update). This updates the scraper to refer to the new working repo.

### DIFF
--- a/repositories.yaml
+++ b/repositories.yaml
@@ -2017,7 +2017,7 @@ applications:
   - app_name: thunderbird_desktop
     canonical_app_name: Thunderbird
     app_description: Thunderbird email client
-    url: https://github.com/mozilla/releases-comm-central
+    url: https://github.com/thunderbird/thunderbird-desktop
     notification_emails:
       - alessandro@thunderbird.net
       - sancus@thunderbird.net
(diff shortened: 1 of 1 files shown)


badboy (approved):
👍
We do have airflow tasks running for this repo as well, these will need to be updated: https://github.com/mozilla/telemetry-airflow//blob/main/dags/probe_scraper.py#L208

freaktechnik:
> 👍 We do have airflow tasks running for this repo as well, these will need to be updated: https://github.com/mozilla/telemetry-airflow//blob/main/dags/probe_scraper.py#L208

Hm, would the name also change, or is that something that'll stay the same, even if the repo name in the link changes?

badboy:
> > 👍 We do have airflow tasks running for this repo as well, these will need to be updated: [mozilla/telemetry-airflow//blob/main/dags/probe_scraper.py#L208](https://github.com/mozilla/telemetry-airflow//blob/main/dags/probe_scraper.py#L208)
> 
> Hm, would the name also change, or is that something that'll stay the same, even if the repo name in the link changes?

That should probably stay the same. It's only used in the task name, but that allows to keep continuity in Airflow when those tasks are updated with the new url

PR #1044, opened 2026-07-21
rename enum and structure fields when serializing metrics

jeddai:
This renames enum and structure internal fields so that we can have that data in the probe info service.


travis79 (approved):
Was going to ask for a changelog update too, but no changelog so this lgtm!

PR #1051, opened 2026-07-22
Bug 2056699 - CI: Auto-merge PRs from admins that are only touching repositories.yaml

badboy:
They are ONLY merged after tests succeed!

---

We could add additional checks to make sure this only applies to PRs opened by the fog-updater.
It's also currently missing a secret (`GH_PAT`) that is a Personal Access Token from one of us (e.g. me) with admin privileges, so that the github action can force the merge (and overwriting our "requires a reviewer" protection).

Alternatively we could use either a bot account or a GitHub app with the right permissions, both of which are a bit more involved to do.

### DIFF
--- a/fog-updater/src/fog_updater/__init__.py
+++ b/fog-updater/src/fog_updater/__init__.py
@@ -274,9 +274,6 @@ def run(argv, repo, author, debug=False, dry_run=False):
         base=release_branch_name,
     )
 
-    # Enable auto-merge on the new pull request
-    pr.enable_automerge(merge_method="REBASE")
-
     print(f"{ts()} Pull request at {pr.html_url}")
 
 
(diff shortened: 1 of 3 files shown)


fog-updater/src/fog_updater/__init__.py
```diff
@@ -268,9 +268,6 @@ def run(argv, repo, author, debug=False, dry_run=False):
         base=release_branch_name,
     )
 
-    # Enable auto-merge on the new pull request
-    pr.enable_automerge(merge_method="REBASE")
```
badboy, line 278: Yes, I just recently added that, but it turns out it doesn't help us much because the workflow won't merge until a review anyway, so we're forcing the merge in the other workflow (see above). Thus we don't need GitHub's own auto-merge workflow

badboy:
This is now using a (not-yet-set-up) GitHub app to get a token to use.

PR #1062, opened 2026-08-06
glean-push: Use Python 3.12, the newest runtime supported on GCF 1st gen

akkomar:
Commit f0b28ff bumped the runtime to python314 to clear a gcloud deprecation warning about Python 3.10, but python313 and python314 are 2nd gen only, so every deploy since has failed with:

  ERROR: (gcloud.functions.deploy) argument `--runtime`: python314 is
  not a supported runtime on GCF 1st gen.

glean-push is a 1st gen function, where python312 is the newest GA runtime. That still resolves the Python 3.10 deprecation. Moving to python313/314 would require migrating the function to 2nd gen, which cannot be done in place.


badboy (approved):

akkomar:
Follow up: bump function to v2: https://mozilla-hub.atlassian.net/browse/DENG-11458

